PR #171, opened 2019-08-15
fix(component): select will now change input text if value changes

jorgemoya:
Component will now match the input text to the value selected, without having to click an option.

### DIFF
--- a/packages/big-design/src/components/Select/spec.tsx
+++ b/packages/big-design/src/components/Select/spec.tsx
@@ -200,6 +200,21 @@ test('select item should be highlighted when opened', () => {
   expect(input.getAttribute('aria-activedescendant')).toEqual(options[1].id);
 });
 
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Select/spec.tsx
```diff
@@ -200,6 +200,21 @@ test('select item should be highlighted when opened', () => {
...
+      <Select.Option value="us">United States</Select.Option>
+      <Select.Option value="mx">Mexico</Select.Option>
+      <Select.Option value="ca">Canada</Select.Option>
+      <Select.Option value="en">England</Select.Option>
+    </Select>,
+  );
+
+  const input = getByLabelText('Countries');
+
+  expect(input.getAttribute('value')).toEqual('Mexico');
```
deini: Can you 🙏 add another expect after re-rendering with a different value?

PR #177, opened 2019-08-16
fix(component): added scroll prevention on open modals

chanceaclark:
## What
Prevent scrolling on open modals.

Could be BREAKING CHANGE, so let me know if I need to amend the commit.

### DIFF
--- a/packages/big-design/src/components/Modal/Modal.tsx
+++ b/packages/big-design/src/components/Modal/Modal.tsx
@@ -17,6 +17,7 @@ import {
 } from './styled';
 
 export interface ModalProps {
+  appendBodyOverflow: boolean;
   backdrop: boolean;
   isOpen: boolean;
   closeOnClickOutside: boolean;
@@ -49,6 +50,7 @@ const ModalHeader: React.FC<ModalHeaderProps> = ({ children, withBorder }) => {
 
 export class Modal extends React.PureComponent<ModalProps> {
   static defaultProps: Partial<ModalProps> = {
+    appendBodyOverflow: true,
     backdrop: true,
     isOpen: false,
     closeOnClickOutside: false,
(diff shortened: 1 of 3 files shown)


packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -17,6 +17,7 @@ import {
 } from './styled';
 
 export interface ModalProps {
+  appendBodyOverflow: boolean;
```
deini: Should we call this something like `lockPageScroll`?

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -49,6 +50,7 @@ const ModalHeader: React.FC<ModalHeaderProps> = ({ children, withBorder }) => {
 
 export class Modal extends React.PureComponent<ModalProps> {
   static defaultProps: Partial<ModalProps> = {
+    appendBodyOverflow: true,
```
deini: Can you think of a use case where we wouldn't want to lock the scroll? We chatted about making it an opt-out but I'm not sure anymore. What do you think about always locking the page scroll and we can add an opt-out prop in the future if we get the use case? 🤔

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -74,13 +76,21 @@ export class Modal extends React.PureComponent<ModalProps> {
     if (this.modalContainer) {
       document.body.removeChild(this.modalContainer);
     }
+
+    if (this.props.appendBodyOverflow) {
+      document.body.style.overflowY = 'initial';
```
deini: A bit of a corner case but wouldn't this override the style if it had a different value? I think we should we be storing and resetting the previous value.

PR #178, opened 2019-08-19
fix(icons): now treeshakeable

deini:
🌳 🥛


deini:
It tells Terser or UglifyJS that the following function call is `pure` (side-effects free) and can be deleted if not used.

PR #179, opened 2019-08-19
fix: export margin and padding mixins/functions

chanceaclark:
Exported `MarginProps`, `PaddingProps`, `withMargins()`, and `withPadding()`.

### DIFF
--- a/packages/big-design/src/index.ts
+++ b/packages/big-design/src/index.ts
@@ -1,1 +1,2 @@
 export * from './components';
+export * from './mixins';


packages/big-design/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './components';
+export * from './mixins';
```
deini: Looks good, can you change the commit to either a `feat` or `fix` (not sure how you want to look at this), since `chore` commits don't generate changelog entries.

PR #182, opened 2019-08-22
feat(component): add ability to pass arrays into input error

chanceaclark:
## What

Add ability to pass in an array of strings into `Input`'s.

Also noticed that `React.ReactChild` includes strings, so removed erroneous `string` type declarations.

### DIFF
--- a/packages/big-design/src/components/Input/Input.tsx
+++ b/packages/big-design/src/components/Input/Input.tsx
@@ -10,7 +10,7 @@ import { StyledIconWrapper, StyledInput, StyledInputWrapper } from './styled';
 
 interface Props {
   description?: React.ReactChild;
-  error?: React.ReactChild;
+  error?: React.ReactChild | React.ReactChild[];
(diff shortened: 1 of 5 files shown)


packages/big-design/src/components/Input/Input.tsx
```diff
@@ -10,7 +10,7 @@ import { StyledIconWrapper, StyledInput, StyledInputWrapper } from './styled';
 
 interface Props {
   description?: React.ReactChild;
-  error?: React.ReactChild;
+  error?: React.ReactChild | React.ReactChild[];
```
deini: How does this look when we pass an array of errors to an input not wrapped in a group?

chanceaclark: We removed the functionality in #129 in favor of always wrapping `Input`'s in `Form.Group`'s.

deini: Ahh true

PR #188, opened 2019-09-04
feat(component): create basic pagination component

jordan-massingill:
# WHAT

Created Pagination component to be used with lists and tables 

# TESTING

jest testing

# SCREENSHOTS
screencapture-localhost-3000-pagination-2019-09-12-13_58_29


packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export const Pagination: React.FunctionComponent<PaginationProps> = props => {
+  const { currentPerPage, currentPage, numOfItems, perPageOptions, onPageUpdate, onRangeUpdate } = props;
+
+  const [maxPages, setMaxPages] = useState(Math.ceil(numOfItems / currentPerPage));
+  const [itemRange, setItemRange] = useState([0, 0]);
+
+  useEffect(() => {
+    const first = currentPerPage * (currentPage - 1) + 1;
```
chanceaclark: What is the context of `first` and `second`? Could you be a bit more descriptive?

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+    let second = currentPerPage * currentPage;
+    if (second > numOfItems) {
+      second = numOfItems;
+    }
+    const range = [first, second];
+    setItemRange(range);
```
chanceaclark: You could get a bit more cleaner by removing the `const` declaration
```suggestion
    setItemRange([first, second]);
```

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+
+  const handleRangeChange = (range: number) => {
+    setMaxPages(Math.ceil(numOfItems / range));
+    onRangeUpdate(range);
+  };
+
+  const showRanges = () => {
+    if (itemRange[0] !== itemRange[1]) {
+      return itemRange[0] + ' - ' + itemRange[1] + ' of ' + numOfItems;
```
chanceaclark:
```suggestion
      return `${itemRange[0]} - ${itemRange[1]} of ${numOfItems}`;
```

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+    } else {
+      return itemRange[0] + ' of ' + numOfItems;
```
chanceaclark:
```suggestion
      return `${itemRange[0]} of ${numOfItems}`;
```

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+  };
+
+  return (
+    <Flex flexDirection="row" justifyContent="space-between" role="navigation" aria-label="pagination">
+      <Dropdown
+        onItemClick={handleRangeChange}
+        trigger={
+          <StyledButton>
```
chanceaclark: You may want to use the `StyleableButton` to extend from so that way you can interact with it as it was one of our own instead of manually creating one. See the Tabs component for an example usage of it.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+            {showRanges()}
+            <ArrowDropDownIcon color="secondary70" size="xxLarge" />
+          </StyledButton>
+        }
+      >
+        {perPageOptions.map(range => {
+          return <Dropdown.Item value={range}>{range} per page</Dropdown.Item>;
+        })}
+      </Dropdown>
+      {currentPage > 1 ? (
```
chanceaclark: I would wrap these icons in a button, specifically one you extend from above comment, so you don't have to manually deal with the accessibility stuff, plus you can determine `disabled` through the button prop rather than a conditional statement.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const showRanges = () => {
+    if (itemRange[0] !== itemRange[1]) {
+      return itemRange[0] + ' - ' + itemRange[1] + ' of ' + numOfItems;
+    } else {
+      return itemRange[0] + ' of ' + numOfItems;
+    }
+  };
+
+  return (
+    <Flex flexDirection="row" justifyContent="space-between" role="navigation" aria-label="pagination">
```
chanceaclark: Gotta get that vertical alignment. Also might want to wrap the left and right arrows in another `Flex` container so they are together.
```suggestion
    <Flex flexDirection="row" alignItems="center" justifyContent="space-between" role="navigation" aria-label="pagination">
```

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,102 @@
...
+export interface PaginationProps extends MarginProps {
+  currentPerPage: number;
+  currentPage: number;
+  numOfItems: number;
+  perPageOptions: number[];
+  onPageUpdate(page: number): void;
+  onRangeUpdate(range: number): void;
+}
+
+export const Pagination: React.FunctionComponent<PaginationProps> = props => {
```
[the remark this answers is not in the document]
deini: I prefer if we keep the extension of `MarginProps` since its simpler to just import `PaginationProps` and assume everything is going to be there.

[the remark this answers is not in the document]
chanceaclark: ^ he's right

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+import { StyledButton } from './styled';
+
+export interface PaginationProps extends MarginProps {
+  currentPerPage: number;
+  currentPage: number;
+  numOfItems: number;
+  perPageOptions: number[];
+  onPageUpdate(page: number): void;
+  onRangeUpdate(range: number): void;
```
jorgemoya: Talking to @chanceaclark it seems there's an opportunity to make some of this `prop` names descriptive, since they were a little confusing to him first.

Some suggestions would be to change `onPageUpdate` & `onRangeUpdate` to `onPageChange` & `onRangeChange`.

I also think we need to match `currentPerPage` to `onRangeUpdate` and change it to `currentRange`. Or vice versa.

Another suggestion would be to change `numOfItems` to `maxItems`.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+    let nextPage = currentPage + 1;
+    if (nextPage > maxPages) {
+      nextPage = 1;
+    }
+    onPageUpdate(nextPage);
+  };
+
+  const handlePageDecrease = () => {
+    let nextPage = currentPage - 1;
+    if (nextPage < 1) {
```
jorgemoya: Same as above.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (second > numOfItems) {
+      second = numOfItems;
+    }
+    const range = [first, second];
+    setItemRange(range);
+  }, [currentPage, currentPerPage, numOfItems]);
+
+  const handlePageIncrease = () => {
+    let nextPage = currentPage + 1;
+    if (nextPage > maxPages) {
```
jorgemoya: I believe we no longer need this loop logic since you're disabling the button.

packages/big-design/src/components/Pagination/Pagination.tsx
jorgemoya: Err, same as below.

packages/docs/PropTables/PaginationPropTable.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import { PropTable } from '../components';
+
+export const PaginationPropTable: React.FC = () => {
+  return (
+    <PropTable>
+      <PropTable.Prop name="currentRange" types="number" defaults="" required>
+        Indicates how many items are displayed per page
```
chanceaclark: Would you mind adding periods to the ends of these?

packages/big-design/src/components/Pagination/styled.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components';
+
+import { StyleableButton } from '../Button/private';
+
+export const StyledButton = styled(StyleableButton)`
+  color: ${({ theme }) => theme.colors.secondary70};
+`;
+
+// in case we need to add separate styles to the icon buttons or to the dropdown trigger
+export const StyledIconButton = styled(StyleableButton)`
```
chanceaclark: We probably don't want another button unless we absolutely need it. You could get rid of this and just focus on styling one button above.

packages/big-design/src/components/Pagination/styled.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components';
+
+import { StyleableButton } from '../Button/private';
+
+export const StyledButton = styled(StyleableButton)`
+  color: ${({ theme }) => theme.colors.secondary70};
```
chanceaclark: You'll probably also want to add state styles seen on the [`ButtonSubtle` component](https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Button/styled.tsx#L186)

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <StyledButton variant="subtle" iconRight={<ArrowDropDownIcon size="xxLarge" color="secondary70" />}>
+            {showRanges()}
+          </StyledButton>
+        }
+      >
+        {rangeOptions.map(range => {
+          return <Dropdown.Item value={range}>{range} per page</Dropdown.Item>;
+        })}
+      </Dropdown>
+      <Box>
```
chanceaclark:
```suggestion
      <Flex.Item>
```

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+        <StyledIconButton
+          variant="subtle"
+          disabled={currentPage === maxPages}
+          aria-label="next page"
+          onClick={handlePageIncrease}
+        >
+          <ChevronRightIcon />
+        </StyledIconButton>
+      </Box>
```
chanceaclark:
```suggestion
      </Flex.Item>
```

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  useEffect(() => {
+    const firstItemInRange = currentRange * (currentPage - 1) + 1;
+    let lastItemInRange = currentRange * currentPage;
+    if (lastItemInRange > totalItems) {
+      lastItemInRange = totalItems;
+    }
+
+    setItemRange([firstItemInRange, lastItemInRange]);
+  }, [currentPage, currentRange, totalItems]);
```
chanceaclark: Correct me if I'm wrong, but if `totalItems` changes, `maxPages` will not update. So with that in mind, you'll maybe want to have another `useEffect` listening to only `totalItems` and update it whenever that changes.

jordan-massingill: Since the state of `currentPage` is dependent on `totalItems` and `currentRange` props, it should/does change dynamically without the need for a `useEffect` listener (at least, that has been my experience with the component). We may need to look at messing around with this to be certain though.

packages/big-design/src/components/Pagination/Pagination.tsx
jordan-massingill: I'm now editing this logic to instead never allow the page number to go outside the range of 1-`maxPages` to cover any edge cases, rather than loop.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <Dropdown.Item key={key} value={range}>
+            {range} per page
+          </Dropdown.Item>
+        ))}
+      </Dropdown>
+      <Flex.Item>
+        <StyledButton
+          variant="subtle"
+          disabled={currentPage <= 1}
+          aria-label="previous-page"
```
chanceaclark: Remove this `aria-label` and add it to the `ChevronLeftIcon` title prop. Screen readers will pick up the `<title>...</title>` from a svg element. Also rename to `Previous page`.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,119 @@
...
+          onClick={handlePageDecrease}
+        >
+          <ChevronLeftIcon />
+        </StyledButton>
+
+        <StyledButton
+          variant="subtle"
+          disabled={currentPage >= maxPages}
+          aria-label="next-page"
```
chanceaclark: Same as above (`aria-label`)

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+import { StyledButton } from './styled';
+
+export interface PaginationProps extends MarginProps {
+  currentPage: number;
+  itemsPerPage: number;
+  itemsPerPageOptions: number[];
+  totalItems: number;
+  onPageChange(page: number): void;
+  onRangeChange(range: number): void;
```
chanceaclark: 🤔Should this now be `onItemsPerPageChange`?

packages/big-design/src/components/Pagination/Pagination.tsx
jordan-massingill: TY - will do now

packages/big-design/src/components/Pagination/Pagination.tsx
jordan-massingill: Nice catch, TY

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export const Pagination: React.FC<PaginationProps> = props => {
+  const { itemsPerPage, currentPage, totalItems, itemsPerPageOptions, onPageChange, onItemsPerPageChange } = props;
+
+  const [maxPages, setMaxPages] = useState(Math.ceil(totalItems / itemsPerPage));
+  const [itemRange, setItemRange] = useState({ start: 0, end: 0 });
+
+  const handlePageOutOfBounds = () => {
+    if (currentPage < 1 || isNaN(currentPage) || currentPage === undefined) {
```
jorgemoya: I'm wondering if we need to verify for `isNaN` or `undefined` when typescript has `currentPage` as `number`. I guess it makes sense to verify for those not using `ts`, not sure what others think?

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,119 @@
...
+      onPageChange(1);
+    } else if (currentPage > maxPages) {
+      onPageChange(maxPages);
+    }
+  };
+
+  const handlePerPageOutOfBounds = () => {
+    if (itemsPerPage < 1 || isNaN(itemsPerPage) || itemsPerPage === undefined) {
```
jorgemoya: Same as above.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  const calculateRange = () => {
+    let firstItemInRange = itemsPerPage * (currentPage - 1) + 1;
+    let lastItemInRange = itemsPerPage * currentPage;
+
+    firstItemInRange = Math.min(firstItemInRange, totalItems);
+    lastItemInRange = Math.min(lastItemInRange, totalItems);
+
+    if (lastItemInRange === 0 || isNaN(lastItemInRange) || isNaN(firstItemInRange)) {
```
jorgemoya: Same as above.

PR #189, opened 2019-09-05
feat(component): Add chip component

animesh1987:
## What
- Add chip component.

## Testing
- Jest tests

@jorgemoya


packages/big-design/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const renderDeleteButton = () =>
+    onDelete && (
+      <StyledCloseButton
+        variant="subtle"
+        onClick={handleOnDelete}
+        iconOnly={<CloseIcon size="medium" title="Delete" />}
+      ></StyledCloseButton>
+    );
+
+  return (
```
chanceaclark: Do we even want to render the chip if there is no `label`?
```suggestion
  return label ? (
```

deini:
Hey @animesh1987  👋 pushed a commit with a couple of changes mainly:

- Changed the API to `<Chip>Text</Chip>`
- There is only one prop now, `onDelete` which we also use to display/hide the close button.
- The  close svg is now wrapped in a `Button` component for accessibility

packages/big-design/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,47 @@
...
+      margin="xxSmall"
+      borderRadius="normal"
+    >
+      <Text margin="none" marginRight="xxSmall">
+        {label}
+      </Text>
+
+      {renderDeleteButton()}
+    </StyledChip>
+  );
```
chanceaclark: Addition from above.
```suggestion
  ) : null;
```

packages/big-design/src/components/Chip/Chip.tsx
chanceaclark: If we add the above, you should add another test with no children.

packages/big-design/src/components/Chip/styled.tsx
```diff
@@ -0,0 +1,26 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+import { Box } from '../Box';
+import { StyleableButton } from '../Button/Button';
+
+export const StyledChip = styled(Box)`
+  align-items: center;
+  display: inline-flex;
+  user-select: none;
```
chanceaclark: 🍹Is there a specific reason why we wouldn't want people to select the text?

deini: Not really, it just looks terrible and doesn't seem that useful 🤔

packages/big-design/src/components/Chip/Chip.tsx
deini: I thought about that but I _think_ it's a bit easier to spot if we render a chip which looks wrong than to not render at all. Thoughts?

jorgemoya: Just from my own experience experimenting with `chips`, I think it's better if we display an empty chip than no chip at all. I was able to debug I wasn't sending the right text immediately.

animesh1987:
Thanks for pushing it through @deini 👍

PR #193, opened 2019-09-12
fix: updates the input label if a matching children appears

rtalvarez:
## What?
Updates the input text of a Select component if a matching children suddenly appears within its children

## Testing done
- Verified bug cant be reproduced anymore
- Unit test

### DIFF
--- a/packages/big-design/src/components/Select/Select.tsx
+++ b/packages/big-design/src/components/Select/Select.tsx
@@ -62,8 +62,8 @@ export class Select extends React.PureComponent<SelectProps, SelectState> {
     this.updatedSelectedItem();
   }
 
-  componentDidUpdate(prevProps: Props) {
-    const { value } = this.props;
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Select/Select.tsx
```diff
@@ -74,6 +74,10 @@ export class Select extends React.PureComponent<SelectProps, SelectState> {
     if (prevProps.value !== value) {
       this.updatedSelectedItem();
     }
+
+    if (prevProps.children !== children) {
```
deini: Should we be returning early when calling `updatedSelectedItem`? 

Seems like we could be calling it multiple times.

jorgemoya: ☝️That makes sense. Should've added it before.

rtalvarez: I'll add that in

PR #201, opened 2019-09-23
feat(component): add Table component

chanceaclark:
## What

Adding basic Table component.

A few notes:
- Documentation page is disabled for now; still a WIP.
- Needed the checkbox to render without a label. For now, just adding boolean as a property type so we could pass in false. Need to come up with a way to have a visually hidden label.
- Wrapped `Pagination` Dropdown in a Flex.Item for correct positioning.

## Screenshot
Screen Shot 2019-09-23 at 10 42 01 AM


packages/big-design/src/components/Table/Body/Body.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { TableSectionContext } from '../context';
+
+import { StyledTableBody } from './styled';
+
+export interface TableBodyProps extends React.TableHTMLAttributes<HTMLTableSectionElement> {}
+
+export const TableBody: React.FC<TableBodyProps> = ({ className, children, style, ...props }) => (
+  <TableSectionContext.Provider value={'tbody'}>
```
rtalvarez: 🍹  `value="tbody"`

packages/big-design/src/components/Table/Footer/Footer.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { TableSectionContext } from '../context';
+
+import { StyledTableFooter } from './styled';
+
+export type TableFooterProps = React.TableHTMLAttributes<HTMLTableSectionElement>;
+
+export const TableFooter: React.FC<TableFooterProps> = props => (
+  <TableSectionContext.Provider value={'tfoot'}>
```
rtalvarez: 🍹 same

packages/big-design/src/components/Table/Head/Head.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { TableSectionContext } from '../context';
+
+import { StyledTableHead } from './styled';
+
+export type TableHeadProps = React.TableHTMLAttributes<HTMLTableSectionElement>;
+
+export const TableHead: React.FC<TableHeadProps> = props => (
+  <TableSectionContext.Provider value={'thead'}>
```
rtalvarez: 🍹 same

packages/big-design/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+
+import { uniqueId } from '../../utils';
+
+import { TableContext } from './context';
+import { StyledTable, StyledTableFigure } from './styled';
+import { TableActions, TableActionsProps } from './Actions/Actions';
```
deini: Can we create `index.ts` files for these imports?

packages/big-design/src/components/Table/Row/Row.tsx
```diff
@@ -0,0 +1,44 @@
...
+export interface TableRowProps extends React.TableHTMLAttributes<HTMLTableRowElement> {
+  selected?: boolean;
+  verticalAlign?: 'top' | 'bottom' | 'middle';
+  onRowSelect?(selected: boolean): void;
+}
+
+export const TableRow: React.FC<TableRowProps> = memo(({ children, selected, onRowSelect, ...props }) => {
+  const tableContext = useContext(TableContext);
+  const tableSectionContext = useContext(TableSectionContext);
+  const [isSelected, setIsSelected] = useState(selected);
```
jorgemoya: I was under the impression `state` shouldn't be set from a `prop` received. Is this a pattern that has changed with hooks?

packages/big-design/src/components/Table/Cell/styled.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  ${props =>
+    props.isCheckbox &&
+    css`
+      width: ${({ theme }) => theme.helpers.addValues(theme.spacing.xLarge, theme.spacing.small)};
+      white-space: nowrap;
+    `}
+
+  &:last-of-type {
+    width: 50%;
```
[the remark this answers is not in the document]
chanceaclark: Good catch, that was stylings from the prop tables that need to be removed from this.

packages/big-design/src/components/Table/Table.tsx
chanceaclark: Is there a reason why we haven't been doing that for `Flex` and `Grid` static members then?

PR #206, opened 2019-09-30
feat(docs): add figma UI kit and sample app to resources section

amckemie:
**What**
Added links in Helpful Resources section for Figma UI Kit and sample app. 

**Why**
https://jira.bigcommerce.com/browse/CHP-6053

To provide developers and designers with more resources to help them in building apps with BigDesign

### DIFF
--- a/packages/docs/pages/GettingStarted/GettingStartedPage.tsx
+++ b/packages/docs/pages/GettingStarted/GettingStartedPage.tsx
@@ -54,10 +54,7 @@ export default () => {
             </Link>
           </List.Item>
           <List.Item>
-            <Link
-              href="https://www.figma.com/file/jTVuUkiZ1j3rux8WHG4IKK/BigDesign-UI-Kit?node-id=0%3A1/duplicate"
-              target="_blank"
-            >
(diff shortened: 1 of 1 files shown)


packages/docs/pages/GettingStarted/GettingStartedPage.tsx
```diff
@@ -48,6 +48,19 @@ export default () => {
               Beta Community Group
             </Link>
           </List.Item>
+          <List.Item>
+            <Link
+              href="https://www.figma.com/file/jTVuUkiZ1j3rux8WHG4IKK/BigDesign-UI-Kit?node-id=0%3A1/duplicate"
```
jorgemoya: May be a better idea to change url to `https://www.figma.com/file/jTVuUkiZ1j3rux8WHG4IKK/BigDesign-UI-Kit`. In fact there's no need to include the title, but I'm not sure if it would still redirect to the file if the title were to change (but probably would).

amckemie: Updated to remove that, as you were correct.

PR #207, opened 2019-09-30

chanceaclark:
## What

Add `ButtonGroup` component to Library. Only accepts `Button` and `Dropdown` components as children.

## Screenshot
screencapture-localhost-3000-button-group-2019-09-30-14_31_22

### DIFF
--- a/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
@@ -1,0 +1,29 @@
+import { ThemeInterface } from '@bigcommerce/big-design-theme';
+import React from 'react';
+
+import { Button } from '../Button/Button';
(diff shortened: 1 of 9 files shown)


packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { StyledButtonGroup } from './styled';
+
+export interface ButtonGroupProps extends FlexProps {
+  theme?: ThemeInterface;
+}
+
+export const ButtonGroup: React.FC<ButtonGroupProps> = ({ children, className, style, ...props }) => {
+  const renderChildren = () => {
+    return React.Children.map(children, child => {
+      if (React.isValidElement(child)) {
```
rtalvarez: could these two if's be combined into one? unless im missing something here

PR #208, opened 2019-10-03
feat(component): add optional text to non required inputs

jorgemoya:
## What
Appends `(optional)` to non required inputs and selects.

## Screenshot
Screen Shot 2019-10-03 at 3 07 24 PM

### DIFF
--- a/packages/big-design/src/components/Form/Label/Label.tsx
+++ b/packages/big-design/src/components/Form/Label/Label.tsx
@@ -1,7 +1,14 @@
 import React from 'react';
 
-import { StyledLabel } from './styled';
+import { StyledLabel, StyledOptionalSpan } from './styled';
 
-export type LabelProps = React.LabelHTMLAttributes<HTMLLabelElement>;
+export interface LabelProps extends React.LabelHTMLAttributes<HTMLLabelElement> {
+  renderOptional?: boolean;
+}
 
-export const Label: React.FC<LabelProps> = ({ className, style, ...props }) => <StyledLabel {...props} />;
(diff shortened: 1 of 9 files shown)


packages/big-design/src/components/Form/Label/Label.tsx
```diff
@@ -1,7 +1,14 @@
...
-export type LabelProps = React.LabelHTMLAttributes<HTMLLabelElement>;
+export interface LabelProps extends React.LabelHTMLAttributes<HTMLLabelElement> {
+  renderOptional?: boolean;
+}
 
-export const Label: React.FC<LabelProps> = ({ className, style, ...props }) => <StyledLabel {...props} />;
+export const Label: React.FC<LabelProps> = ({ children, className, renderOptional, style, ...props }) => (
+  <StyledLabel {...props}>
+    {children}
+    {renderOptional && <StyledSpan> (optional)</StyledSpan>}
```
chanceaclark: 🍹Why did you end up going with a styled `span` instead of css `content: "(optional)";`?

jorgemoya: Didn't even think about it, tbh. I seem to forget about this way of rendering extra text. Seems like a good idea.

jorgemoya: Any idea how to test pseudoelements in jest? Can't seem to get it to work.

chanceaclark: Beat me to suggesting `toHaveStyleRule`!

PR #209, opened 2019-10-07
feat(component): restrict actions and header from modals

jorgemoya:
# BREAKING CHANGES

## What
Headers and Actions are now a prop of `modal` component. Visually no change.

Added `header` & `actions` props.

`actions` accepts an array of `Objects` with `Button` props and a `text` prop.

Example:
```ts
[
  { text: 'Cancel', onClick: () => setIsOpen(false) },
  { text: 'Apply', onClick: () => setIsOpen(false) },
]
```

Updated tests and docs.


packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -16,11 +16,15 @@ import {
...
 export interface ModalProps {
+  actions?: React.ReactNode;
   backdrop: boolean;
-  isOpen: boolean;
   closeOnClickOutside: boolean;
   closeOnEscKey: boolean;
+  header?: string;
+  isOpen: boolean;
   variant: 'modal' | 'dialog';
+  withActionsBorder?: boolean;
```
deini: Do we have multiple designs? I think this should be consistent in all modals and not configurable.

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -136,18 +116,28 @@ export class Modal extends React.PureComponent<ModalProps, ModalState> {
...
-    return React.Children.map(children, child => {
-      if (React.isValidElement(child) && child.type === Modal.Header) {
-        return React.cloneElement(child as React.ReactElement<HeadingProps>, {
-          id: this.headerUniqueId,
-        });
-      }
+    return (
+      header && (
+        <StyledModalHeader id={this.headerUniqueId} withBorder={withHeaderBorder}>
+          <H2 margin="none">{header}</H2>
```
deini: Should we do a runtime check to make sure `header` is a string? 🤔

deini:
Can you also add the `BREAKING CHANGE` section to your commit body with examples of how to migrate? 🙏

packages/big-design/src/components/Modal/Modal.tsx
chanceaclark: @deini It seems like you added it in #133 to support @animesh1987 use case. Is this something we need to touch back with design?

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -136,18 +116,28 @@ export class Modal extends React.PureComponent<ModalProps, ModalState> {
...
-      return child;
-    });
+    return (
+      actions && (
+        <StyledModalActions justifyContent="flex-end" withBorder={withActionsBorder}>
+          {actions}
+        </StyledModalActions>
+      )
+    );
   }
```
chanceaclark: To go along with @deini runtime check for strings, should we be consistent and do the same for `Buttons` within this `renderActions` method?

packages/big-design/src/components/Modal/spec.tsx
```diff
@@ -211,3 +213,40 @@ test('body has scroll locked on modal open', () => {
...
+test('renders header', () => {
+  const { getByText } = render(<Modal isOpen={true} header="Header Title" />);
+
+  expect(getByText('Header Title')).toBeInTheDocument();
+});
+
+test('renders header border', () => {
+  render(<Modal isOpen={true} header="Header Title" withHeaderBorder={true} />);
+
+  expect(document.body).toMatchSnapshot();
```
chanceaclark: 🍹You may also want to check if it has `toHaveStyleRule`.

packages/big-design/src/components/Modal/spec.tsx
```diff
@@ -211,3 +213,40 @@ test('body has scroll locked on modal open', () => {
...
+  const actions = (
+    <>
+      <Button variant="subtle">Cancel</Button>
+      <Button>Apply</Button>
+    </>
+  );
+
+  render(<Modal isOpen={true} actions={actions} withActionsBorder={true} />);
+
+  expect(document.body).toMatchSnapshot();
```
chanceaclark: 🍹Same as above

packages/docs/PropTables/ModalPropTable.tsx
```diff
@@ -4,11 +4,8 @@ import { Code, PropTable } from '../components';
 
 export const ModalPropTable: React.FC = () => (
   <PropTable>
-    <PropTable.Prop name="isOpen" types="boolean" required>
-      Determines if the modal/dialog is open.
-    </PropTable.Prop>
-    <PropTable.Prop name="onClose" types="() => void" required>
-      Function that will be called on close events.
+    <PropTable.Prop name="actions" types="React.ReactNode">
+      Sets the actions of the modal.
```
chanceaclark: Maybe we should reword this to what _should_ be passed into the actions (i.e. `Button`'s).

packages/docs/PropTables/ModalPropTable.tsx
```diff
@@ -19,24 +16,23 @@ export const ModalPropTable: React.FC = () => (
     <PropTable.Prop name="closeOnEscKey" types="boolean" defaults="true">
       Controls whether <Code>onClose</Code> is called when pressing the ESC key.
     </PropTable.Prop>
+    <PropTable.Prop name="header" types="string">
+      Header title of the modal.
```
chanceaclark: 🤷‍♀️Maybe different verbiage, but along these lines?
```suggestion
      Sets visible text that describes the content of the modal.
```

jorgemoya:
> Can you also add the `BREAKING CHANGE` section to your commit body with examples of how to migrate? 🙏

Is this done when you squash and merge the commit?

chanceaclark:
> > Can you also add the `BREAKING CHANGE` section to your commit body with examples of how to migrate? 🙏
> 
> Is this done when you squash and merge the commit?

Yeah, you can write it in full markdown.

packages/big-design/src/components/Modal/Modal.tsx
jorgemoya: Makes sense however how can you test for `Buttons` when you receive a set of actions?

For example in:
```
  const actions = (
    <>
      <Button variant="subtle">Cancel</Button>
      <Button>Apply</Button>
    </>
  );
```
In this case `React.Children.map` wouldn't iterate over the buttons.

packages/big-design/src/components/Modal/Modal.tsx
jorgemoya: Removed borders.

packages/docs/PropTables/ModalPropTable.tsx
```diff
@@ -1,14 +1,11 @@
...
 export const ModalPropTable: React.FC = () => (
   <PropTable>
-    <PropTable.Prop name="isOpen" types="boolean" required>
-      Determines if the modal/dialog is open.
-    </PropTable.Prop>
-    <PropTable.Prop name="onClose" types="() => void" required>
-      Function that will be called on close events.
+    <PropTable.Prop name="actions" types="object[]">
+      Accepts an array of objects with <NextLink href="/button">Button</NextLink> props and a <Code>text</Code> label.
     </PropTable.Prop>
```
jorgemoya: What do you guys think about this prop description?

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -29,42 +31,20 @@ interface ModalState {
...
 }
 
-export interface ModalActionsProps {
-  withBorder?: boolean;
-}
-
-export interface ModalHeaderProps {
-  withBorder?: boolean;
+interface Action extends Omit<ButtonProps, 'children'> {
+  text?: string;
```
deini: Should this be required?

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -29,42 +31,20 @@ interface ModalState {
   modalContainer: HTMLDivElement | null;
 }
 
-export interface ModalActionsProps {
-  withBorder?: boolean;
-}
-
-export interface ModalHeaderProps {
-  withBorder?: boolean;
+interface Action extends Omit<ButtonProps, 'children'> {
```
deini: What do you think about renaming this `ModalAction`? We also need to export it.

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -136,18 +118,34 @@ export class Modal extends React.PureComponent<ModalProps, ModalState> {
...
-        });
-      }
+  private renderActions() {
+    const { actions } = this.props;
 
-      return child;
-    });
+    return (
+      actions &&
+      Array.isArray(actions) && (
```
deini: I don't think we need the initial `actions &&`

packages/big-design/src/components/Modal/Modal.tsx
jorgemoya: I was thinking we could have only icon buttons

packages/big-design/src/components/Modal/Modal.tsx
jorgemoya: Why export it?

packages/big-design/src/components/Modal/Modal.tsx
deini: Good point, can't think of it being a good use case but let's leave it for now. Good catch.

packages/big-design/src/components/Modal/Modal.tsx
deini: For example, if you don't want to inline the `actions` prop you could do.

```tsx
const actions: ModalAction[] = [
  { text: 'Cancel', onClick: () => {} }
];

// And use it like
<Modal actions={actions} />
```

packages/docs/pages/Modal/ModalPage.tsx
```diff
@@ -26,29 +26,28 @@ export default () => (
...
-                </Text>
-              </Modal.Body>
-
-              <Modal.Actions>
-                <Button variant="subtle" onClick={() => setIsOpen(false)}>
-                  Cancel
-                </Button>
-                <Button onClick={() => setIsOpen(false)}>Apply</Button>
-              </Modal.Actions>
+            <Modal
```
icatalina: This looks great! 👍

PR #210, opened 2019-10-10
fix: setCustomValidity when Input updates/mounts

rtalvarez:
## What?
The current approach for `setCustomValidity` on Input components is broken, because it relies on user interaction via the `onChange` handler. Additionally, if no additional onChange handlers are run, the `setValidity` remains calculated based on the props from the previous render

## How?
I'm fixing this by making sure the custom validity is set when the component mounts or updates. This makes sure it is always kept up to date with the most recent props

## Testing done
- Verified component's validity on initial mount
- Verified component's validity on re-renders (on change events)
- Verified swapping the error prop syncs with the validity API
- Verified forms cannot be submitted when error prop exists

All of the above cases were verified by using no ref, `React.createRef` and ref callback functions.

I will continue to run tests, but everything I see works pretty well. LMK if i've missed anything

### DIFF
--- a/packages/big-design/src/components/Input/Input.tsx
+++ b/packages/big-design/src/components/Input/Input.tsx
@@ -40,10 +40,29 @@ class StyleableInput extends React.PureComponent<InputProps & PrivateProps, Inpu
   };
 
   private readonly uniqueId = uniqueId('input_');
+  private readonly defaultRef = React.createRef<HTMLInputElement>();
+  private callbackRef?: HTMLInputElement;
+
+  componentDidMount() {
+    this.setInputValidity();
+  }
+
+  componentDidUpdate() {
+    this.setInputValidity();
+  }
 
   render() {
     const { chips, description, disabled, error, label, forwardedRef, onChipDelete, ...props } = this.props;
     const id = this.getId();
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Input/Input.tsx
```diff
@@ -40,10 +40,29 @@ class StyleableInput extends React.PureComponent<InputProps & PrivateProps, Inpu
...
+  }
+
+  componentDidUpdate() {
+    this.setInputValidity();
+  }
 
   render() {
     const { chips, description, disabled, error, label, forwardedRef, onChipDelete, ...props } = this.props;
     const id = this.getId();
+    let downstreamRef: React.RefObject<HTMLInputElement> | Ref<HTMLInputElement>;
```
deini: 🍹 Can we move this to a `getDownstreamRef` fn?

rtalvarez:
So, my vscode broke and I wrote the unit tests using vim.. hence the TS errors there. Fixing 🚧

rtalvarez:
- offline conversations: we're moving away from this route for various reasons. maybe biggest one  is components should be kept relatively dumb, and it seems this path is just making them smarter and more complex. we'll remove the existing logic in a separate PR -

PR #213, opened 2019-10-15
Update Table component API

deini:
Update `<Table>` component for an easier and more restrictive API.

Shoutouts to @icatalina @RobinOng @chanceaclark 🏆for their immense help and feedback along the way.

localhost_3000_table-basic (1)

PS: PR is pretty big since `<PropTable>` got an API update too, which touches all the component docs.


packages/docs/components/PropTable/Prop/index.tsx
```diff
@@ -15,28 +13,27 @@ interface Props {
...
-        <TypesData types={types} />
-      </StyledTableCell>
-      <StyledTableCell>
-        <Code highlight={false}>{defaults}</Code>
-      </StyledTableCell>
-      <StyledTableCell>{children}</StyledTableCell>
-    </Table.Row>
-  );
+  return null;
+  // <Table.Row>
```
RobinOng: Is this comment not removed intentionally?

deini: Not intentional, I need to remove this file.

packages/docs/pages/Table/TablePage.tsx
```diff
@@ -0,0 +1,166 @@
...
+        {/* jsx-to-string:end */}
+      </CodePreview>
+
+      <H1>API</H1>
+
+      <H2>Table</H2>
+
+      <TablePropTable />
+
+      <H2 id="table-columns-prop-table">Table[Columns]</H2>
```
RobinOng: Just thinking maybe `PropTable` can accept `title` param to include? So that if we decide to change how the title looks like, we do not need to go to all the component to change them

deini: Good call, I initially just made it compatible with the new `<Table>` component but this is solid advice.

I addressed it in a new commit, added a `title` and `collapsible` prop since when we render "Inherited Props" tables we show the title in a different way and collapse the table by default.

packages/docs/pages/Table/TablePage.tsx
```diff
@@ -0,0 +1,154 @@
...
+            { sku: 'CGLD', name: '[Sample] Laundry Detergent', stock: 29 },
+          ]}
+        />
+        {/* jsx-to-string:end */}
+      </CodePreview>
+
+      <H1>API</H1>
+      <TablePropTable />
+      <TableColumnsPropTable id="table-columns-prop-table" />
+      <TableSelectablePropTable id="table-selectable-prop-table" />
```
jorgemoya: Why the `ids`?

packages/docs/PropTables/TablePropTable.tsx
```diff
@@ -1,45 +1,117 @@
...
+export const TablePropTable: React.FC<PropTableWrapper> = props => (
+  <PropTable title="Table" propList={tableProps} {...props} />
 );
 
-export const TableBodyPropTable: React.FC = () => (
-  <Text>
-    Supports all native <Code>&lt;tbody&nbsp;/&gt;</Code> element attributes.
-  </Text>
+export const TableColumnsPropTable: React.FC<PropTableWrapper> = props => (
+  <PropTable title="Table[Columns]" propList={tableColumnsProps} {...props} />
```
jorgemoya: Not sure if we I'm 100% with using `Table[]`. Shouldn't `Columns` be enough?

packages/docs/pages/Table/TablePage.tsx
deini: I'm using them to scroll to them from `<TablePropTable>`

packages/big-design/src/components/Table/Item/Item.tsx
```diff
@@ -0,0 +1,34 @@
...
+  onItemSelect?(nextValue: boolean): void;
+}
+
+export const Item: React.FC<ItemProps> = memo(({ children, isSelectable, selected = false, onItemSelect }) => {
+  const tableSectionContext = useContext(TableSectionContext);
+
+  const handleSelect = () => {
+    if (typeof onItemSelect === 'function') {
+      onItemSelect(!selected);
+    }
```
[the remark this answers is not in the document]
deini: In this case I'm doing it since we don't need to return anything. It has an implicit return type of `void`. If we change it to `return typeof onItemSelect === 'function' && onItemSelect(!selected)` then it would return a boolean | whatever `onItemSelect` returns (in this case void). 

We use it it in our components jsx a ton because if you return `false` react will just not render anything from it.

packages/big-design/src/components/Table/Table.tsx
```diff
@@ -1,54 +1,115 @@
...
+import { Body } from './Body';
+import { Cell } from './Cell';
+import { Head } from './Head';
+import { Item } from './Item';
+
+export const Table = <T extends TableItem>(props: TableProps<T>): React.ReactElement<TableProps<T>> => {
+  const {
+    className,
+    stickyHeader,
+    style,
```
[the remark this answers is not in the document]
deini: Yes, we should only allow them in our "Util" components (Flex / Box / Grid, etc).

Most of the components have tests around this, should probably add them to `Dropdown` too.

packages/big-design/src/components/Table/Item/Item.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { memo, useContext } from 'react';
```
chanceaclark: Why did you end up changing the name to `Item` instead of `Row`? `Row`, to me at least, is more verbose of what you are actually using in a `Table` component. `Item` could be anything.

packages/big-design/src/components/Table/Actions/Actions.tsx
```diff
@@ -1,18 +1,70 @@
...
+      const totalItemsInPage = items.length;
+      const isChecked = totalSelectedItems === totalItemsInPage && totalItemsInPage > 0;
+      const isIndeterminate = totalSelectedItems > 0 && totalSelectedItems !== totalItemsInPage;
+
+      return (
+        <Flex.Item flexGrow={2}>
+          <Flex flexDirection="row">
+            <Checkbox isIndeterminate={isIndeterminate} checked={isChecked} onChange={handleSelectAll} />
+            <Text marginLeft="small">
+              {totalSelectedItems}/{totalItemsInPage} {itemType}
```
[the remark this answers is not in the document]
deini: It currently only shows `0/10 Products` when the user starts selecting items.

packages/big-design/src/components/Table/Actions/Actions.tsx
```diff
@@ -1,18 +1,70 @@
...
+    return (
+      <StyledActions
+        alignItems="center"
+        aria-controls={tableId}
+        justifyContent="stretch"
+        padding="small"
+        flexDirection="row"
+        {...props}
+      >
+        {selectable && renderSelectAllAction(selectable)}
```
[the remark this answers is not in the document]
deini: Currently, we don't support any titles as the designs don't have any yet. Do you have any examples I can 👀 ?

packages/big-design/src/components/Table/Item/Item.tsx
deini: I changed it since we accept an array of `Items`, so it made sense to me to have an `Item` component,  I don't mind changing it to back `Row` tho (just the component, not the prop).

packages/big-design/src/components/Table/Actions/Actions.tsx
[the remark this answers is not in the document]
chanceaclark: Yeah come see offline.

packages/big-design/src/components/Table/Item/Item.tsx
chanceaclark: I feel like we should change it over to `Row` internally.

packages/docs/PropTables/TablePropTable.tsx
deini: 🤔yeah I'm not sure whats the best name for these, I do want to somehow how hierarchy tho, this way if `Columns` gets imported and rendered somewhere else it's pretty explicit it belongs to Table.

packages/docs/pages/Table/TablePage.tsx
```diff
@@ -0,0 +1,154 @@
...
+const columns = [
+  { header: 'Sku', render: ({ sku }) => sku },
+  { header: 'Name', render: ({ name }) => name },
+  { header: 'Stock', render: ({ stock }) => stock },
+];
+
+export default () => {
+  return (
+    <>
+      <H0>Table</H0>
```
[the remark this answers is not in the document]
deini: Good catch, it's missing because I forgot about it 😂 . I'll wait for the guidelines tho 👍

packages/big-design/src/components/Table/Actions/Actions.tsx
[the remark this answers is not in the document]
jorgemoya: Was this resolved?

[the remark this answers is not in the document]
deini: Not yet, design is thinking about it.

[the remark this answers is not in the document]
deini: It's resolved now 🕺

PR #216, opened 2019-10-21
feat(component): sortable functionality for Table

deini:
image

Adds `sortable` props to `<Table>` component.

I will do a follow-up PR to address performance optimizations.


packages/big-design/src/components/Table/types.ts
```diff
@@ -9,15 +9,25 @@ export interface TableSelectable<T> {
   onSelectionChange(selectedItems: T[]): void;
 }
 
+export type TableSortDirection = 'ASC' | 'DESC';
```
chanceaclark: I've seen some tables out there that have 3 states: Default, Ascending, and Descending.

As an example of what I'm talking about, see the `Birth Place` column on this table: https://material-ui.com/components/tables/#material-table.

Is our sortable columns only going to have 2 states? Is this what design intended?

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }: HeaderCellProps) => {
+    const renderSortIcon = () => {
+      if (!isSorted) {
+        return null;
+      }
+
+      return sortDirection === 'ASC' ? (
+        <ArrowUpwardIcon size="medium" data-testid="asc-icon" />
+      ) : (
+        <ArrowDownwardIcon size="medium" data-itestid="desc-icon" />
```
chanceaclark: You may want to add `title` props to these SVG's

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -0,0 +1,68 @@
...
+    stickyHeader,
+    width,
+  }: HeaderCellProps) => {
+    const renderSortIcon = () => {
+      if (!isSorted) {
+        return null;
+      }
+
+      return sortDirection === 'ASC' ? (
+        <ArrowUpwardIcon size="medium" data-testid="asc-icon" />
```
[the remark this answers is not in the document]
deini: I spoke with design last week and Table sorting hasn't been finalized yet, current designs are explorations. So I'm basing this on the most common pattern and we can update accordingly when our designs are ready.

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
deini: Good catch.

packages/big-design/src/components/Table/types.ts
deini: You can achieve that third state by not setting a `sortable.columnHash`.

chanceaclark: When you remove that key/val, the arrow drops off.

deini: Yeah since it's no longer `ASC` or `DESC`, it is kind of confusing pressing it and the arrow not changing. The user doesn't really know / cares if its the default or not 🤔

chanceaclark: As a user, I would care that the arrow appears, specifically to show which order the column is sorted in. Most tables I've seen or used has 3 different states out of the box. Since you mentioned that design is still in-between these patterns/designs, is this something we want to wait for design to finish? If not, we can accept only two states for now.

chanceaclark:
Also, what are your thoughts around sorting asynchronously?

packages/big-design/src/components/Table/types.ts
deini: Can you link me to the tables you are referring to other than material-ui?

I don't agree with a third state, however, adding it later is an additive an easy change.

deini:
Async sorting is handled by the app, we can expose an `isLoading` prop eventually to show a spinner.

packages/big-design/src/components/Table/types.ts
chanceaclark: Carbon is probably the best one I found:
https://www.carbondesignsystem.com/components/data-table/code

Both Atlaskit and Polaris intially default to no sorting, then it toggles sort on click.
https://atlaskit.atlassian.com/packages/core/dynamic-table
https://polaris.shopify.com/components/lists-and-tables/data-table#navigation

I also noticed with that with those two that it removes the arrow when sorting another column. I tested your PR locally and noticed that it sorts the column, but the arrow stays on the initial column.

chanceaclark: >I also noticed with that with those two that it removes the arrow when sorting another column. I tested your PR locally and noticed that it sorts the column, but the arrow stays on the initial column.

I think this is due to not being able to pass in multiple columnHashes and sort functions.

deini: So we have the same approach as Atlaskit and Polaris then, right?

To move the icon to the new column you need to change `columnHash` to the column that you need.

chanceaclark: > So we have the same approach as Atlaskit and Polaris then, right?

Atlaskit and Polaris default it to unsorted, as in unsorted data returned from an API. Whereas your implementation defaults the sort to `ASC` or `DSC` based off the `direction`, there's no default unsorted state.

> To move the icon to the new column you need to change columnHash to the column that you need.

Gotcha, then that's an implementation detail.

deini: That just depends on how you use it, you can do an unsorted state by just not passing a `columnHash` since you aren't sorting any column.

chanceaclark: So if we are starting to restrict functionality, why are we not baking that into the sort functionality itself rather than letting the app have to think about the logic of switching the `columnHash` and the sort function?

deini: Because it's a 1-liner for the app and if we add it to the library then it will have to have its own state and can potentially be toggling between ASC / DESC without the app actually doing the sorting and getting out of sync.

chanceaclark: 🤷‍♀ I'll take it

deini:
Offline chat with @chanceaclark update:
We will introduce separate examples to clarify use cases for sortable / selectable / paginable.

PR #218, opened 2019-10-23
Restrict Select

jorgemoya:
Now accepts an array of `options` to display the items. Also an `action` object. Changed props to use `onChange`.

select


packages/big-design/src/components/Select/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  onChange(value: string | number | Array<string | number>): void;
+}
+
+interface BaseItem extends Omit<ListItemProps, 'children' | 'content' | 'onClick' | 'value'> {
+  content: string;
+  icon?: React.ReactElement;
+}
+
+export interface Option extends Omit<BaseItem, 'actionType'> {
+  value: string | number;
```
deini: With the new approach `value` should be able to accept any value right? Is there a reason you are limiting it to `string | number`?

jorgemoya: @chanceaclark and I talked about this yesterday. Since `Select` is a form element we felt it should adhere to the native behavior.

deini: I don't think it would complicate multi-selects, maybe we can chat about it offline?

I think it's valuable to be able to pass whatever I want as a value 🤔. Especially since you aren't doing any runtime check to verify its a string | number.

rtalvarez: Hmm, I haven't crossed into a case where I wanted to pass something that is not a string or a number. I would imagine that would cover the majority of use cases? It would be nice to be able to pass whatever, but is that needed for right now?

packages/big-design/src/components/Select/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  options: Option[];
+  placement?: Placement;
+  positionFixed?: boolean;
+  required?: boolean;
+  value?: string | number | Array<string | number>;
+  onChange(value: string | number | Array<string | number>): void;
+}
+
+interface BaseItem extends Omit<ListItemProps, 'children' | 'content' | 'value'> {
+  content: string;
```
RobinOng: Just wondering whether we should allow `content` to take in component as well? One of the use case is 
Screen Shot 2019-10-24 at 10 41 25 am

deini: @RobinOng What do you think instead of accepting a `component` we add another prop `subContent` / `subText` or something like that, to keep consistency as much as possible. If we end up getting a ton of different designs we can fallback to `component` in the future.

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -228,31 +190,13 @@ export class Select extends React.PureComponent<SelectProps, SelectState> {
...
-        const selectedItem = listItems.find(item => item && String(item.value) === String(val));
-
-        return selectedItem && selectedItem.text;
-      });
-    } else {
-      const listItem = listItems.find(item => item && String(item.value) === String(value));
-
-      return listItem && [listItem.text];
-    }
+    return Array.isArray(values) ? values.map(value => options.find(option => option.value === value)) : [];
```
chanceaclark: 🍹You could default `values` to an empty `[]` and remove the `Array.isArray` check.

packages/big-design/src/components/Select/types.ts
RobinOng: The `subContent` / `subText` approach sounds much better 👍

packages/big-design/src/components/Select/Select.tsx
jorgemoya: Doesn't seem to work, since values can be of type `T` or `T[]`.

packages/docs/PropTables/DropdownPropTable.tsx
```diff
@@ -33,10 +33,10 @@ const dropdownProps: Prop[] = [
   },
   {
     name: 'options',
-    types: 'Array<DropdownItem | DropdownLinkItem>',
+    types: 'Array<Item | LinkItem>',
     description: (
       <>
-        Accepts an array of <Code>DropdownItems</Code> and <Code>DropdownLinkItems</Code>. See example for usage.
+        Accepts an array of <Code>Items</Code> and <Code>LinkItems</Code>. See example for usage.
```
rtalvarez: 🍹 Should this be `Array of Items _or_ LinkItems` ?

jorgemoya: Good observation. The Dropdown actually accepts a mix of both.

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -263,95 +190,130 @@ export class Select extends React.PureComponent<SelectProps, SelectState> {
...
+        </ListItem>
+      </Box>
+    );
+  }
+
+  private renderIcon(action: Action, isHighlighted: boolean) {
+    return (
+      React.isValidElement(action.icon) &&
+      React.cloneElement(action.icon, {
+        color: action.disabled
```
chanceaclark: 🍹Could you refactor these ternary statements into something a little more legible?

jorgemoya: Tried to make it more legible. Let me know if you have a better idea.

packages/docs/PropTables/SelectPropTable.tsx
```diff
@@ -1,16 +1,22 @@
 import React from 'react';
 
-import { Code, Prop, PropTable, PropTableWrapper } from '../components';
+import { Code, NextLink, Prop, PropTable, PropTableWrapper } from '../components';
 
 const selectProps: Prop[] = [
+  { name: 'action', types: 'Action', description: 'Action option displayed at the end of the list.' },
```
RobinOng: 🍹 Should this be multi-line?

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -111,17 +111,19 @@ export class Dropdown extends React.PureComponent<DropdownProps, DropdownState>
...
 
-  private renderIcon(option: DropdownItem | DropdownLinkItem, isHighlighted: boolean) {
+  private renderIcon(item: Item<T> | LinkItem<T>, isHighlighted: boolean) {
+    const getColor = () => {
+      if (item.disabled) {
+        return 'secondary40';
+      } else {
+        return isHighlighted ? (item.actionType === 'destructive' ? 'danger50' : 'primary') : 'secondary60';
+      }
+    };
```
deini: 🍹

```js
const getColor = () => {
  if (item.disabled) {
    return 'secondary40';
  }

  if (!isHighlighted) {
    return 'secondary60';
  }

  return item.actionType === 'destructive' ? 'danger50' : 'primary';
}
```

deini: 🍹 I'd also move `getColor` to a method in the `Dropdown` class.

packages/big-design/src/components/Dropdown/types.ts
```diff
@@ -2,25 +2,25 @@ import { Placement } from 'popper.js';
...
   content: string;
   icon?: React.ReactElement;
-  value?: any;
-  onClick?(item: DropdownItem | DropdownLinkItem): void;
+  value?: T;
+  onClick?(item: Item<T> | LinkItem<T>): void;
 }
 
-export interface DropdownItem extends BaseItem {
+export interface Item<T> extends BaseItem<T> {
```
deini: Is there a reason we renamed this to `Item` instead of `DropdownItem`? `Item` is too generic IMO. Same for `LinkItem`.

packages/big-design/src/components/Select/Select.tsx
chanceaclark: I was thinking of something along the lines of what @deini put below.

packages/big-design/src/components/Dropdown/Dropdown.tsx
jorgemoya: The reason why I didn't move it to it's own method in the class is because it needs access to `item` and `isHighlighted`. Would you still do it that way and just pass the params again?

PR #220, opened 2019-10-23
test(component): add tests for Tabs component

chanceaclark:
## What
Add missing tests for `Tabs` component.

### DIFF
--- a/packages/big-design/src/components/Tabs/spec.tsx
+++ b/packages/big-design/src/components/Tabs/spec.tsx
@@ -1,0 +1,99 @@
+import { fireEvent, render } from '@testing-library/react';
+import 'jest-styled-components';
+import React from 'react';
+
+import { Tabs, TabItem } from './Tabs';
+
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Tabs/spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  const { getByText, rerender } = render(<Tabs activeTab={activeTab} items={items} onTabClick={setActiveTab} />);
+
+  let trigger = getByText('Tab 2');
+
+  expect(activeTab).toBe('tab1');
+
+  fireEvent.click(trigger);
+
+  expect(setActiveTab).toHaveBeenCalledTimes(1);
```
deini: 🍹 You could change these to use a vanilla `jest.fn` and `toHaveBeenCalledWith`. That way you don't have to manually set `activeTab` and check for its value.

PR #222, opened 2019-10-24
refactor(component): table optimizations

deini:
`<Table />` optimizations 🏎


packages/big-design/src/utils/useEventCallback.ts
```diff
@@ -0,0 +1,22 @@
+import { useCallback, useEffect, useRef } from 'react';
+
+type Callback<T> = (...args: T[]) => void;
+
+export function useEventCallback<T>(fn: Callback<T>, dependencies: any[]) {
```
chanceaclark: Super 🌶 🔥

PR #224, opened 2019-10-24
feat(component): allow state override of table select all checkbox

deini:
- allow state override of table select all checkbox
- `itemName` prop nos on root level


packages/big-design/src/components/Table/Actions/Actions.tsx
```diff
@@ -31,20 +34,48 @@ export const Actions = memo(
...
-    const renderSelectAllAction = ({ itemType, selectedItems }: TableSelectable<T>) => {
+    const getSelectAllChecked = () => {
+      if (!selectable) {
+        return false;
+      }
+
+      const { selectAllState, selectedItems } = selectable;
+
+      switch (selectAllState) {
+        case 'ALL':
```
rtalvarez: 🍹 consider putting these strings in a constants definition or similar

packages/big-design/src/components/Table/Actions/Actions.tsx
```diff
@@ -70,7 +111,8 @@ export const Actions = memo(
         padding="small"
```
chanceaclark: 🍹Could you try and remove the `style={{ marginLeft: 'auto' }}` on the pagination, and add `justifyContent="space-between"` to `StyledActions`? We should try and clean up those `style` attributes.

deini: In this case, we have multiple actions, one of them being the `itemName` so if we do `space-between` we won't be getting what we want :(

However, I can move this to a custom-styled Flex.

We should add `auto` to `MarginProps`

chanceaclark: Ahh yeah, well in that case maybe you could change the `Flex.Item` for both SelectAll and ItemName to have `alignSelf="flex-start"` and Pagination `alignSelf="flex-end"`? 🤷‍♀️

PR #225, opened 2019-10-28
feat(component): table multi-page select

deini:
Get rid of `selectAllState` prop and support multi-page select by default.

### DIFF
--- a/packages/big-design/src/components/Table/Table.tsx
+++ b/packages/big-design/src/components/Table/Table.tsx
@@ -1,4 +1,4 @@
-import React, { memo, useRef } from 'react';
+import React, { memo, useCallback, useEffect, useRef, useState } from 'react';
 
 import { typedMemo, uniqueId } from '../../utils';
 import { useEventCallback } from '../../utils/useEventCallback';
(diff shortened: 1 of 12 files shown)


packages/big-design/src/components/Table/Table.tsx
```diff
@@ -36,31 +44,34 @@ const InternalTable = <T extends TableItem>(props: TableProps<T>): React.ReactEl
...
-          const { selectedItems, onSelectionChange } = selectable;
-          const nextIsSelected = !isItemSelected(item);
+          const { onSelectionChange } = selectable;
+          const nextIsSelected = !selectedItems.has(item);
 
           if (nextIsSelected) {
             onSelectionChange([...selectedItems, item]);
           } else {
-            onSelectionChange(selectedItems.filter(selectedItem => selectedItem !== item));
+            onSelectionChange([...selectedItems].filter(selectedItem => selectedItem !== item));
```
chanceaclark: `.filter` should return a new Array, any reason you destructed it?

deini: `selectedItems` in this case is a `Set` so I'm converting it to an array and filtering it.

PR #226, opened 2019-10-29
feat(docs): add helpful links section

chanceaclark:
## What
Adding the magical Github and NPM link to be visible on all pages.

## Screenshot
Screen Shot 2019-10-29 at 12 15 38 PM


chanceaclark:
>Is the npm link helpful? What do you think of having:
>
> - Github
> - Design Guidelines

I like that better. I can use the BD icon we already have.

>Can we make the icons the same color as the nav section titles?

Since we are using the `svgs` with an `img` we can't use the `currentColor` to change the fill.

PR #231, opened 2019-11-04
fix(component): table header not aligning properly

deini:
`align` prop on table headers broke after introducing sorting.

### DIFF
--- a/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
+++ b/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
@@ -50,14 +50,8 @@ const InternalHeaderCell = <T extends TableItem>({
   };
 
   return (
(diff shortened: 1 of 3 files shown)


deini:
@chanceaclark in this case, we are still using `align` so `props` didn't change. Unless I'm missing something 🤔

chanceaclark:
> @chanceaclark in this case, we are still using `align` so `props` didn't change. Unless I'm missing something 🤔

It's now using Flex `justifyContent` keys instead of `text-align`.

deini:
@chanceaclark Yeah, but we still want to only want to expose `left / center / right`.

chanceaclark:
> @chanceaclark Yeah, but we still want to only want to expose `left / center / right`.

Sounds good then, if you think it's beneficial, we may want to runtime checks 🤷‍♀

Otherwise, LGTM.

PR #233, opened 2019-11-04
feat: CHI-696 select component accepts an inputRef prop

rtalvarez:
## What?
The `Select` component now accepts an optional `inputRef` of type `RefObject<HTMLInputElement> | React.Ref<HTMLInputElement>`

## Usage:
The prop accepts three usages:
- `inputRef={undefined}`
- `inputRef={React.createRef<HTMLInputElement>()}`
- `inputRef={(input: HTMLInputElement) => void}`

### DIFF
--- a/packages/big-design/src/components/Select/Select.tsx
+++ b/packages/big-design/src/components/Select/Select.tsx
@@ -39,7 +39,9 @@ export class Select<T extends any> extends React.PureComponent<SelectProps<T>, S
     selectedOptionContent: '',
   };
 
-  private inputRef: RefObject<HTMLInputElement> = React.createRef();
+  private callbackRef?: HTMLInputElement;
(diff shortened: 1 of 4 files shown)


jorgemoya:
Need to add new `prop` to the docs. 🙏

deini:
Can we use `forwardRef` 🙏

rtalvarez:
👷 🔨

rtalvarez:
Offline: this prop name will remain as `inputRef` as opposed to `ref` (by using `forwardRef`) because it is not possible to pass generic type arguments to `React.forwardRef` as of right now, so the type `T` for the select options would be lost by doing this:

```javascript
export const Select = React.forwardRef<HTMLInputElement, SelectProps<any>>( ... );
```

Various workarounds for this did not yield acceptable results

PR #234, opened 2019-11-06
chore(all): use Babel and TS 3.7

deini:
Transform TS with Babel + Upgrade to TS 3.7 

**Note: Node >= 10.16 is now required.**

### Why?
This is the first step for our tree-shaking goals:
```js
import { Button } from '@bigcommerce/big-design';
```
Should only bring `Button`'s code.

Direct import goal (useful for apps that don't support tree-shaking):
```js
import Button from '@bigcommerce/big-design/Button';
```

Our current rollup setup using the ts plugin is pretty slow, which if we want to run the build for every single component (direct-import goal) it's going to be pretty slow. By switching to babel we get faster builds which will also make the dev experience a bit better. We still run `tsc` to check for type errors.

Misc changes:
- Added `isolatedModules` flag.
- Refactored how we export types.
- Testing out some `PURE` annotations and side-effects free code for tree-shaking.

Would love some help testing this 🙏 ❤️


packages/big-design/src/index.ts
```diff
@@ -1,2 +1,2 @@
 export * from './components';
-export * from './mixins';
+// export * from './mixins';
```
jorgemoya: Remove comment?

packages/docs/components/SideNav/styled.tsx
```diff
@@ -2,7 +2,7 @@ import { Flex } from '@bigcommerce/big-design';
 import styled from 'styled-components';
 
 export const StyledFlex = styled(Flex)`
-  ${({ theme }) => theme.shadow.raised}
+  ${({ theme }) => theme.shadow.raised};
```
chanceaclark: 🍹 You shouldn't need to add a semi-colon here. `theme.shadow.raised` already outputs semi's on each CSS declaration already.
```suggestion
  ${({ theme }) => theme.shadow.raised}
```

PR #241, opened 2019-11-07
feat(component): add display prop to utility components

chanceaclark:
## What
Adds a `display` prop to some of our utility components. Useful for layouts that require responsive displays (i.e. "none" on mobile, "block" on table+) and other usefulness.

## Screenshot
screencapture-localhost-3000-display-2019-11-07-15_44_39

### DIFF
--- a/packages/big-design/src/mixins/display/__snapshots__/spec.tsx.snap
+++ b/packages/big-design/src/mixins/display/__snapshots__/spec.tsx.snap
@@ -1,0 +1,23 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`responsive display 1`] = `
+@media (min-width:0px) {
+  .c0 {
+    display: none;
+  }
+}
+
+@media (min-width:720px) {
+  .c0 {
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+  }
+}
+
+<div
+  class="c0"
+  display="[object Object]"
+/>
+`;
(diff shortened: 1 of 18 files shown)


packages/big-design/src/mixins/display/__snapshots__/spec.tsx.snap
```diff
@@ -0,0 +1,23 @@
...
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+  }
+}
+
+<div
+  class="c0"
+  display="[object Object]"
```
chanceaclark: @deini it looks like this will be a thing for `display` not sure if we want to rethink this?

packages/big-design/src/components/Flex/Flex.tsx
```diff
@@ -1,12 +1,13 @@
...
 
+import { DisplayProps } from '../../mixins';
 import { BoxProps } from '../Box';
 
 import { StyledFlex } from './styled';
 import { FlexedProps } from './types';
 import { FlexItem } from './Item';
 
-export type FlexProps = BoxProps & FlexedProps;
+export type FlexProps = BoxProps & DisplayProps & FlexedProps;
```
deini: `BoxProps` should include `DisplayProps` right?

packages/big-design/src/components/Grid/Grid.tsx
```diff
@@ -1,12 +1,13 @@
...
 
+import { DisplayProps } from '../../mixins';
 import { BoxProps } from '../Box';
 
 import { StyledGrid } from './styled';
 import { GridedProps } from './types';
 import { GridItem } from './Item';
 
-export type GridProps = BoxProps & GridedProps;
+export type GridProps = BoxProps & DisplayProps & GridedProps;
```
deini: Same as above

Issue #236, opened 2019-11-07
Feature Req: Display treatment for "external" link


bookernath:
**Is your feature request related to a problem? Please describe.**
I can't clearly communicate to users the difference between internal "navigation" links and external links with the native capabilities of BigDesign.

**Describe the solution, feature, or improvement you'd like**
Sometimes, you need to create links that will open in a new tab, to link to an external resource like documentation. It's nice to hint to users that this will happen with some sort of display treatment like this:

![image](https://user-images.githubusercontent.com/8922457/68391685-c9dbc200-012d-11ea-83ae-0ac2cb905086.png)

This sort of visual language makes the user less surprised when we open a new tab instead of navigate.

It seems like a reasonable extension to the `<Link>` component API. Perhaps it can be an explicitly set flag, or perhaps it can implicitly react to use of `target="_blank"` and so on.

**Describe alternatives you've considered**
Styling the link by hand.

deini:
Fixed by #243

PR #243, opened 2019-11-08
feat(component): add icon props to link component

chanceaclark:
## What
Add the ability to pass icons to `Link` components. Should help with #236.

## Screenshot
Screen Shot 2019-11-08 at 12 40 08 PM

### DIFF
--- a/packages/big-design/src/components/Link/Link.tsx
+++ b/packages/big-design/src/components/Link/Link.tsx
@@ -1,17 +1,28 @@
+import { OpenInNewIcon } from '@bigcommerce/big-design-icons';
 import React, { memo, Ref } from 'react';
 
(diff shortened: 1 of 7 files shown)


deini:
Instead of icon props what do you think of a Boolean prop that would render the appropriate open in new tab icon with the right size and placement? Also only allowing it on target `_blank`?

chanceaclark:
> Instead of icon props what do you think of a Boolean prop that would render the appropriate open in new tab icon with the right size and placement? Also only allowing it on target `_blank`?

Do you see any possibility of Design not just using the `OpenInNewIcon`?

deini:
Not at the moment. We should call the prop something related to that icon/action. We can always add a generic icon prop in the future if needed.

PR #244, opened 2019-11-08
fix(docs): opening links in new tab

chanceaclark:
## What
Fix the ability to open links in new tabs.


chanceaclark:
> Can these be `”#”`?

We can, but it'll show up in the url upon navigation.

PR #245, opened 2019-11-09
fix(component): grid and flex forward as prop

deini:
`Grid` and `Flex` internally extend `Box` using `styled(Box)`. The issue is now that we introduced the `as` prop for util components when doing something like:
 ```jsx
<Flex as="section" />
```
This will swap that `styled(Box)` with `styled('section')`, which in this case is not exactly what we want since we lose all the functionality from `Box` like `margin` props, etc.

styled-components `4.3` introduced a `forwardedAs` prop for this specific use case. So doing something like this would work:
```jsx
<Flex forwardedAs="section" />
```

However, it is inconsistent with `Box` and requires the user to know how `Flex` and `Grid` are implemented. This PR makes `Grid` and `Flex` accept an `as` prop but internally swap them with `forwardedAs`.

Gotchas:
- Breaking change since had to bump SC peer dep version
- `@types/styled-components` doesn't have the definition for `forwardedAs` prop, in the meantime, I'm manually typing it, we should be able to remove the manual type in the future.

### DIFF
--- a/packages/big-design/src/components/Flex/Item/styled.tsx
+++ b/packages/big-design/src/components/Flex/Item/styled.tsx
@@ -6,7 +6,8 @@ import { withFlexedItems } from '../withFlex';
 
 import { FlexItemProps } from './Item';
 
-export const StyledFlexItem = styled(Box)<FlexItemProps>`
+// TODO: Remove the `forwardedAs` manual prop definition when @types get updated
+export const StyledFlexItem = styled(Box)<FlexItemProps & { forwardedAs?: FlexItemProps['as'] }>`
   ${withFlexedItems()}
 `;
 
--- a/packages/big-design/src/components/Flex/styled.tsx
+++ b/packages/big-design/src/components/Flex/styled.tsx
@@ -7,7 +7,8 @@ import { Box } from '../Box';
 import { withFlexedContainer } from './withFlex';
 import { FlexProps } from './Flex';
 
-export const StyledFlex = styled(Box)<FlexProps>`
+// TODO: Remove the `forwardedAs` manual prop definition when @types get updated
+export const StyledFlex = styled(Box)<FlexProps & { forwardedAs?: FlexProps['as'] }>`
   ${withFlexedContainer()}
 
   display: flex;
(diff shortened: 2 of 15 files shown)


packages/big-design/src/components/Flex/Item/styled.tsx
```diff
@@ -6,7 +6,8 @@ import { withFlexedItems } from '../withFlex';
 
 import { FlexItemProps } from './Item';
 
-export const StyledFlexItem = styled(Box)<FlexItemProps>`
+// Remove the `forwardedAs` manual prop definition when @types get updated
```
chanceaclark: 🍹Can we prefix these with `TODO:` so we can easily search for them?
```suggestion
// TODO: Remove the `forwardedAs` manual prop definition when @types get updated
```

packages/big-design/src/components/Flex/styled.tsx
```diff
@@ -7,7 +7,8 @@ import { Box } from '../Box';
 import { withFlexedContainer } from './withFlex';
 import { FlexProps } from './Flex';
 
-export const StyledFlex = styled(Box)<FlexProps>`
+// Remove the `forwardedAs` manual prop definition when @types get updated
```
chanceaclark: 🍹
```suggestion
// TODO: Remove the `forwardedAs` manual prop definition when @types get updated
```

packages/big-design/src/components/Grid/Item/styled.tsx
```diff
@@ -5,6 +5,7 @@ import { withGridedItems } from '../withGrid';
 
 import { GridItemProps } from './Item';
 
-export const StyledGridItem = styled(Box)<GridItemProps>`
+// Remove the `forwardedAs` manual prop definition when @types get updated
```
chanceaclark: 🍹
```suggestion
// TODO: Remove the `forwardedAs` manual prop definition when @types get updated
```

packages/big-design/src/components/Grid/styled.tsx
```diff
@@ -7,7 +7,8 @@ import { Box } from '../Box';
 import { withGridedContainer } from './withGrid';
 import { GridProps } from './Grid';
 
-export const StyledGrid = styled(Box)<GridProps>`
+// Remove the `forwardedAs` manual prop definition when @types get updated
```
chanceaclark: 🍹
```suggestion
// TODO: Remove the `forwardedAs` manual prop definition when @types get updated
```

PR #251, opened 2019-11-11
feat(component): change Select onChange to be a synthetic event

chanceaclark:
## What
See `BREAKING CHANGE` below.

BREAKING CHANGE:
`onChange` is renamed to `onItemChange`


deini:
Offline chat: get rid of `onChange`

chanceaclark:
♻️ 🙏

PR #256, opened 2019-11-12
fix(component): table header right alignment

deini:
Fix `align: right` on a table header.


packages/big-design/src/components/Table/HeaderCell/styled.tsx
```diff
@@ -47,5 +53,17 @@ export const StyledTableHeaderCheckbox = styled(StyledTableHeaderCell)`
...
 `;
 
+export const StyledFlex = styled(Flex)<StyledFlexProps>`
+  ${({ align }) =>
+    align === 'right'
+      ? css`
+          justify-content: flex-end;
+        `
+      : css`
+          justify-content: ${align};
```
chanceaclark: `left` is not going to be the same as `flex-start` css layout wise, is this something we want to account for too?

PR #263, opened 2019-11-13
feat(component):  add actions prop to Table and StatefulTable

deini:
Allow `customActions` in `<Table>` and `<StatefulTable>`

Example:
image

### DIFF
--- a/packages/big-design/src/components/Table/types.ts
+++ b/packages/big-design/src/components/Table/types.ts
@@ -35,6 +35,7 @@ export interface TableColumn<T> {
 export type TablePaginationProps = Omit<PaginationProps, keyof MarginProps>;
 
 export interface TableProps<T> extends React.TableHTMLAttributes<HTMLTableElement> {
+  actions?: React.ComponentType<T>;
   columns: Array<TableColumn<T>>;
   itemName?: string;
   items: T[];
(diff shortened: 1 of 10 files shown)


packages/big-design/src/components/Table/types.ts
```diff
@@ -35,6 +35,7 @@ export interface TableColumn<T> {
 export type TablePaginationProps = Omit<PaginationProps, keyof MarginProps>;
 
 export interface TableProps<T> extends React.TableHTMLAttributes<HTMLTableElement> {
+  customActions?: React.ComponentType<T>;
```
chanceaclark: 🤔 should we rename this to something else? not sure if prepending `custom` adds any value

deini: I called it `custom` in case we eventually add more config-driven pre-defined `actions`. Thoughts?

chanceaclark: If that's the case, then `actions` could be appropriate. If it's config-driven, it's customizable by nature–prepending `custom` would seem verbose. From an api standpoint, we should probably keep it simple. If it's not descriptive enough, we should clarify more in the documentation.

deini: 🤔 not sure I want to mix config-driven with component-driven.

chanceaclark: 🤔 What about something along the lines of `actionsComponent` instead? `custom` would be implied.

deini: Changed it to your initial proposal of just calling it `actions` (internally still calling them `customActions`). If we add more config-based actions we can be more descriptive with their names and split them into multiple props. eg: `filterAction` / `menuAction` or something like that.

PR #264, opened 2019-11-13
feat: add hiddenLabel prop to Checkbox component

### LINKED ISSUE #203: Visually hidden `Checkbox` labels
For accessibility, we need a way to render a `Checkbox` component without a visual label. For instance, a selectable `Table.Row` needs a `Checkbox`, but with a visually-hidden label that specifies the actions to be made.

Open to recommendations but here are some possible solutions:

Add a `visuallyHidden` prop to the `Checkbox.Label` element and add conditional visually-hidden styles. This will allow someone have an implementation like so:
```tsx
<Checkbox label={<Checkbox.Label visuallyHidden>...</Checkbox.Label>} />
```

Another possible solution is to just add a prop to the `Checkbox` component:
```tsx
<Checkbox label="..." visuallyHidden />
```

**Additional context**
We changed the `label` prop to be optional during the meantime, but we'll need to make it required again with this improvement.

jordan-massingill:
#WHAT 
Adds `hiddenLabel` prop to Checkbox component to allow for visually hidden label, maintain accessibility. Closes #203.

#TESTING 
Tested locally to ensure prop allows component to render correctly

### DIFF
--- a/packages/big-design/src/components/Checkbox/Checkbox.tsx
+++ b/packages/big-design/src/components/Checkbox/Checkbox.tsx
@@ -10,7 +10,7 @@ import { CheckboxContainer, HiddenCheckbox, StyledCheckbox, StyledLabel } from '
 interface Props {
   hiddenLabel?: boolean;
   isIndeterminate?: boolean;
-  label?: React.ReactChild;
+  label: React.ReactChild;
   theme?: ThemeInterface;
 }
 
--- a/packages/big-design/src/components/Checkbox/Checkbox.tsx
+++ b/packages/big-design/src/components/Checkbox/Checkbox.tsx
@@ -8,6 +8,7 @@ import { uniqueId } from '../../utils';
 import { CheckboxContainer, HiddenCheckbox, StyledCheckbox, StyledLabel } from './styled';
 
 interface Props {
+  hiddenLabel?: boolean;
   isIndeterminate?: boolean;
   label?: React.ReactChild;
   theme?: ThemeInterface;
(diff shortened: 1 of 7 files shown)


packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -8,6 +8,7 @@ import { uniqueId } from '../../utils';
 import { CheckboxContainer, HiddenCheckbox, StyledCheckbox, StyledLabel } from './styled';
 
 interface Props {
+  hiddenLabel?: boolean;
   isIndeterminate?: boolean;
   label?: React.ReactChild;
```
chanceaclark: We probably want to add back `label` as required again. See #201 on why we made it optional. Afterwards, you may want to go into the table components and update the Checkboxes accordingly.

jordan-massingill: Yeah, I went back and forth on whether to do that or make it part of another ticket. If this all looks okay, I'll go ahead and update the table component too.

packages/big-design/src/components/Table/Row/Row.tsx
```diff
@@ -28,18 +30,20 @@ const InternalRow = <T extends TableItem>({
     }
   };
 
+  const label = isSelected ? `Deslect item number ${itemIndex + 1}` : `Select item number ${itemIndex + 1}`;
```
chanceaclark: Typo:
```suggestion
  const label = isSelected ? `Deselect item number ${itemIndex + 1}` : `Select item number ${itemIndex + 1}`;
```

🍹What about renaming to something like "Deselect row {number}"?

packages/big-design/src/components/Table/SelectAll/SelectAll.tsx
```diff
@@ -51,7 +51,13 @@ export const SelectAll = <T extends TableItem>({
   return (
     <Flex.Item marginRight="xxSmall">
       <Flex flexDirection="row">
-        <Checkbox isIndeterminate={someInPageSelected} checked={allInPageSelected} onChange={handleSelectAll} />
+        <Checkbox
+          isIndeterminate={someInPageSelected}
+          hiddenLabel
+          label="Select All"
```
chanceaclark: Might want to add some logic around deselecting all rows as well.

packages/big-design/src/components/Table/Row/Row.tsx
deini: Do we need the item index? The Row already has the context of the `item`, is there any benefits of having the index as part of the label?

jordan-massingill: This was a suggestion from @chanceaclark as an option for an accessible label, but the label for the checkbox was arbitrarily chosen and can be changed. Let's discuss offline 🌮

PR #274, opened 2019-11-15
Add transitions to components

capsula4:
Apply small transition for a smoother hover effect on inputs/buttons.

Depends on https://github.com/bigcommerce/big-design/pull/273

### DIFF
--- a/packages/big-design/src/components/Button/styled.tsx
+++ b/packages/big-design/src/components/Button/styled.tsx
@@ -2,11 +2,14 @@ import { addValues, theme as defaultTheme } from '@bigcommerce/big-design-theme'
 import styled, { css } from 'styled-components';
 
 import { withMargins, MarginProps } from '../../mixins';
+import { withTransition } from '../../mixins/transitions';
 import { Flex } from '../Flex';
 
 import { ButtonProps } from './index';
 
 export const StyledButton = styled.button<ButtonProps & MarginProps>`
+  ${withTransition(['background-color', 'border-color', 'box-shadow', 'color'])}
+
   && {
     ${withMargins()};
   }
--- a/packages/big-design/src/components/Checkbox/Checkbox.tsx
+++ b/packages/big-design/src/components/Checkbox/Checkbox.tsx
@@ -24,7 +24,7 @@ class RawCheckbox extends React.PureComponent<CheckboxProps & PrivateProps> {
   private readonly labelUniqueId = uniqueId('checkBox_label_');
 
   render() {
-    const { checked, className, isIndeterminate, label, forwardedRef, style, ...props } = this.props;
+    const { checked, className, disabled, isIndeterminate, label, forwardedRef, style, ...props } = this.props;
     const id = this.getInputId();
 
     return (
@@ -33,6 +33,7 @@ class RawCheckbox extends React.PureComponent<CheckboxProps & PrivateProps> {
           type="checkbox"
           checked={checked}
           id={id}
+          disabled={disabled}
           {...props}
           aria-labelledby={this.labelUniqueId}
           ref={checkbox => {
@@ -54,7 +55,13 @@ class RawCheckbox extends React.PureComponent<CheckboxProps & PrivateProps> {
           }}
         />
 
-        <StyledCheckbox isIndeterminate={isIndeterminate} checked={checked} htmlFor={id} aria-hidden={true}>
+        <StyledCheckbox
+          disabled={disabled}
+          isIndeterminate={isIndeterminate}
+          checked={checked}
+          htmlFor={id}
+          aria-hidden={true}
+        >
           {!checked && isIndeterminate ? <RemoveIcon /> : <CheckIcon />}
         </StyledCheckbox>
         {this.renderLabel()}
--- a/packages/big-design/src/components/Checkbox/styled.tsx
+++ b/packages/big-design/src/components/Checkbox/styled.tsx
@@ -2,11 +2,13 @@ import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import { hideVisually } from 'polished';
 import styled, { css, DefaultTheme, StyledComponent } from 'styled-components';
 
+import { withTransition } from '../../mixins/transitions';
 import { StyleableText } from '../Typography/private';
 
 interface StyledCheckboxProps {
   checked?: boolean;
   isIndeterminate?: boolean;
+  disabled?: boolean;
 }
 
 export interface StyledLabelProps {
@@ -24,6 +26,8 @@ export const HiddenCheckbox = styled.input`
 `;
 
 export const StyledCheckbox = styled.label<StyledCheckboxProps>`
+  ${withTransition(['border-color', 'background', 'box-shadow', 'color', 'opacity'])}
+
   align-items: center;
   background: ${({ checked, isIndeterminate, theme }) =>
     checked || isIndeterminate ? theme.colors.primary : theme.colors.white};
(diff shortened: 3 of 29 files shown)


chanceaclark:
Hey @capsula4, adding transitions/animations is definitely an action item. If we are to add them, we should probably add this as a full fledge feature instead and get it approved through design. What do you think?

capsula4:
@chanceaclark sure, will talk with Valter but I guess some transitions are better than no transitions at all. IMO we don't need to tackle everything at once.

capsula4:
@chanceaclark  as per discussed, hard-coded the transitions for now, until we figure out how we want to expose this properly via the theme

capsula4:
Note to self: only transition the specific properties

packages/big-design/src/components/Button/styled.tsx
```diff
@@ -30,6 +30,7 @@ export const StyledButton = styled.button<ButtonProps & MarginProps>`
   position: relative;
   text-align: center;
   text-decoration: none;
+  transition: background-color 150ms ease-out;
```
chanceaclark: For active, focus, and disabled state:
```suggestion
  transition: 150ms ease-out;
  transition-property: background-color, border-color, box-shadow, color;
```

packages/big-design/src/components/Link/styled.tsx
```diff
@@ -13,6 +13,15 @@ export const StyledLink = styled.a<LinkProps & { isExternal?: boolean }>`
...
   text-decoration: none;
+  transition: background-color 150ms ease-out;
+
+  &:active {
+    background-color: ${({ theme }) => theme.colors.primary60};
+  }
+
+  &:hover:not(:active) {
+    background-color: ${({ theme }) => theme.colors.primary50};
+  }
```
chanceaclark: Did you intend for this to be `color` instead of `background-color`?

packages/big-design/src/components/Checkbox/styled.tsx
```diff
@@ -32,25 +32,32 @@ export const StyledCheckbox = styled.label<StyledCheckboxProps>`
     checked || isIndeterminate ? theme.colors.primary : theme.colors.secondary30};
   border-radius: ${({ theme }) => theme.borderRadius.normal};
   color: ${({ theme }) => theme.colors.white};
+  cursor: pointer;
   display: inline-flex;
   height: ${({ theme }) => theme.spacing.large};
   justify-content: center;
-  transition: all 150ms;
+  transition: all 150ms ease-out;
```
chanceaclark: I know this was done before this PR, but could you explicitly set the transition instead of using `all`? 
```suggestion
  transition: 150ms ease-out;
  transition-property: border-color, box-shadow, color, visibility;
```

packages/big-design/src/components/Checkbox/styled.tsx
```diff
@@ -32,25 +32,32 @@ export const StyledCheckbox = styled.label<StyledCheckboxProps>`
...
   user-select: none;
   width: ${({ theme }) => theme.spacing.large};
 
+  &:hover {
+    border-color: ${({ checked, isIndeterminate, theme }) =>
+      checked || isIndeterminate ? theme.colors.primary : theme.colors.secondary40};
+  }
+
```
chanceaclark: This doesn't work well with `disabled` checkboxes. Can you account for this?

packages/big-design/src/components/List/Item/styled.tsx
```diff
@@ -13,6 +13,8 @@ export const StyledListItem = styled.li<ListItemProps>`
   min-width: ${({ theme }) => theme.helpers.remCalc(256)};
   outline: none;
   padding: 0 ${({ theme }) => theme.spacing.medium};
+  transition: all 150ms ease-out;
```
chanceaclark: Shouldn't need all if using `transition-property`
```suggestion
  transition: 150ms ease-out;
```

packages/big-design/src/components/ProgressBar/styled.tsx
```diff
@@ -18,7 +18,7 @@ export const StyledProgressBarFiller = styled.div<ProgressBarProps>`
   ${({ percent, theme }) =>
     typeof percent === 'number'
       ? css`
-          transition: width 0.2s ease-in;
+          transition: all 150ms ease-out;
```
chanceaclark: Instead of `all`:
```suggestion
          transition: 150ms ease-out;
          transition-property: left, width;
```

packages/big-design/src/components/Radio/styled.tsx
```diff
@@ -28,30 +29,41 @@ export const StyledRadio = styled.label<StyledRadioProps>`
...
   height: ${({ theme }) => theme.spacing.large};
   position: relative;
-  transition: all 150ms;
+  transition: border-color 150ms ease-out;
   user-select: none;
   width: ${({ theme }) => theme.spacing.large};
 
+  &:hover {
+    border-color: ${props => (props.checked ? props.theme.colors.primary50 : props.theme.colors.secondary40)};
+  }
```
chanceaclark: Same issue with checkboxes, doesn't work well with `disabled`.

packages/big-design/src/components/Textarea/styled.tsx
```diff
@@ -17,6 +17,7 @@ export const StyledTextarea = styled.textarea<TextareaProps>`
   line-height: ${({ theme }) => theme.lineHeight.medium};
   max-height: ${({ theme }) => theme.helpers.remCalc(224)};
   padding: ${({ theme }) => `${theme.spacing.xxSmall} ${theme.spacing.small}`};
+  transition: border 150ms ease-out;
```
chanceaclark: Focus state:
```suggestion
  transition: 150ms ease-out;
  transition-property: border, box-shadow;
```

packages/big-design/src/components/Checkbox/styled.tsx
capsula4: I have fixed it, however disabled checkboxes seem to be broken at the moment. You can still click on disabled checkboxes.

packages/big-design/src/components/Radio/styled.tsx
capsula4: at the moment there doesn't seem to be support for disabled radio buttons (no styling)

capsula4:
@chanceaclark thanks for the thorough review. I have addressed most of your comments, and created a mixing so things are not too hard-coded. Eventually the mixin could consume the values from the theme. 

Disabled state for checkbox is broken at the moment, maybe better to fix in a different PR.
Radio buttons do not have support for disabled state.

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -26,7 +26,7 @@ class RawCheckbox extends React.PureComponent<CheckboxProps & PrivateProps> {
   private readonly labelUniqueId = uniqueId('checkBox_label_');
 
   render() {
-    const { checked, className, isIndeterminate, label, forwardedRef, style, ...props } = this.props;
+    const { checked, className, disabled, isIndeterminate, label, forwardedRef, style, ...props } = this.props;
```
chanceaclark: You'll probably not want to destructure this as it's needed for the `HiddenCheckbox` component. Probably just want to use `this.props.disabled` on the `StyledCheckbox`.

packages/big-design/src/mixins/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './display';
 export * from './margins';
 export * from './paddings';
+export * from './transitions';
```
chanceaclark: I don't think we'll want to expose this publicly for now. After you remove it and update the imports, we'll be good to go!

capsula4: hopefully is all good now

chanceaclark:
Fix tests, but should be good to go.

PR #275, opened 2019-11-18
fix(component): properly handle Dropdown if all options are disabled

jorgemoya:
When all `options` are disabled, no items are "available" for selection. Now properly validate if `listItemRefs` has items.

### DIFF
--- a/packages/big-design/src/components/Dropdown/spec.tsx
+++ b/packages/big-design/src/components/Dropdown/spec.tsx
@@ -329,3 +329,19 @@ test("doesn't render tooltip on enabled item", () => {
 
   expect(queryByText(tooltipText)).not.toBeInTheDocument();
 });
+
+test('no errors expected if all options are disabled', () => {
+  const { getByRole } = render(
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Dropdown/spec.tsx
```diff
@@ -329,3 +329,19 @@ test("doesn't render tooltip on enabled item", () => {
...
+      options={[
+        { content: 'Option 1', value: '0', disabled: true },
+        { content: 'Option 2', value: '1', disabled: true },
+      ]}
+      trigger={<Button>Button</Button>}
+    />,
+  );
+
+  const trigger = getByRole('button');
+  fireEvent.click(trigger);
```
deini: Can you use `toThrow` to better capture what we are trying to test? https://jestjs.io/docs/en/expect#tothrowerror

jorgemoya: I looked for something like this and somehow missed it. 👍

PR #277, opened 2019-11-18
fix(docs): prevent hashes on links from being overwritten

jorgemoya:
This fixes the issue of `NextLinks` with hashes not working in live docs. 

I wasn't able to reproduce this issue locally while `yarn run start`, but verified they don't work on production.


deini:
I believe the issue is related to https://github.com/bigcommerce/big-design/pull/244 changes. What do you think about fixing it there so we can keep using the same component?

Issue #237, opened 2019-11-07
Doc request: A more composite example of a complete page with several components


bookernath:
**Is your feature request related to a problem? Please describe.**
I used BigDesign for the first time yesterday, and I found it a little harder to onboard to it than I'd like.

The biggest issue for me personally is that I was provided with a lot of low-level components, but no example of how to string them together into an overall page layout. I'd love to have something more copy-pasteable here to get a full page going quickly, that I could start to modify and make my own.

The open-source channels app was provided, but it's actually _quite_ complex relative to what I'm looking for.

A simple use case like a form accepting user data, or a getting-started page for an app would be closer to what I needed. This would serve as implicit documentation for how to arrange components on a page in a way that looks like and complied with the design guidelines. Currently, it seems like a user of BigDesign must exercise visual design skills themselves by comparing the design guidelines with the provided components to try to put something cohesive together. Can this be made easier?

**Describe the solution, feature, or improvement you'd like**
Some sort of "intermediate" code example that's between the existing component-level documentation and the pretty expansive Channels App.

deini:
Thanks for the feedback! What are your thoughts on having a link in our docs/readme to a [CodeSandbox](https://codesandbox.io/) running an example of create-react-app using `big-design`?

bookernath:
That sounds ideal.

amckemie:
JIRA - CHP-6165

PR #281, opened 2019-11-26
feat(example): setup CodeSandbox example

deini:
CodeSandbox simple example.

Closes https://github.com/bigcommerce/big-design/issues/237.


packages/example/src/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  }
+`;
+
+render(
+  <ThemeProvider theme={theme}>
+    <>
+      <AppGlobalStyles />
+      <GlobalStyles />
+      <App />
+    </>
```
jordan-massingill: Why use a fragment here?

deini: In this case we need it because `<ThemeProvider>` accepts only a single children node.

jordan-massingill: Ah makes sense - thanks! 👍

packages/docs/components/SideNav/SideNav.tsx
```diff
@@ -28,6 +29,8 @@ export const SideNav: React.FC = () => {
           <SideNavLink href="/GettingStarted/GettingStartedPage" as="/">
             Getting Started
           </SideNavLink>
+
+          <Link url={CodeSandboxUrl} title="CodeSandbox Example" />
```
chanceaclark: 🍹What about adding this to the `Getting Started` page instead of adding a link here? Or we could have both the link here, but a description about the sample app on the `Getting Started` page?

🍹Also, we may want to use the `external` link variant. Would require some modification to the `Link` below.

PR #283, opened 2019-12-02
[POC] QPT-7 Establish basic test structure

KenZSyd:
1. Adding functional tests in a seperate package
2. Enable standalone-selenium
3. Enable Typescript
4. Enable logging

@aaron-humerickhouse @lorentzkim 

https://github.com/bigcommerce/big-design/pull/278 represents the other approach. However I'm not sure whether browserstack will work well with this approach or not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ packages/docs/.next
 packages/docs/out
 packages/docs/pages/Dev
 .idea
+packages/tests/artifacts
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "diff": "lerna diff",
     "lint": "lerna run lint --stream",
     "start": "lerna run start --stream --parallel --ignore @bigcommerce/examples",
-    "test": "lerna run test --stream"
+    "test": "lerna run test --stream",
+    "test:functional": "lerna run --scope tests test:functional"
   },
   "husky": {
     "hooks": {
--- a/packages/tests/README.md
+++ b/packages/tests/README.md
@@ -1,0 +1,18 @@
+# `tests`
+
+> TODO: description
+
+## Usage
+
+```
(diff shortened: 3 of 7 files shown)


deini:
We need to discuss this further with the big-design team.

aaron-humerickhouse:
The plan is to get some basics up and running for a couple of tests.  WDIO makes it really easy to add plugins and should be quick to configure.  We wanted to show the big-design team how this would work.  If everyone agrees that it's a worthy venture, then we'll continue forward.  Otherwise, we'll find something else.

We have not dismissed Happo.  Because Happo doesn't support Android, we wanted to investigate a framework that would first.

rtalvarez:
Sorry I may be lacking some context here, but what's the broader strategy? What are the specific use cases you are aiming to test with this approach?

Will this setup be also running with every commit via circleCI ?

aaron-humerickhouse:
The broader strategy is to include cross-browser functional and visual regression testing.  We would have to figure out how feasible it is to run in every commit.  We can find a way to run a subset of tests on every commit and run the entire suite overnight or something.

rtalvarez:
Overnight? How long are we expecting these test suites to run for? 2-4 hours?

aaron-humerickhouse:
Running Selenium across 5 browsers and at least 3 mobile devices could take a while.  This is something we'll have to consider in determining whether or not this is the right approach moving forward.  If we have an average of 4 tests for every component and there are 26 components we're looking at over 800 tests (cross-browser).  

These are tests that mimic user behavior and don't rely on manipulating javascript or the DOM to make assertions.  They simulate what users are doing which makes it a little different than some of the faster tools.  Things like Jest aren't exercising components as a user would.  It's a different type of validation.

PR #288, opened 2019-12-03
feat(component): Chp 5713 add list groups w/headers to Dropdown component

jordan-massingill:
## WHAT
Adds optional list groups and headers to dropdown component, updates tests.

## TESTING
Added tests for grouped dropdowns, tested with screen reader. 

See screenshot below

Screen Shot 2019-12-03 at 2 03 52 PM

### DIFF
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -1,4 +1,4 @@
-import React, { RefObject } from 'react';
+import React, { Fragment, RefObject } from 'react';
 import { Manager, Reference } from 'react-popper';
 import scrollIntoView from 'scroll-into-view-if-needed';
 
@@ -128,10 +128,10 @@ export class Dropdown<T extends any> extends React.PureComponent<DropdownProps<T
 
   private renderGroup(group: DropdownOptionGroup<T>, groupIndex: number) {
     return (
-      <>
+      <Fragment key={groupIndex}>
         <ListGroupHeader role="group">{group.label}</ListGroupHeader>
         {this.renderOptions(group.options, groupIndex)}
-      </>
+      </Fragment>
     );
   }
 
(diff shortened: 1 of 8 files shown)


packages/big-design/src/components/Dropdown/types.ts
```diff
@@ -6,7 +6,7 @@ export type DropdownOption<T> = DropdownItem<T> | DropdownLinkItem<T>;
 
 export interface DropdownProps<T> extends Omit<React.HTMLAttributes<HTMLUListElement>, 'children'> {
   maxHeight?: number;
-  options: Array<DropdownOption<T>>;
+  options: Array<DropdownOption<T> | DropdownOptionGroup<T>>;
```
deini: Should this be `Array<DropdownOptionGroup<T>>`? 🤔

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -104,6 +126,15 @@ export class Dropdown<T extends any> extends React.PureComponent<DropdownProps<T
     );
   }
 
+  private renderGroup(group: DropdownOptionGroup<T>, groupIndex: number) {
+    return (
+      <Fragment key={groupIndex}>
+        <ListGroupHeader role="group">{group.label}</ListGroupHeader>
```
deini: Can we move the `role` directly into `ListGroupHeader`?

packages/big-design/src/components/Dropdown/types.ts
jordan-massingill: Both the `DropdownOption<T>` and the `DropdownOptionGroup<T>` are inside the `Array<>` distinction. I could also separate them to read `Array<DropdownOption<T>> | Array<DropdownOptionGroup<T>>` - is that what you mean?

deini: Ahh I missed that, however, I think your second proposal makes more sense.

If we do `Array<DropdownOption<T> | DropdownOptionGroup<T>>` that means you could mix both DropdownOptionGroup and DropdownOptions in the same array. Since your code is checking if all are options or all are groups (which makes sense) then the type is a bit off.

If we do `Array<DropdownOption<T>> | Array<DropdownOptionGroup<T>>` we would ensure that you can only use one kind.

jordan-massingill: Oooh good call, thanks for pointing that out 🙌

jordan-massingill:
Pulled most recent updates from master, tests no longer pass. 👀 looking now for fixes.

PR #290, opened 2019-12-04
feat(component): add tab trap for modals and dialogs

### LINKED ISSUE #252: Trap tab key on modals
For accessibility, we need to trap the tab key for Modals/Dialogs.

Rolling your own functionality for this can be quite cumbersome, so we'll probably want to use a library for it.

chanceaclark:
## What
trap

Add Tab trapping to dialogs and modals. Closes #252.

`focus-trap` is about 2kb (gzip + min) so it shouldn't add too much on the bundle size of BigDesign: https://bundlephobia.com/result?p=focus-trap@5.1.0


jordan-massingill:
The code lgtm - easy to read and understand 👍

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -68,18 +71,33 @@ export class Modal extends React.PureComponent<ModalProps, ModalState> {
...
     }
 
-    this.props.isOpen
-      ? (document.body.style.overflowY = 'hidden')
-      : (document.body.style.overflowY = this.state.initialBodyOverflowY);
+    if (this.props.isOpen) {
+      document.body.style.overflowY = 'hidden';
+
+      if (this.focusTrap) {
+        this.focusTrap.activate();
```
deini: 🍹 `this.focusTrap?.activate();`

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -68,18 +71,33 @@ export class Modal extends React.PureComponent<ModalProps, ModalState> {
     }
 
     document.body.style.overflowY = initialBodyOverflowY;
+
+    this.returnFocus();
   }
 
   componentDidUpdate(prevProps: ModalProps) {
+    if (this.modalRef.current) {
```
deini: Offline review: cleanup `focusTrap` and add a `null` check for it here.

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -68,18 +71,33 @@ export class Modal extends React.PureComponent<ModalProps, ModalState> {
...
+      document.body.style.overflowY = 'hidden';
+
+      if (this.focusTrap) {
+        this.focusTrap.activate();
+      }
+    } else {
+      document.body.style.overflowY = this.state.initialBodyOverflowY;
+
+      if (this.focusTrap) {
+        this.focusTrap.deactivate();
```
deini: 🍹 same as above

packages/big-design/src/components/Modal/Modal.tsx
chanceaclark: Totally forgot we could use those now 🌶

PR #299, opened 2019-12-16
fix(component): makes label prop required for dropdown groups

jordan-massingill:


jordan-massingill:
*quick note* - discovered that, while `options` aren't set as required prop for dropdowns (and therefore have not been made required for dropdown option groups), the dropdown will not render without them. May want to make them required as well as either part of this ticket or as part of another.

chanceaclark:
> _quick note_ - discovered that, while `options` aren't set as required prop for dropdowns (and therefore have not been made required for dropdown option groups), the dropdown will not render without them. May want to make them required as well, either part of this ticket or as part of another.

Probably a good idea to add to this PR too.

jorgemoya:
I believe `options` are are set as `required` in https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Dropdown/types.ts#L9, unless I'm missing something.

jorgemoya:
I think I see what you mean now. Docs are incorrect. Would be cool if you can fix them and mark as required. 🙏

jordan-massingill:
> Probably a good idea to add to this PR too.

Done 👍

PR #289, opened 2019-12-03
QPT-11: Implement a Logger in Testing

aaron-humerickhouse:
Add logging for tests and actions to a file
Keeps logs in the console as well

Proof:
From file `1575414593343-Bigcommerce_page_should_have_the_right_title.log`:

```
[2019-12-03T17:09:53.348] [INFO] test - --- Starting test Bigcommerce page should have the right title ---
[2019-12-03T17:09:53.350] [INFO] test - Navigating to bigcommerce.com
[2019-12-03T17:09:57.229] [INFO] test - --- Ending test Bigcommerce page should have the right title ---
```

From console:

```
[0-0] --- Starting test Bigcommerce page should have the right title ---
[0-0] Navigating to bigcommerce.com
[0-0] --- Ending test Bigcommerce page should have the right title ---
```

PR #301, opened 2019-12-16
test(configs): qpt-20 Initial visual regression setup

aaron-humerickhouse:
Add visual regression using Spectre and Selenium
Clean up some typescripty things
Add tsLint
Chrome or Firefox selection via env variable BROWSER

## Proof
Screen Shot 2019-12-16 at 5 08 34 PM

Screen Shot 2019-12-16 at 5 09 26 PM

Screen Shot 2019-12-16 at 5 12 14 PM

## Testing
* Start Spectre - cloud-dev-vm `./bcloud start spectre`
* Start docs - big-design `yarn start`
* Run tests - big-design `yarn test:functional` or `BROWSER=chrome yarn test:functional`

Navigate to https://spectre.service.bcdev/


packages/tests/src/components/bd-element.ts
```diff
@@ -0,0 +1,9 @@
+export default class BigDesignElement {
+  selector: string
+  browser: WebdriverIO.BrowserObject
+
+  constructor(selector: string, browser: WebdriverIO.BrowserObject) {
```
rtalvarez: Typescript hint:

Doing `constructor(foo: string) { this.foo = foo }` is equivalent to `constructor(private readonly foo: string) {}`

So in your case you should be able to get away with doing just `constructor(public selector: string, public browser: WebdriverIO.BrowserObject)` and delete the `this.selector = selector` and `this.browser = browser` as well as the class declarations `selector: string` and `browser: WebdriverIO.BrowserObject`

If you want to futher play around with any things that TS offers I suggest the playground: http://www.typescriptlang.org/play/

You can see the compiled, human readable JS right there

packages/tests/src/components/bd-checkbox.ts
```diff
@@ -0,0 +1,16 @@
+import BigDesignElement from './bd-element';
+
+// TODO: Make more typescripty
+export default class BigDesignCheckbox extends BigDesignElement {
+  toggle = (): void => {
```
rtalvarez: I would suggest to not manually annotate function return types unless you have a  compelling reason for it. Just let the compiler do its thing.

Also if you don't specifically need a bound function (arrow syntax) you can just do 

```
toggle() {
... 
}
```

packages/tests/src/components/bd-checkbox.ts
```diff
@@ -0,0 +1,16 @@
...
+    this.getBoxElement().click()
+    // tslint:disable-next-line:no-console
+    console.log(`Toggled checkbox with selector: ${this.selector}`)
+  }
+
+  private getBoxElement = (): WebdriverIO.Element => {
```
rtalvarez: same here, `private getBoxElement() { ... }` should be enough from what I can tell

packages/tests/src/helpers/assertions.ts
```diff
@@ -0,0 +1,22 @@
+import ScreenshotService from './../services/ScreenshotService'
+
+// TODO - .env
+const SPECTRE_UI_URL='https://spectre.service.bcdev'
```
rtalvarez: Is your lint working? I would expect this to fail because it expects whitespaces before and after the `=`, also a missing `;`

packages/tests/src/helpers/assertions.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// TODO - Add as browser command
+export async function assertElementLooks(browser:WebdriverIO.BrowserObject, selector:string, name:string): Promise<void> {
```
rtalvarez: Based on your type here, are you supposed to be returning something from this function?

packages/tests/src/services/ScreenshotService.ts
```diff
@@ -0,0 +1,72 @@
...
+
+// TODO - .env
+const SPECTRE_API_HOST= 'http://spectre.linkerd:4140';
+const PROJECT = 'Big Design';
+const SUITE = 'Components';
+
+export default class ScreenshotService {
+  runId: number
+  browser: WebdriverIO.BrowserObject
+  spectreClient: SpectreClient
```
rtalvarez: I believe can initialize and declare this in one line `spectreClient = new SpectreClient(SPECTRE_API_HOST)` without the need for a constructor

packages/tests/src/services/ScreenshotService.ts
```diff
@@ -0,0 +1,72 @@
...
+  private instantiateRun = async (): Promise<number> => {
+      const response = await this.spectreClient.createTestrun(PROJECT, SUITE)
+
+      return response.id
+  }
+
+  private takeImage = (selector:string, name:string): ScreenshotOutput => {
+    this.browser.$(selector).waitForDisplayed();
+
+    // TODO - Make sure TS definitions are added to browser
```
rtalvarez: Curious, what's the issue here? we should be aiming to avoid the use of ts-ignore

packages/tests/src/test/basic.ts
```diff
@@ -0,0 +1,14 @@
+import { assertElementLooks } from '../helpers/assertions'
+
+describe('Bigcommerce page', () => {
+    it('should have the right title', async () => {
+        // tslint:disable-next-line:no-console
```
rtalvarez: It looks like we are going to be using console.log a lot in these tests, I'd suggest configuring the linter to disable this specific rule instead of disabling every line where its used

packages/tests/src/components/bd-checkbox.ts
aaron-humerickhouse: I thought in the other PR you were asking to use the bound function syntax... https://github.com/bigcommerce/big-design/pull/289#discussion_r355610086

aaron-humerickhouse: or is it the difference between an instance function and a class function?

packages/tests/src/helpers/assertions.ts
aaron-humerickhouse: It did come up with several errors, currently it's clean.

packages/tests/src/components/bd-checkbox.ts
rtalvarez: Yep sorry must've been more clear. That is precisely the difference here

What I meant in the previous comment is doing `foo(() => { ... })` instead of `foo(function() { ... })`, but for class instance methods `foo() { .. }` is the standard way

packages/tests/src/components/bd-checkbox.ts
aaron-humerickhouse: ✅

packages/tests/src/components/bd-checkbox.ts
aaron-humerickhouse: ✅

packages/tests/src/helpers/assertions.ts
aaron-humerickhouse: Fixed the linting issue, but don't know why the linter didn't pick it up.

packages/tests/src/helpers/assertions.ts
aaron-humerickhouse: ✅ Removed the return type because it was a void promise based on advice in this PR

packages/tests/src/services/ScreenshotService.ts
aaron-humerickhouse: browser doesn't have a ts definition for saveElement.  It'll need some investment to make it happen.  I figured we can clean that up later if this is the route taken.

packages/tests/src/test/basic.ts
aaron-humerickhouse: ✅

packages/tests/src/components/bd-element.ts
aaron-humerickhouse: ✅

packages/tests/src/services/ScreenshotService.ts
rtalvarez: I see. Then may I suggest `return (this.browser as any).saveElement($(selector), name, { /* some options*/ })`

This basically casts `this.browser`, just for this line, to be the type `any`, which is the most generic type in TS. I think then you should be able to remove the ts-ignore. It's still not ideal but a bit better than disabling TS altogether IMO

Another suggestion I would have is, there is also a way to "augment" the type definition and merge in whatever else you need. Might be worthwhile if you'll be using `saveElement` a lot

aaron-humerickhouse: There is a way to add commands to browser.  I just didn't want to dig into it for a proof of concept.  I'll make the change over to `this.browser as any` and make sure the TODO is still in place

aaron-humerickhouse: ✅

packages/tests/src/services/ScreenshotService.ts
aaron-humerickhouse: ✅

aaron-humerickhouse:
@rtalvarez I don't see why my code changes have affected linting in packages/big-design.  Look at the circle build for further information

chanceaclark:
@aaron-humerickhouse something I just noticed that you are using `npm install` to install dependencies. Before the `wdio` branch gets merged into master, we'll need to remove the `package-lock.json` and use yarn to install dependencies instead. Just a heads up.

PR #303, opened 2019-12-23
feat(component): convert Dropdown/Select to FC and add MultiSelect

jorgemoya:
# ** BREAKING CHANGES**

## Changes

1. Refactor Dropdown and Select to use Downshift. The goal is to leverage this library's accessibility logic. Prop breaking changes.
2. Split MultiSelect to have its own component.
3. Updated the way we handle Input and Chips for a more streamlined logic.


packages/docs/PropTables/SelectPropTable.tsx
```diff
@@ -114,7 +119,7 @@ const selectProps: Prop[] = [
 
   {
     name: 'onItemChange',
-    types: '(value: string | number | Array<string | number>) => void',
+    types: '(value: any | Array<any>, option: Option | Array<Option>) => void',
```
chanceaclark: `SelectOptionPropTable` needs the `title` changed to `Select[Item]` I believe.

(Left the comment here, since it would only let me add comments in changed code sections.)

packages/big-design/src/components/Select/types.ts
```diff
@@ -18,10 +19,10 @@ export interface SelectProps<T> extends Omit<React.HTMLAttributes<HTMLUListEleme
...
   required?: boolean;
   value?: T | T[];
-  onItemChange(value: T | T[], option: Option<T> | Array<Option<T>>): void;
+  onItemChange(value: (T | undefined) | T[], option: (Option<T> | undefined) | Array<Option<T>>): void;
 }
 
-interface BaseItem extends Omit<ListItemProps, 'children' | 'content' | 'value'> {
+interface BaseItem extends Omit<ListItemProps, 'children' | 'content' | 'isHighlighted' | 'isSelected' | 'value'> {
   content: string;
   icon?: React.ReactElement;
```
chanceaclark: Something discussed in Slack: The icon prop doesn't actually render an icon in `Selects`. Is this something you can fix in this PR?

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,200 +1,183 @@
...
-        }
-
-        const id = this.getItemId(option, index, groupIndex);
-        const isHighlighted = Boolean(highlightedItem && id === highlightedItem.id);
-        const ref = React.createRef<HTMLLIElement>();
-
-        if (!option.disabled) {
-          this.listItemsRefs.push(ref);
-        }
+      <Fragment key={groupIndex}>
```
deini: `key` can't be specified here, it has to be on the `map`.

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,200 +1,183 @@
...
+  maxHeight = 250,
+  id,
+  items,
+  placement = 'bottom-start' as 'bottom-start',
+  trigger,
+  style,
+  ...rest
+}: DropdownProps<T>): ReturnType<React.FC<DropdownProps<T>>> {
+  // We need to keep track of key since we need to use it between groups.
+  let key = -1;
```
deini: Let's chat offline about how this `key` is being used.

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -206,168 +189,39 @@ export class Dropdown<T extends any> extends React.PureComponent<DropdownProps<T
...
-
-    return scrollIntoView(element, {
-      behavior: instantScroll ? 'instant' : 'smooth',
-      block: 'nearest',
-      inline: 'nearest',
-      scrollMode: 'if-needed',
-    });
-  };
+  return (
+    <div>
```
deini: Do we need this `div`?

packages/big-design/src/components/Input/Input.tsx
```diff
@@ -147,19 +147,9 @@ class StyleableInput extends React.PureComponent<InputProps & PrivateProps, Inpu
...
-    }
-
-    return chips.map((chip, key) =>
-      onChipDelete ? (
-        <Chip key={key} label={chip} marginBottom="none" onDelete={onChipDelete(chip)} />
-      ) : (
-        <Chip key={key} label={chip} marginBottom="none" marginTop="none" />
-      ),
-    );
+    return chips && chips.map((chip, key) => <Chip key={key} marginBottom="none" {...chip} />);
```
deini: Can we use something like `chip.label` as the `key`? In theory there shouldn't be any duplicate labels.

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -1,715 +1,427 @@
...
-              onChange={this.handleOnInputChange}
-              onChipDelete={chips && onChipDelete}
-              onClick={this.handleOnInputSelected}
-              onFocus={this.handleOnInputSelected}
-              onKeyDown={this.handleOnInputKeyDown}
-              placeholder={placeholder}
+              {...rest}
+              {...getInputProps({
+                disabled,
+                onClick: () => {
```
deini:
```
onClick: openMenu,
onFocus: openMenu,
```

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -1,715 +1,427 @@
...
+                  openMenu();
+                },
+                onFocus: () => {
+                  openMenu();
+                },
+                onKeyDown: (event: any) => {
```
deini: Can we type the event :pray:?

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -1,715 +1,427 @@
...
+                  if (multi && event.key === 'Backspace' && !inputValue) {
```
deini: Should we be checking for `inputValue.length`? Could there be a corner case with `0`?

packages/big-design/src/components/Select/styled.tsx
```diff
@@ -20,8 +22,28 @@ export const StyledDropdownIcon = styled(ArrowDropDownIcon)<{ disabled?: boolean
...
+    `}
+`;
+
 export const StyledInputContainer = styled.div`
   input[readonly] {
     cursor: pointer;
   }
 `;
+
+export const DropdownButton = styled(StyledButton)`
```
deini: All these require a `defaultTheme`

packages/docs/PropTables/SelectPropTable.tsx
```diff
@@ -103,7 +108,7 @@ const selectProps: Prop[] = [
 
   {
     name: 'value',
-    types: 'string | number | Array<string | number>',
+    types: 'any| Array<any>',
```
deini: whitespace

packages/big-design/src/components/List/List.tsx
```diff
@@ -1,56 +1,25 @@
...
-          </ListPopperElement>
-        </StyledList>
-      )}
-    </Popper>
-  ),
-);
+const StyleableList: React.FC<ListProps & PrivateProps> = ({ forwardedRef, scheduleUpdate, ...props }) => {
+  useEffect(() => {
+    scheduleUpdate();
+  }, [React.Children.count(props.children), props.isOpen]);
```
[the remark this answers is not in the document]
jorgemoya: Wouldn't changing a `prop` on a `children` trigger the `useEffect`? I was thinking this would minimize the `useEffect` to only when the number of children changes, doesn't care about the props.

packages/big-design/src/components/Chip/types.ts
```diff
@@ -0,0 +1,8 @@
+import { MarginProps } from '../../mixins/margins/margins';
+
+export interface ChipProps extends MarginProps {
+  label: string;
+  onDelete?(): void;
+}
+
+export type ChipObject = ChipProps;
```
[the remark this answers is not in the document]
jorgemoya: It's being used in `Input`. 

```
interface Props {
  ...
  chips?: ChipObject[];
}
```

I was trying to make it more explicit but can easily change it to `chips?: ChipProps[]`. Not sure if there's a better approach?

packages/docs/PropTables/SelectPropTable.tsx
jorgemoya: `Select` still uses `options` and `action` as items. Maybe I should change the prop `onItemChange` to `onOptionChange`?

packages/big-design/src/components/Dropdown/spec.tsx
```diff
@@ -164,114 +161,92 @@ test('esc should close menu', () => {
   expect(queryByRole('listbox')).toHaveStyle('height: 1px');
 });
 
-test('tab should close menu', () => {
-  const { getByRole, queryByRole } = render(DropdownMock(onClick));
-  const trigger = getByRole('button');
+// test('tab should close menu', () => {
```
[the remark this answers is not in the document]
jorgemoya: Yes, I mentioned this test in my original PR submit. Can't get it to work, need to 👀 more.

packages/docs/PropTables/SelectPropTable.tsx
chanceaclark: I know we've had some offline conversations about `items` vs. `options`. Maybe we need to have another one so that these are consistent?

packages/big-design/src/components/Chip/types.ts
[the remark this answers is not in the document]
chanceaclark: We should be consistent all across the library, which I think we do a similar thing with modal actions. The only reason why we gave a new type for `ModalActions[]` was so we could omit `children` from the prop list.

packages/big-design/src/components/List/List.tsx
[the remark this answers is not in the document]
chanceaclark: There might be some instances where we do care about running the function when a child prop changes. It may be that the text of some item expands and drops onto a new line, then we would need to run the `scheduleUpdate` function.

packages/big-design/src/components/Select/Select.tsx
jorgemoya: Hm, I fail to see why it needs to change, tbh. `inputValue` is a string, and will only return falsey if it's `""` right?

packages/big-design/src/components/Input/Input.tsx
jorgemoya: Sure.

packages/big-design/src/components/Dropdown/Dropdown.tsx
jorgemoya: `groupIndex` comes from the map.

deini: https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,200 +1,183 @@
...
+      if (!selectedItem) {
+        return;
+      }
 
-  render() {
-    const { children, maxHeight, options, placement, trigger, ...rest } = this.props;
-    const { highlightedItem, isOpen } = this.state;
+      // send to url if selectedItem is of type 'link'
+      if (selectedItem.type === 'link' && selectedItem.url) {
+        window.location.href = selectedItem.url;
```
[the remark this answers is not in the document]
jorgemoya: This is trying to fix the behavior of pressing enter on a `LinkItem`. The use case you mention wouldn't work, so let me see how to fix this.

packages/big-design/src/components/Select/types.ts
jorgemoya: Will take a 👀

packages/docs/pages/MultiSelect/MultiSelectPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+          ]}
+          placeholder="Choose from the left"
+          placement="left-end"
+          required
+        />
+      </Grid>
+      {/* jsx-to-string:end */}
+    </CodePreview>
+
+    <H1>Max Height</H1>
```
chanceaclark: Example titles should be `H2`'s. See `PAGES.md`.
```suggestion
    <H1>Examples</H1>
    
    <H2>Max Height</H2>
```

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,373 +1,235 @@
...
-    return triggerId || this.uniqueTriggerId;
-  }
+      if (Array.isArray(items) && items.every(isItem)) {
+        return renderItems(items as Array<DropdownItem<T>>);
+      }
 
-  private updateHighlightedItem(element: HTMLLIElement | null, scroll?: boolean, instantScroll?: boolean) {
-    if (!element) {
       return;
+    }, [highlightedIndex, items]);
```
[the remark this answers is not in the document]
jorgemoya: `renderGroup` and `renderItems` are called inside this function, so I fail to see why they're a dependency.

packages/big-design/src/components/Dropdown/spec.tsx
```diff
@@ -96,61 +108,46 @@ test('opens/closes dropdown menu when trigger is clicked', () => {
...
+  fireEvent.click(trigger);
 
-  fireEvent.keyDown(menu, { key: 'ArrowUp' });
-  expect(options[0].dataset.highlighted).toBe('true');
+  fireEvent.keyDown(menu, { key: 'ArrowDown' });
+  expect(options[1].getAttribute('aria-selected')).toBe('true');
 
   fireEvent.keyDown(menu, { key: 'ArrowUp' });
-  expect(options[3].dataset.highlighted).toBe('true');
+  expect(options[0].getAttribute('aria-selected')).toBe('true');
```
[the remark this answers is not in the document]
jorgemoya: I'm gonna add the circle functionality instead and make sure it still circles back on the tests. I somehow missed initially that this was an option for the hook.

packages/big-design/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
chanceaclark: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,407 @@
...
+                icon={item.icon}
+                isHighlighted={isHighlighted}
+              >
+                {getContent(item, isHighlighted)}
+              </ListItem>
+            </Box>
+          )
+        );
+      });
+    }, [memofiedAction, highlightedIndex, items, selectedOptions]);
```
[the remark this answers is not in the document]
jorgemoya: Can't seem to find what else is missing.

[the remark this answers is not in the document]
jorgemoya: Never mind, I get it now.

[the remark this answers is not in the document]
chanceaclark: `getContent`, `getItemProps`, `removeItem`, `addSelectedItem`

PR #307, opened 2020-01-02
feat(component): add text-transform props to Text and Small

deini:
New text modifiers:

- uppercase
- lowercase
- capitalize

Screen Shot 2020-01-02 at 12 35 17 PM

### DIFF
--- a/packages/docs/pages/Typography/TypographyPage.tsx
+++ b/packages/docs/pages/Typography/TypographyPage.tsx
@@ -78,7 +78,13 @@ export default () => (
         <Text bold>This text is bold.</Text>
         <Text italic>This text is italic.</Text>
         <Text strikethrough>This text is strikethrough.</Text>
-        <Text underline>This color is underline.</Text>
+        <Text underline>This text is underlined.</Text>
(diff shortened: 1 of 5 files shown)


packages/docs/pages/Typography/TypographyPage.tsx
```diff
@@ -79,6 +79,12 @@ export default () => (
         <Text italic>This text is italic.</Text>
         <Text strikethrough>This text is strikethrough.</Text>
         <Text underline>This color is underline.</Text>
+        <Text uppercase>This text is uppercase.</Text>
+        <Text lowercase>This text is lowercase.</Text>
+        <Text capitalize>This text is capitalize.</Text>
```
chanceaclark:
```suggestion
        <Text capitalize>This text is capitalized.</Text>
```

deini: Should we change all the other ones (underline, etc) 🤔? I was just using the css value.

chanceaclark: No I was just thinking about proper grammar on this one.

PR #312, opened 2020-01-03
feat(icons): forwardRef to svg

deini:
- Forward ref to `svg` in all Icons
- Add `titleId` prop to icons which is used as a title id and `aria-labelledb` of the svg.
- Add `useUniqueId` hook.

### DIFF
--- a/packages/big-design/tests/utils.tsx
+++ b/packages/big-design/tests/utils.tsx
@@ -1,0 +1,16 @@
+import { render, RenderOptions } from '@testing-library/react';
+import React from 'react';
+import { UIDReset } from 'react-uid';
+
+const BigDesignWrapper: React.FC = ({ children }) => {
+  return <UIDReset>{children}</UIDReset>;
+};
+
+const customRender = (ui: React.ReactElement<any>, options: RenderOptions = {}) =>
+  render(ui, { wrapper: BigDesignWrapper, ...options });
+
+// re-export everything
+export * from '@testing-library/react';
+
+// override render method
+export { customRender as render };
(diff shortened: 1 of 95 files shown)


packages/docs/pages/_app.tsx
```diff
@@ -46,38 +47,42 @@ export default class MyApp extends App {
             }
           `}
         </style>
-        <ThemeProvider theme={theme}>
-          <>
-            <GlobalStyles />
-            {router.query.noNav ? (
-              <Component {...pageProps} />
-            ) : (
+        <UIDReset>
```
chanceaclark: Are we going to have to add these to our microapps now? 🤔Should we create a HOC for everything that's essential:
```tsx
const Wrapper = ({ children, theme }) => (
  <UIDReset>
    <UIDFork>
      <ThemeProvider theme={theme}>
        <GlobalStyles />
        {children}
      </ThemeProvider>
    </UIDFork>
  </UIDReset>
);
```
Just a 💭

deini: We don't need to add them to our microapps. These are only to provide better SSR support. We can create a wrapper in the future when we need it.

packages/big-design/tests/utils.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const BigDesignWrapper: React.FC = ({ children }) => {
+  return <UIDReset>{children}</UIDReset>;
+};
+
+const customRender = (ui: React.ReactElement<any>, options: RenderOptions = {}) =>
+  render(ui, { wrapper: BigDesignWrapper, ...options });
+
+// re-export everything
+export * from '@testing-library/react';
```
rtalvarez: cool trick (🍹 although not sure the comment above adds anything, the code itself is pretty clear)

PR #316, opened 2020-01-07
feat(component): convert ProgressCircle to FC

chanceaclark:
## What
- Refactor ProgressCircle to a functional component.

_Untitled Pull Request_

### DIFF
--- a/packages/big-design/src/components/ProgressCircle/ProgressCircle.tsx
+++ b/packages/big-design/src/components/ProgressCircle/ProgressCircle.tsx
@@ -1,5 +1,7 @@
 import { CheckCircleIcon, ErrorIcon } from '@bigcommerce/big-design-icons';
-import React from 'react';
+import React, { useMemo } from 'react';
+
+import { typedMemo } from '../../utils';
 
 import { CIRCLE_DIMENSIONS } from './constants';
 import { StyledCircle, StyledCircleFiller, StyledProgressCircle, StyledText } from './styled';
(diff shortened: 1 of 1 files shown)


packages/big-design/src/components/ProgressCircle/ProgressCircle.tsx
```diff
@@ -71,21 +55,23 @@ export class ProgressCircle extends React.PureComponent<ProgressCircleProps> {
...
-        size={this.getDimensions()}
+        size={getDimensions()}
       />
     );
-  }
+  };
 
-  renderSuccess() {
+  const renderSuccess = () => {
     return (
```
golcinho: 🍸 I think you could return `CheckCircleIcon ` immediately. `const renderSuccess = () => (...);`

packages/big-design/src/components/ProgressCircle/ProgressCircle.tsx
```diff
@@ -10,26 +12,10 @@ export interface ProgressCircleProps {
...
-  getDimensions() {
-    const { size } = this.props;
-
-    return size ? CIRCLE_DIMENSIONS[size] : CIRCLE_DIMENSIONS.medium;
-  }
-
-  renderCircle() {
-    const { percent, size } = this.props;
+export const ProgressCircle: React.FC<ProgressCircleProps> = typedMemo(({ error, percent, size = 'medium' }) => {
+  const getDimensions = () => (size ? CIRCLE_DIMENSIONS[size] : CIRCLE_DIMENSIONS.medium);
```
golcinho: I think it's a redundant condition because by default `size` equal to `medium`. So it could just `CIRCLE_DIMENSIONS[size]`

packages/big-design/src/components/ProgressCircle/ProgressCircle.tsx
```diff
@@ -58,11 +44,9 @@ export class ProgressCircle extends React.PureComponent<ProgressCircleProps> {
         )}
       </StyledProgressCircle>
     );
-  }
-
-  renderError() {
-    const { percent } = this.props;
+  };
 
+  const renderError = () => {
```
golcinho: 🍸 I think you could return ErrorIcon immediately. `const renderError = () => (...);`

packages/big-design/src/components/ProgressCircle/ProgressCircle.tsx
golcinho: Could it be `null`? if not you could use just `CIRCLE_DIMENSIONS[size]` because y default size is `medium`

deini: We could probably get rid of `getDimensinos` and just do 
```ts
const dimensions = CIRCLE_DIMENSIONS[size];
``` 

Since it will re-render if `size` changes we don't need to call a fn anymore.

PR #317, opened 2020-01-08
feat(component): convert Modal to FC

chanceaclark:
## What
Convert `Modal` to functional component.

_Till Hacktember Ends_

### DIFF
--- a/packages/big-design/src/components/Modal/Modal.tsx
+++ b/packages/big-design/src/components/Modal/Modal.tsx
@@ -1,9 +1,9 @@
 import { CloseIcon } from '@bigcommerce/big-design-icons';
 import focusTrap, { FocusTrap } from 'focus-trap';
-import React from 'react';
+import React, { createRef, useEffect, useMemo, useState } from 'react';
 import { createPortal } from 'react-dom';
 
-import { uniqueId } from '../../utils';
+import { typedMemo, uniqueId } from '../../utils';
 import { Button, ButtonProps } from '../Button';
 import { H2 } from '../Typography';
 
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -18,170 +18,146 @@ import {
...
+    useEffect(() => {
+      return () => {
+        if (modalContainer) {
+          document.body.removeChild(modalContainer);
+        }
+
+        document.body.style.overflowY = initialBodyOverflowY;
+        returnFocus();
+      };
+    }, [modalContainer]);
```
deini: Should this have `returnFocus` as a dep too?

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -18,170 +18,146 @@ import {
...
+        setInitialBodyOverflowY(document.body.style.overflowY || '');
+        document.body.style.overflowY = 'hidden';
+        internalTrap?.activate();
+      } else {
+        setInitialBodyOverflowY('');
+        document.body.style.overflowY = initialBodyOverflowY;
+        internalTrap?.deactivate();
+        setInternalTrap(null);
+      }
+    }, [internalTrap, isOpen, modalRef]);
```
deini: There are some missing deps here too. Is that intentional?

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -18,170 +18,146 @@ import {
...
+
+    useEffect(() => {
+      const container = document.createElement('div');
+
+      document.body.appendChild(container);
+      setModalContainer(container);
+    }, []);
+
+    useEffect(() => {
+      return () => {
```
jordan-massingill: I'm not fully clear on why a function is returned from this useEffect() - would you mind explaining?

packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -18,170 +18,146 @@ import {
...
-    }
-  };
-
-  private onKeyDown = (event: React.KeyboardEvent) => {
-    if (this.props.closeOnEscKey && event.key === 'Escape') {
-      this.props.onClose();
-    }
-  };
-}
+    return !(isOpen && modalContainer) ? null : createPortal(renderedContent, modalContainer);
```
jordan-massingill: What is the difference between this and `return isOpen && modalContainer ? createPortal(renderedContent, modalContainer) : null` ?

packages/big-design/src/components/Modal/Modal.tsx
chanceaclark: Returning a function on a `useEffect` is roughly the same as `componentWillUnmount`. A common use-case for using this cleanup functionality is removing event listeners on the DOM:
```tsx
useEffect(() => {
  window.addEventListener('resize', handleResize);

  return () => {
    window.removeEventListener('resize', handleResize);
  };
});
```

This is extremely useful when components get added/removed from the DOM frequently.

Here's some more information on the effect cleanup: https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect

---

In our use case we want the users focus to return back to their last focusable element. We use portals to add the modal onto the page so it'll mount/unmount every time it open/closes. Since we want to return focus, we run it on the effect cleanup.

packages/big-design/src/components/Modal/Modal.tsx
chanceaclark: Good catch, I can change it over.

packages/big-design/src/components/Modal/Modal.tsx
jordan-massingill: Oh thanks! I remember that now that you mention it. It's been a while since I've used the useEffect hook, will brush up on it again for my current work.

PR #319, opened 2020-01-09
feat(component): covert Fieldset to FC and remove static members

chanceaclark:
## What
- Export `Fieldset` instead of it being a static member.
- Refactor `Fieldset` component to be a functional component.
- Export `FieldsetLegend` and `FieldsetDescription`.

## warning.ts
We know throw an error in console during runtime type checking for the `Fieldset` `legend` and `descirption` props. This ensures the consumer of the component is using the right components when needing to use custom `legend` and `descriptions`'s. The implementation of `warning.ts` is a slightly modified version of [`tiny-warning` module](https://github.com/alexreardon/tiny-warning/blob/master/src/index.js).

---

BREAKING CHANGE:
`Form.Fieldset` renamed to `Fieldset` and will now have to
`import { Fieldset } from '@bigcommerce/big-design';`


deini:
Offline review 🕺

PR #320, opened 2020-01-13
feat(component): convert Input to FC and remove static members

chanceaclark:
## What
Convert `Input` to functional component and remove static members.

---

BREAKING CHANGE:
Use `FormControlDescription`, `FormControlError`, and `FormControlLabel`
instead of `Input.Description`, `Input.Error`, and `Input.Label` respectively.

### DIFF
--- a/packages/big-design/src/components/Input/Input.tsx
+++ b/packages/big-design/src/components/Input/Input.tsx
@@ -1,7 +1,7 @@
-import hoistNonReactStatics from 'hoist-non-react-statics';
-import React, { Ref } from 'react';
+import React, { forwardRef, useMemo, useState, Ref } from 'react';
 
-import { uniqueId } from '../../utils';
+import { typedMemo, uniqueId } from '../../utils';
+import { warning } from '../../utils/warning';
 import { Chip } from '../Chip';
 import { FormControlDescription, FormControlError, FormControlLabel } from '../Form';
 
@@ -10,7 +10,7 @@ import { StyledIconWrapper, StyledInput, StyledInputContent, StyledInputWrapper 
 interface Props {
   chips?: string[];
   description?: React.ReactChild;
-  error?: React.ReactChild | React.ReactChild[];
+  error?: React.ReactNode | React.ReactNode[];
   iconLeft?: React.ReactChild;
   iconRight?: React.ReactChild;
   label?: React.ReactChild;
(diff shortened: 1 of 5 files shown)


packages/big-design/src/components/Input/Input.tsx
```diff
@@ -161,13 +121,62 @@ class StyleableInput extends React.PureComponent<InputProps & PrivateProps, Inpu
...
       ),
     );
-  }
-}
+  }, [chips, onChipDelete]);
 
-const InputWithForwardedRef = React.forwardRef<HTMLInputElement, InputProps>(({ className, style, ...props }, ref) => (
-  <StyleableInput {...props} forwardedRef={ref} />
-));
+  const isErrorValid = (err: Props['error']) => {
```
deini: Can we rename this function? I would expect this to return a `boolean` by its name.

packages/big-design/src/components/Input/Input.tsx
```diff
@@ -22,134 +22,94 @@ interface PrivateProps {
...
+    const { onFocus } = props;
+
+    setFocus(true);
 
     return onFocus && onFocus(event);
   };
 
-  private onInputBlur = (event: React.FocusEvent<HTMLInputElement>) => {
-    const { onBlur } = this.props;
+  const onInputBlur = (event: React.FocusEvent<HTMLInputElement>) => {
```
golcinho: 🍸 What do you think if we rename this event handler method to `handleInputBlur` or just `handleBlur`?

packages/big-design/src/components/Input/Input.tsx
```diff
@@ -22,134 +22,94 @@ interface PrivateProps {
...
+  forwardedRef,
+  label,
+  labelId,
+  onChipDelete,
+  ...props
+}) => {
+  const [focus, setFocus] = useState(false);
+  const id = useMemo(() => (props.id ? props.id : uniqueId('input_')), [props.id]);
+
+  const onInputFocus = (event: React.FocusEvent<HTMLInputElement>) => {
```
golcinho: 🍸 What do you think if we rename this event handler method to`handleInputFocus` or just `handleFocus`?

packages/big-design/src/components/Input/Input.tsx
```diff
@@ -161,13 +121,62 @@ class StyleableInput extends React.PureComponent<InputProps & PrivateProps, Inpu
...
+
+    if (!err) {
+      return null;
+    }
+
+    warning('error must be either a string or a FormControlError component.');
+  };
 
-export const Input = hoistNonReactStatics(InputWithForwardedRef, StyleableInput);
+  const errors = useMemo(() => {
```
golcinho: 🍸 Could the name be `getErrors`? What do you think?

PR #321, opened 2020-01-13
feat(component): convert Textarea to FC and remove static members

chanceaclark:
## What
Convert `Input` to functional component and remove static members.

---

BREAKING CHANGE:
Use `FormControlDescription`, `FormControlError`, and `FormControlLabel`
instead of `Textarea.Description`, `Textarea.Error`, and `Textarea.Label` respectively.

### DIFF
--- a/packages/big-design/src/components/Textarea/Textarea.tsx
+++ b/packages/big-design/src/components/Textarea/Textarea.tsx
@@ -1,14 +1,14 @@
-import hoistNonReactStatics from 'hoist-non-react-statics';
-import React, { Ref } from 'react';
+import React, { forwardRef, useMemo, Ref } from 'react';
 
-import { uniqueId } from '../../utils';
+import { typedMemo, warning } from '../../utils';
+import { useUniqueId } from '../../utils/useUniqueId';
 import { FormControlDescription, FormControlError, FormControlLabel } from '../Form';
 
 import { StyledTextarea, StyledTextareaWrapper } from './styled';
 
 interface Props {
   description?: React.ReactChild;
-  error?: React.ReactChild;
+  error?: React.ReactNode | React.ReactNode[];
   label?: React.ReactChild;
   labelId?: string;
   rows?: 1 | 2 | 3 | 4 | 5 | 6 | 7;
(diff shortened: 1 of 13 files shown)


packages/big-design/src/components/Textarea/Textarea.tsx
```diff
@@ -21,88 +21,105 @@ interface PrivateProps {
...
+    warning('description must be either a string or a FormControlDescription component.');
+  }, [description]);
 
-    if (typeof label === 'string') {
-      return (
-        <Textarea.Label id={labelId} htmlFor={id} renderOptional={!required}>
-          {label}
-        </Textarea.Label>
-      );
+  const numOfRows = useMemo(() => {
```
deini: Let's not do a `useMemo` here, the returned value is a primitive and wont cause a re-render.

packages/big-design/src/components/Textarea/Textarea.tsx
```diff
@@ -21,88 +21,105 @@ interface PrivateProps {
...
+    if (Array.isArray(error)) {
+      error.forEach(validateError);
 
-export const Textarea = hoistNonReactStatics(TextareaWithForwardedRef, StyleableTextarea);
+      return error;
+    }
 
-Textarea.displayName = 'Textarea';
+    return validateError(error);
+  }, [error]);
```
deini: Missing `validateError` dependency, consider moving it inside this memo if not being used elsewhere.

packages/big-design/src/components/Textarea/Textarea.tsx
```diff
@@ -21,88 +21,105 @@ interface PrivateProps {
...
+    if (React.isValidElement(label) && label.type === FormControlLabel) {
+      return React.cloneElement(label as React.ReactElement<React.LabelHTMLAttributes<HTMLLabelElement>>, {
+        id: labelId,
+        htmlFor: id,
+      });
+    }
 
-    return id ? id : this.uniqueId;
-  }
+    if (!label) {
```
deini: What do you think about movin this check to be the first one? Corner case: `label = ''`.

PR #322, opened 2020-01-13
feat(component): convert Checkbox to FC and remove static members

chanceaclark:
## What
Convert `Checkbox` to a functional component and remove static members.

---

BREAKING CHANGE:
Use `CheckboxLabel` instead of `Checkbox.Label`.

### DIFF
--- a/packages/big-design/src/components/Checkbox/Checkbox.tsx
+++ b/packages/big-design/src/components/Checkbox/Checkbox.tsx
@@ -1,10 +1,11 @@
 import { CheckIcon, RemoveIcon } from '@bigcommerce/big-design-icons';
-import hoistNonReactStatics from 'hoist-non-react-statics';
-import React, { memo, Ref } from 'react';
+import React, { forwardRef, useMemo, Ref } from 'react';
 
-import { uniqueId } from '../../utils';
+import { typedMemo, warning } from '../../utils';
+import { useUniqueId } from '../../utils/useUniqueId';
 
-import { CheckboxContainer, HiddenCheckbox, StyledCheckbox, StyledLabel } from './styled';
+import { CheckboxContainer, HiddenCheckbox, StyledCheckbox } from './styled';
+import { CheckboxLabel } from './Label';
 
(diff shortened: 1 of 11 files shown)


packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -18,102 +19,92 @@ interface PrivateProps {
...
+          }
+
+          if (forwardedRef) {
+            if (typeof forwardedRef === 'function') {
+              forwardedRef(checkbox);
+            } else {
+              // RefObject.current is readonly in DefinitelyTyped but in practice you
+              // can still write to it.
+              // See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/31065
+              // @ts-ignore
```
rtalvarez: 🍹 would doing `(forwardRef as any).current = checkbox`  be a bit less destructive?

chanceaclark: To be more explicit, I could do `(forwardedRef as React.MutableRefObject<HTMLInputElement | null>).current = checkbox;` instead. What do you think?

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -18,102 +19,92 @@ interface PrivateProps {
...
 
-export const Checkbox = hoistNonReactStatics(CheckboxWithForwardedRef, RawCheckbox);
-
-export const StyleableCheckbox = React.forwardRef<HTMLInputElement, CheckboxProps>((props, ref) => (
-  <RawCheckbox {...props} forwardedRef={ref} />
-));
-
-Checkbox.displayName = 'Checkbox';
-StyleableCheckbox.displayName = 'StyleableCheckbox';
+export const StyleableCheckbox = typedMemo(
```
deini: Since we aren't using `StyleableCheckbox` anywhere, maybe we should remove it and add it back if we need it. Thoughts?

PR #323, opened 2020-01-13
feat(component): convert Radio to FC and remove static members

chanceaclark:
## What
Convert `Radio` to a functional component and remove static members.

---

BREAKING CHANGE:
Use `RadioLabel` instead of `Radio.Label`.

### DIFF
--- a/packages/big-design/src/components/Radio/Radio.tsx
+++ b/packages/big-design/src/components/Radio/Radio.tsx
@@ -1,9 +1,10 @@
-import hoistNonReactStatics from 'hoist-non-react-statics';
-import React, { memo, Ref } from 'react';
+import React, { forwardRef, useMemo, Ref } from 'react';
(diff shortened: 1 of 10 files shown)


packages/big-design/src/components/Radio/Radio.tsx
```diff
@@ -15,70 +16,64 @@ interface PrivateProps {
...
+  );
+};
+
+export const Radio = typedMemo(
+  forwardRef<HTMLInputElement, RadioProps>(({ className, style, ...props }, ref) => (
+    <RawRadio {...props} forwardedRef={ref} />
+  )),
+);
+
+export const StyleableRadio = typedMemo(
```
deini: Same as Checkbox, not part of your changes but probably worth it to remove the `StyleableRadio` export.

PR #324, opened 2020-01-13
feat(component): convert Form to FC and remove static members

chanceaclark:
## What
Convert `Form` to a functional component and remove static members.

---

BREAKING CHANGE:
Use `FormControlError`, `FormControlLabel`, `FormGroup`, and `Fieldset`
instead of `Form.Error`, `Form.Label`, `Form.Group`, and `Form.Fieldset` respectively.


deini:
@golcinho Yes, we are and will continue to use `semver` after beta. (No breaking changes after `1.x.x` without a major version bump).

PR #326, opened 2020-01-14
feat(component): convert Tooltip to FC and remove static members

jordan-massingill:
### WHAT

Convert the Tooltip class component to a FC and remove its static members

### DIFF
--- a/packages/big-design/src/components/Tooltip/Tooltip.tsx
+++ b/packages/big-design/src/components/Tooltip/Tooltip.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { createPortal } from 'react-dom';
 import { Manager, Popper, PopperProps, Reference } from 'react-popper';
 
+import { typedMemo } from '../../utils';
 import { Small } from '../Typography';
 
 import { StyledTooltip, StyledTooltipTrigger } from './styled';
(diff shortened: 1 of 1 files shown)


packages/big-design/src/components/Tooltip/Tooltip.tsx
```diff
@@ -13,90 +14,80 @@ export interface TooltipProps extends React.HTMLAttributes<HTMLDivElement> {
...
+        if (tooltipContainer) {
+          document.body.removeChild(tooltipContainer);
+        }
+      };
+    }, [tooltipContainer]);
 
-  private tooltipContainer?: HTMLDivElement;
+    const renderContent = (content: React.ReactNode) => {
+      return typeof content === 'string' ? <Small color="white">{content}</Small> : content;
+    };
```
chanceaclark: 🤔 May want to refactor to something like this so the content doesn't cause any unnecessary rerenders.
```tsx
    const renderedContent = useMemo(() => {
      return typeof children === 'string' ? <Small color="white">{children}</Small> : children;
    }, [children]);

    ...further down...
                    <StyledTooltip ref={ref} style={style} data-placement={placement}>
                      {renderedContent}
                    </StyledTooltip>

```

packages/big-design/src/components/Tooltip/Tooltip.tsx
```diff
@@ -13,90 +14,80 @@ export interface TooltipProps extends React.HTMLAttributes<HTMLDivElement> {
...
+    useEffect(() => {
+      return () => {
+        if (tooltipContainer) {
+          document.body.removeChild(tooltipContainer);
+        }
+      };
+    }, [tooltipContainer]);
 
-  private tooltipContainer?: HTMLDivElement;
+    const renderContent = useMemo(() => {
```
deini: @jordan-massingill @chanceaclark I don't think this `useMemo` is needed here. `{renderContent}` is already  behind the `isVisible` check. When `isVisible = true` it has to re-render anyway.

packages/big-design/src/components/Tooltip/Tooltip.tsx
```diff
@@ -13,90 +14,80 @@ export interface TooltipProps extends React.HTMLAttributes<HTMLDivElement> {
...
         </Reference>
-        {this.tooltipContainer
+        {tooltipContainer
           ? createPortal(
               <Popper
-                placement={this.props.placement}
-                modifiers={{ offset: { offset: '0, 8' }, ...this.props.modifiers }}
-                eventsEnabled={this.state.visible}
+                placement={props.placement || 'top'}
+                modifiers={{ offset: { offset: '0, 8' }, ...modifiers }}
```
jorgemoya: I had an issue with this `offset` object. React wanted me to memoize it. If you're having no issues with this I need to look at my code haha.

packages/big-design/src/components/Tooltip/Tooltip.tsx
```diff
@@ -13,90 +14,80 @@ export interface TooltipProps extends React.HTMLAttributes<HTMLDivElement> {
...
-  state = {
-    visible: false,
-  };
+    useEffect(() => {
+      return () => {
+        if (tooltipContainer) {
+          document.body.removeChild(tooltipContainer);
+        }
+      };
+    }, [tooltipContainer]);
```
jorgemoya: I don't understand the purpose of this. You set the tooltipContainer up top but then remove it?

jordan-massingill: Yes - this useEffect just handles the unmount.

packages/big-design/src/components/Tooltip/Tooltip.tsx
jordan-massingill: This hasn't given me any issues *_so far_*.

packages/big-design/src/components/Tooltip/Tooltip.tsx
```diff
@@ -13,90 +13,78 @@ export interface TooltipProps extends React.HTMLAttributes<HTMLDivElement> {
...
+  useEffect(() => {
+    return () => {
+      if (tooltipContainer) {
+        document.body.removeChild(tooltipContainer);
+      }
+    };
+  }, [tooltipContainer]);
 
-  private renderContent = (content: React.ReactNode) => {
+  const renderContent = (content: React.ReactNode) => {
```
deini: 🍹 Since you have access to `children` here I would use it directly instead of passing it.

jordan-massingill: Yeah, I went back and forth on whether to pass it or use it directly. This implementation was only to cover the edge case that this function could be used elsewhere in future refactors (though unlikely). Will change it back. 👍

PR #327, opened 2020-01-15
fix(component): icon alignment and autofill input fixes

chanceaclark:
## What
Inputs with `Icons` are not perfectly aligned and have extra margins / paddings. This PR is an attempt to resolve that, alongside ensuring the whole field upon autocomplete is filled.

## Screenshot
Screen Shot 2020-01-15 at 2 41 20 PM

### DIFF
--- a/packages/big-design/src/components/Input/styled.tsx
+++ b/packages/big-design/src/components/Input/styled.tsx
@@ -1,6 +1,7 @@
-import { addValues, theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import { addValues, remCalc, theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import styled, { css } from 'styled-components';
 
+import { withPaddings, PaddingProps } from '../../mixins';
 import { withTransition } from '../../mixins/transitions';
 
 import { InputProps } from './Input';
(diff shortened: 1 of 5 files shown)


packages/big-design/src/components/Input/styled.tsx
```diff
@@ -68,47 +69,92 @@ export const StyledInput = styled.input<InputProps>`
     font-size: ${({ theme }) => theme.typography.fontSize.medium};
   }
 
-  ${props =>
-    props.iconRight &&
+  &:-webkit-autofill {
+    background-color: ${({ theme }) => theme.colors.primary10} !important;
```
deini: Can we add a comment on why we are using `!important`?

packages/big-design/src/components/Input/styled.tsx
```diff
@@ -68,47 +69,92 @@ export const StyledInput = styled.input<InputProps>`
...
+    -webkit-box-shadow: 0 0 0px 1000px ${({ theme }) => theme.colors.primary10} inset;
```
deini: This throws a warning on `ts-styled-plugin` 🤔 is there a way to ignore it?

chanceaclark: I'll 👀 to see if there's a way

chanceaclark: Unfortunately there isn't a way to disable it on a per-line basis:
https://github.com/Microsoft/typescript-styled-plugin/issues/70#issuecomment-420457682

We could disable linting validation entirely, but I assume we wouldn't want that.

PR #328, opened 2020-01-16
chore(component): add useRafState and useWindowSize hooks

deini:
Adds `useRafState` and `useWindowSize` hooks.

`useRafState`: Has the same signature as `useState` but only calls it inside a `requestAnimationFrame` callback. Usage: Low priority setState.

`useWindowSize`: Returns the window size 🤷‍♂`{ height, width }` in px.


deini:
@jorgemoya Not being used atm but we will use it in your Select / Dropdown PR.

PR #329, opened 2020-01-16
feat(component): add forwardRef to box, flex and grid

rtalvarez:
## What?
Adds a few missing `forwardRef`s

Example use-case I stumbled upon in my app is needing to scroll to certain elements. Since we don't provide users with the element refs we cannot scroll to them programmatically

### DIFF
--- a/packages/big-design/src/components/Box/Box.tsx
+++ b/packages/big-design/src/components/Box/Box.tsx
@@ -17,8 +17,16 @@ export interface BoxProps extends React.HTMLAttributes<HTMLDivElement>, DisplayP
   borderRadius?: keyof BorderRadius;
 }
 
-export const Box: React.FC<BoxProps> = memo(
-  React.forwardRef<HTMLDivElement, BoxProps>((props, ref) => <StyledBox forwardedRef={ref} {...props} />),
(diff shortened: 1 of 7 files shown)


rtalvarez:
errr my vscode didnt pick on the type errors, looking at them now

packages/big-design/src/components/Box/Box.tsx
```diff
@@ -17,8 +17,16 @@ export interface BoxProps extends React.HTMLAttributes<HTMLDivElement>, DisplayP
   borderRadius?: keyof BorderRadius;
 }
 
-export const Box: React.FC<BoxProps> = memo(
-  React.forwardRef<HTMLDivElement, BoxProps>((props, ref) => <StyledBox forwardedRef={ref} {...props} />),
+interface PrivateProps {
+  forwardedRef: React.Ref<HTMLDivElement>;
+}
+
+const RawBox: React.FC<BoxProps & PrivateProps> = memo(
```
deini: There is no need to `forwardRef` here if you are already forwarding when declaring `Box`. 

Also the order should be `memo -> forwardRef`.

PR #332, opened 2020-02-05
fix(component): fix links in Dropdown items and scroll bug

jorgemoya:
Fixes `DropdownLinkItems` `a` tag not spanning the width of the element, thus clicks wouldn't work unless clicking the link.

Links now also matches background transitions of `DropdownItems`.

Fix scroll bug by scheduling Popper to update when user resizes window.

### DIFF
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -5,12 +5,12 @@ import scrollIntoView from 'scroll-into-view-if-needed';
 import { uniqueId } from '../../utils';
 import { Flex } from '../Flex';
 import { FlexItem } from '../Flex/Item';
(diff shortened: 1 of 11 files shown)


packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -137,9 +136,9 @@ export class Dropdown<T extends any> extends React.PureComponent<DropdownProps<T
 
   private wrapInLink(option: DropdownLinkItem<T>, content: React.ReactChild) {
     return (
-      <Link href={option.url} target={option.target}>
+      <a href={option.url} target={option.target}>
         {content}
-      </Link>
+      </a>
```
jordan-massingill: Why was the `<a>` preferable here to the `<Link>` component?

PR #339, opened 2020-02-14
feat(component): restrict badges to use label props and style fixes

chanceaclark:
## What
- Restrict `Badge` component to use label prop.
- Fix styling inconsistencies:
  - line-height grew inside `Text` component
  - extra 1px space on top

### DIFF
--- a/packages/big-design/src/components/Badge/Badge.tsx
+++ b/packages/big-design/src/components/Badge/Badge.tsx
@@ -5,9 +5,12 @@ import { MarginProps } from '../../mixins';
 import { StyledBadge } from './styled';
 
 export interface BadgeProps extends React.HTMLAttributes<HTMLSpanElement>, MarginProps {
+  label: string;
   variant?: 'danger' | 'secondary' | 'success' | 'warning';
 }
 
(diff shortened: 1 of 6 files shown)


packages/big-design/src/components/Badge/Badge.tsx
```diff
@@ -5,9 +5,12 @@ import { MarginProps } from '../../mixins';
 import { StyledBadge } from './styled';
 
 export interface BadgeProps extends React.HTMLAttributes<HTMLSpanElement>, MarginProps {
+  label: string;
   variant?: 'danger' | 'secondary' | 'success' | 'warning';
 }
 
-export const Badge: React.FC<BadgeProps> = memo(({ className, style, ...props }) => <StyledBadge {...props} />);
+export const Badge: React.FC<BadgeProps> = memo(({ className, style, label, ...props }) => (
+  <StyledBadge {...props}>{label}</StyledBadge>
```
deini: What do you think about doing a type check for the label? 🤔

chanceaclark: I think that's wise.

PR #340, opened 2020-02-17
feat(component): add Alerts, InlineAlerts, and Message components

chanceaclark:
## What
Add `Alerts`, `InlineAlerts`, and `Messages`.

## Screenshots
alerts
inline-alerts
messages


packages/big-design/src/components/Alert/styled.ts
```diff
@@ -0,0 +1,79 @@
...
+export const StyledAlert = styled(Grid)<AlertProps>`
+  ${({ theme }) => theme.shadow.floating}
+
+  animation: ${({ theme }) => theme.keyframes.fadeIn} .5s ease-in-out;
+  background-color: ${({ theme }) => theme.colors.white};
+  grid-gap: ${({ theme }) => theme.spacing.small};
+  max-width: ${({ theme }) => theme.helpers.remCalc(456)};
+  padding: ${({ theme }) => theme.spacing.small};
+  position: fixed;
+  right: ${({ theme }) => theme.helpers.remCalc(16)};
```
deini: Is there a reason we aren't using `spacing.medium` for these?

packages/big-design/src/components/Alert/styled.ts
```diff
@@ -0,0 +1,79 @@
...
+      : css`
+          grid-template-areas: 'icon messages';
+          grid-template-columns: ${({ theme }) => `${theme.spacing.xLarge} 1fr`};
+        `}
+
+  ${({ theme, type }) => type && getBorderStyle(type, theme)};
+`;
+
+export const StyledHeader = styled(StyleableH4)`
+  line-height: ${({ theme }) => theme.helpers.remCalc(16)};
```
deini: Same as above.

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+
+      this.remove(key);
+    };
+
+    const newAlert: PrivateAlert = { ...alert, key, onClose };
+
+    this.alerts = this.alerts.concat([newAlert]).sort(this.sortAlerts);
+
+    this.contactSubscribers();
```
deini: :tropical_drink: `notifySubscribers`

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+    let removed;
+
+    this.alerts = this.alerts.reduce((acc, alert) => {
+      if (alert.key === key) {
+        removed = alert;
+
+        return acc;
+      }
+
+      acc.push(alert);
```
deini: :tropical_drink` `return [...acc, alert]` or concat.

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    return removed;
+  }
+
+  subscribe(subscriber: Subscriber) {
+    this.subscribers.push(subscriber);
+  }
+
+  unsubscribe() {
+    this.alerts = [];
```
deini: Is there a reason we are resetting all this here? Seems to me that we only need to find the subscriber and remove it from the list of subscribers :thinking:

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.counter = 0;
+    this.subscribers = [];
+  }
+
+  private contactSubscribers() {
+    this.subscribers.forEach(subscriber => subscriber(this.alerts[0]));
```
deini: Should this send `null` instead of undefined when the alerts are empty? :thinking:

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+    success: 2,
+    info: 3,
+  };
+
+  add(alert: AlertProps, dismissCallback?: () => void): string {
+    if (alert.key && this.containsKey(alert.key)) {
+      this.remove(alert.key);
+    }
+
+    const key = this.getKey(alert.key);
```
deini: What do you think about just inlining it here instead of calling `this.getKey`?

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private getUniqueId() {
+    this.counter += 1;
+
+    return `alert-${this.counter}`;
+  }
+
+  private getKey(key: AlertProps['key']) {
+    return key ? key : this.getUniqueId();
```
deini: Lets check for undefineds here instead of just falsey.

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private sortAlerts = (a: PrivateAlert, b: PrivateAlert) => {
```
deini: Can't this function just be `return a - b`?

packages/big-design/src/utils/messagingHelpers.tsx
```diff
@@ -0,0 +1,63 @@
...
+      return <WarningIcon size={size} color="warning50" />;
+    case 'info':
+      return <InfoIcon size={size} color="primary60" />;
+  }
+};
+
+export const getBorderStyle = (type: MessagingType, theme: ThemeInterface) => css`
+  ${type === 'success' &&
+    css`
+      border-left: 4px solid ${theme.colors.success};
```
deini: Any reason not to use spacings?

packages/big-design/src/utils/messagingHelpers.tsx
```diff
@@ -0,0 +1,63 @@
...
+export interface MessageItem {
+  text: string;
+  link?: MessageLinkItem;
+}
+
+export type MessageLinkItem = Pick<LinkProps, 'external' | 'href' | 'target'> & {
+  text: string;
+};
+
+export const getMessagingIcon = (type: MessagingType, condensed?: boolean) => {
```
deini: Wherever this function ise used, make sure to memo the call. Icons are expensive.

packages/docs/pages/Alert/AlertPage.tsx
```diff
@@ -0,0 +1,148 @@
...
+  type: 'info',
+  onClose: () => null,
+} as AlertProps;
+
+export default () => (
+  <>
+    <H0>Alerts</H0>
+    <Text>An alert appears at the top right of the interface notifying the user after an action.</Text>
+
+    <CodeSnippet>
```
deini: What do you think about having a working example here? I'm just worried that is not consistent with our other components and you can't see how it looks unless you scroll all the way down.

Not sure how to solve for this tho, any ideas?

packages/big-design/src/managers/alerts/AlertsManager.tsx
```diff
@@ -0,0 +1,24 @@
...
+    alertsManager.subscribe(setAlert);
+
+    return () => {
+      alertsManager.unsubscribe();
+    };
+  }, []);
+
+  return (
+    <>
+      {children}
```
deini: I don't think we should make this a wrapper, can we make it so an app just includes `<AlertsManager />` without having to wrap the whole App?

packages/big-design/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,54 @@
...
+import React, { memo, useMemo } from 'react';
+
+import { excludePaddingProps } from '../../mixins';
+import { getMessagingIcon, SharedMessagingProps } from '../../utils';
+import { Box } from '../Box';
+import { GridItem } from '../Grid';
+
+import { StyledAlert, StyledCloseButton, StyledHeader, StyledLink, StyledMessageItem } from './styled';
+
+// TODO:
```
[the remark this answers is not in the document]
chanceaclark: Just a reminder, keeping till all are done.

packages/big-design/src/components/Alert/styled.ts
chanceaclark: Good catch.

packages/docs/pages/Alert/AlertPage.tsx
chanceaclark: I wasn't sure either. This was the best I could think of at the time. I'm open to suggestions.

packages/big-design-icons/svgs/material/info.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24" viewBox="0 0 24 24" width="24"><path d="M0 0h24v24H0V0z" fill="none"/><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm0 15c-.55 0-1-.45-1-1v-4c0-.55.45-1 1-1s1 .45 1 1v4c0 .55-.45 1-1 1zm1-8h-2V7h2v2z"/></svg>
```
golcinho: I see that this is the same icon as an `ErrorIcon`. I think it's better to leave one component instead of duplication.

I propose to rename ErrorIcon to InfoIcon. What do you think?

chanceaclark: The info icon is different than the error icon. One is an `!` while the other is a `i`.
Screen Shot 2020-02-18 at 10 52 08 AM
vs. 
Screen Shot 2020-02-18 at 10 52 12 AM

golcinho: got it, sorry, my bad!

packages/big-design/src/managers/alerts/spec.tsx
```diff
@@ -0,0 +1,114 @@
...
+  alertsManager = createAlertsManager();
+});
+
+describe('alertsManager', () => {
+  const testKey = 'test';
+  const alert = ({
+    text: 'Test',
+  } as unknown) as AlertProps;
+
+  test('adds a message', () => {
```
deini: Can we change the description or the assertion? If it returns a key then it means it added a message? 🤔

jorgemoya:
Can you add some screenshots of the PR?

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,92 @@
...
+      if (typeof dismissCallback === 'function') {
+        dismissCallback(); // Should we return something with this?
+      }
+
+      if (typeof alert.onClose === 'function') {
+        alert.onClose();
+      }
+
+      this.remove(key);
+    };
```
jorgemoya: What was the reasoning behind keeping both `dismissCallback` and `onClose`?

packages/docs/MethodLists/AlertsManagerMethodLists.tsx
```diff
@@ -0,0 +1,60 @@
...
+      },
+    ]}
+    returnDescription="The value of the alert key. If no key is provided, then an auto-generated one will be provided."
+  />
+);
+
+export const AlertsManagerRemoveMethodList: React.FC = () => (
+  <MethodList
+    name="remove"
+    intro="Removes an alert by key and displays the next alert, if possible."
```
jorgemoya: Maybe change to "if available"? 🍹

packages/docs/PropTables/shared.tsx
```diff
@@ -0,0 +1,83 @@
...
+  },
+];
+
+const messagingLinkItemProps: Prop[] = [
+  {
+    name: 'external',
+    types: 'boolean',
+    description: (
+      <>
+        Shows and external icons When the <Code primary>external</Code> flag is set and target="_blank".
```
jorgemoya: `Shows external icons when the the <Code primary>external</Code> flag is set and target="_blank".`?

packages/docs/pages/Message/MessagePage.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+import { Code, CodePreview, Collapsible } from '../../components';
+import { MarginPropTable } from '../../PropTables';
+import { MessagingItemPropTable, MessagingLinkItemPropTable } from '../../PropTables/shared';
+import { MessagePropTable } from '../../PropTables/MessagePropTable';
+
+export default () => (
+  <>
+    <H0>Messages</H0>
+    <Text>An message, used for displaying page/table, feature discovery/system, or non-critical messaging.</Text>
```
jorgemoya: A message*, however the rest of the copy seems a little weird.

chanceaclark: I'll give it a rewrite.

packages/big-design/src/managers/alerts/manager.ts
chanceaclark: I was a little concerned about this too. I'm not sure we even need a `dismissCallback`, but we definitely want the `onClose`.

packages/big-design/src/components/Button/private.ts
```diff
@@ -1 +1,32 @@
-export { StyleableButton } from './Button';
```
jordan-massingill: Why is this file named `private` instead of `styled` like the others?

chanceaclark: We are only exporting private internal usage components in this file. Initially, I tried to have the `MessagingButton` in the `styled` file, but module resolution crashed with circular imports.

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,92 @@
...
+    const newAlert: PrivateAlert = { ...alert, key, onClose };
+
+    this.alerts = this.alerts.concat([newAlert]).sort(this.sortAlerts);
+
+    this.notifySubscribers();
+
+    return key;
+  }
+
+  remove(key: string): AlertProps | undefined {
```
rtalvarez: do you need to specify this return value manually?

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,92 @@
...
+  private sortAlerts = (a: PrivateAlert, b: PrivateAlert) => {
+    return this.typeMap[a.type as keyof TypeMap] - this.typeMap[b.type as keyof TypeMap];
+  };
+
+  private containsKey(key: string) {
+    return !!this.alerts.find(alert => alert.key === key);
+  }
+}
+
+export const createAlertsManager = () => new AlertsManager();
```
rtalvarez: whats the reasoning behind exporting `const foo = () => new AlertsManager()` vs `const foo = new AlertsManager()` ?

packages/docs/pages/Message/MessagePage.tsx
```diff
@@ -0,0 +1,105 @@
...
+          messages={[{ text: 'Required description copy.' }]}
+          marginVertical="medium"
+        />
+      </>
+      {/* jsx-to-string:end */}
+    </CodePreview>
+
+    <H2>onClose</H2>
+
+    <Text>Used the toggle the visibility the close button and provide an onClick callback.</Text>
```
rtalvarez: Same

packages/docs/pages/InlineAlert/InlineAlertPage.tsx
```diff
@@ -0,0 +1,105 @@
...
+          messages={[{ text: 'Required description copy.' }]}
+          marginVertical="medium"
+        />
+      </>
+      {/* jsx-to-string:end */}
+    </CodePreview>
+
+    <H2>onClose</H2>
+
+    <Text>Used the toggle the visibility the close button and provide an onClick callback.</Text>
```
rtalvarez: Sentence sounds funny, consider restructuring

packages/big-design/src/managers/alerts/manager.ts
chanceaclark: I'll move the type to the `let removed;`. It was only picking up `undefined` as the return type.

packages/big-design/src/managers/alerts/manager.ts
chanceaclark: @deini and I were chatting about this... tldr it's kind of weird for a library to export an instance so instead we are exposing a method to create a new instance.

Also, since we _were_ exporting an instance, it was hard to reset the instance when testing it.

rtalvarez: Gotcha 👍

packages/big-design/src/managers/alerts/manager.ts
chanceaclark: We are going to remove this since it doesn't really make sense.

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -0,0 +1,88 @@
...
+    success: 2,
+    info: 3,
+  };
+
+  add(alert: AlertProps): string {
+    if (alert.key && this.containsKey(alert.key)) {
+      this.remove(alert.key);
+    }
+
+    const key = alert.key !== undefined ? alert.key : this.getUniqueId();
```
rtalvarez: 🍹 switch the conditions around like so `const key = alert.key === undefined ? this.getUniqueId() : alert.key` for readability

PR #342, opened 2020-02-19
fix(component): proper panel action button width

chanceaclark:
## What
Fixes this display issue on mobile with `Panels`:
Screen Shot 2020-02-19 at 3 12 30 PM

## Screenshot
Screen Shot 2020-02-19 at 3 14 32 PM


chanceaclark:
> Wondering if this an isolated use case of if we should rather add an `autoWidth` prop to the button or something like that? 🤔

I think it's a design choice that all buttons go full width on mobile.

Though now that you mention it, my PR will cause all buttons inside the Panel to go apply `width: auto`. I'll have to refactor.

PR #345, opened 2020-03-09
feat(component): add description attribute for Radio

capsula4:
Adds `description` attribute to Radio component, as requested by @valterfatia 

Screen Shot 2020-03-16 at 11 19 33 pm

Screen Shot 2020-03-16 at 11 35 59 pm


valterfatia:
@capsula4 @deini can we also add small text as part this change? The helper text in forms was all supposed to be small text (14/20px)

deini:
@valterfatia Small text is already implemented and all other components use it for their helper text.

deini:
If we are doing this for Radio I think we should do it for Checkboxes too 🤔

packages/big-design/src/components/Radio/Radio.tsx
```diff
@@ -52,6 +54,14 @@ const RawRadio: React.FC<RadioProps & PrivateProps> = ({
     warning('label must be either a string or a RadioLabel component.');
   }, [disabled, id, label, labelId]);
 
+  const renderedDescription = useMemo(() => {
+    if (!description || disabled) {
```
deini: Why don't we want to render the description if it's disabled?

packages/big-design/src/components/Radio/Radio.tsx
```diff
@@ -3,11 +3,12 @@ import React, { forwardRef, useMemo, Ref } from 'react';
 import { useUniqueId } from '../../hooks';
 import { typedMemo, warning } from '../../utils';
 
-import { HiddenRadio, RadioContainer, StyledRadio } from './styled';
+import { HiddenRadio, RadioContainer, RadioDescription, RadioLabelContainer, StyledRadio } from './styled';
 import { RadioLabel } from './Label';
 
 interface Props {
   label: React.ReactChild;
+  description?: React.ReactChild;
```
deini: We are trying to make our components more restrictive and config-based. Do you think it makes sense to make this just a `string`? 

The downside would be if there is a need for something like a "Learn more" url. In that case, we would need to change the API to allow for a description and an optional url.

packages/big-design/src/components/Radio/styled.tsx
```diff
@@ -9,11 +9,25 @@ interface StyledRadioProps {
   disabled?: boolean;
 }
 
+export const RadioLabelContainer = styled.div`
+  margin-left: ${({ theme }) => theme.spacing.xSmall};
+`;
+
+RadioLabelContainer.defaultProps = { theme: defaultTheme };
```
chanceaclark: Can we move both these `defaultProps` statements at the bottom of the file?

packages/big-design/src/components/Radio/styled.tsx
```diff
@@ -9,11 +9,25 @@ interface StyledRadioProps {
...
   display: flex;
 `;
 
+export const RadioDescription = styled.span`
+  color: ${({ theme }) => theme.colors.secondary50};
+  display: block;
+  font-size: ${({ theme }) => theme.typography.fontSize.small};
+`;
+
+RadioLabelContainer.defaultProps = { theme: defaultTheme };
```
chanceaclark: same `defaultProps`.

packages/big-design/src/components/Radio/styled.tsx
```diff
@@ -9,11 +9,25 @@ interface StyledRadioProps {
...
+
 export const RadioContainer = styled.div`
-  align-items: center;
+  align-items: flex-start;
   display: flex;
 `;
 
+export const RadioDescription = styled.span`
```
chanceaclark: Consider using the `StyleableSmall` in `Typography/private`. See how we consume it in the Messaging components (`Alert`, `InlineAlert`, and `Message`).

packages/big-design/src/components/Radio/Radio.tsx
capsula4: we don't have styles yet. Disabled render the main text as secondary40, which probably means the description should be something around secondary30? that means almost unreadable? @valterfatia

packages/big-design/src/components/Radio/Radio.tsx
capsula4: Yeah, pretty much that's why the reason I'm allowing everything. I understand we want it as restrictive as possible, but sometimes we might want links inside the text.

packages/big-design/src/components/Radio/Radio.tsx
valterfatia: I was about to say lets do secondary40 on both, but actually the text of the labels isn't meant to change based on the state of the input as per the design spec on https://www.figma.com/file/FnGsTa6UoCWQaQoOA7HobiDb/Form-Fields?node-id=195%3A26 
image

packages/big-design/src/components/Radio/Radio.tsx
valterfatia: I agree, having links within the text will be nice and something we want to do moving forward. @deini how is the help handled on the other inputs? 

Also should we setup an "internal" component for handling help? Looks like a lot of the logic/styles are shared across components.

deini: I agree that we will need links. However, we are trying to go as strict as possible so if we can come up with an API that only allows description string + link would be ideal. 

@valterfatia There is a FormControlDescription component.

valterfatia: Cool! In that case we should be able to use it here too 👍

capsula4: The string+link works in a scenario where you always have a link at the end as in "Learn more" but I don't think that's always the case. For example, some checkout configurations like privacy policy requires the Store privacy policy to be enabled, so we link directly to the store privacy policy within Checkout, which is not strictly a learn more link.

deini: Can’t we make the link text also configurable as well as the target?

capsula4:
I have fixed the disabled state as per Valter's designs.

Screen Shot 2020-03-10 at 3 17 08 pm

Also addressed your feedback. If you are good with it I was thinking of doing the same for checkboxes, but probably in a follow-up PR?

packages/big-design/src/components/Radio/Radio.tsx
capsula4: is not about the link text and target, it's also about where in the description it's located. Doesn't always necessarily need to be at the end.

Screen Shot 2020-03-10 at 3 19 08 pm
Screen Shot 2020-03-10 at 3 19 20 pm

deini: I understand, we had the same issue with Alerts. After chatting with design we ended up agreeing that links would always be in the same place. It keeps things consistent and we can always reword our description. 

I’m not blocking this PR but from experience we will see some snowflake descriptions.

capsula4: the other thing is that sometimes, you might want a link to potentially execute some custom javascript / react modal? I will refactor so it matches alerts, but IMO having very strict properties tend to causes more problems than solutions, specially for something trivial like a description.

deini: Yeah, another scenario is if you need a tooltip somewhere in your description or an icon. That's why we need a bit more guidance on this from design.

capsula4: @valterfatia  any ideas on this? Are you happy for us to restrict radio description to one link max? Or do you foresee us being able to include more than one link, or special behaviours like opening a modal / tooltip?

valterfatia: I think it will be fine if we restrict to a single link at the end as long as the text of the link is configurable. That way we can voice it as "Learn more" or as an action like "Go to x".

capsula4:
I have added the description link as part of the `FormControlDescription` component

Screen Shot 2020-03-16 at 5 32 08 pm

packages/big-design/src/components/Radio/__snapshots__/spec.tsx.snap
```diff
@@ -42,6 +42,7 @@ exports[`render Radio (checked + disabled) 1`] = `
   transition: all 150ms ease-out;
   -webkit-transition-property: border-color,box-shadow;
   transition-property: border-color,box-shadow;
+  background-color: #F6F7FC;
```
capsula4: this fixes the disabled state

packages/big-design/src/components/Radio/__snapshots__/spec.tsx.snap
```diff
@@ -169,6 +169,7 @@ exports[`render Radio (checked) 1`] = `
   transition: all 150ms ease-out;
   -webkit-transition-property: border-color,box-shadow;
   transition-property: border-color,box-shadow;
+  background-color: #FFFFFF;
```
capsula4: this fixes the disabled state

packages/big-design/src/components/Form/Description/Description.tsx
```diff
@@ -1,5 +1,36 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 
+import { LinkProps } from '../../Link';
 import { Small, TextProps } from '../../Typography';
 
-export const FormControlDescription: React.FC<TextProps> = ({ className, style, ...props }) => <Small {...props} />;
+import { StyledLink as FormControlDescriptionLink } from './styled';
```
chanceaclark: We should keep this consistent like our other components.
```suggestion
import { StyledLink } from './styled';
```

packages/big-design/src/components/Form/Description/Description.tsx
```diff
@@ -1,5 +1,36 @@
...
+  style,
+  link,
+  ...props
+}) => {
+  const renderedDescriptionLink = useMemo(() => {
+    if (!link) {
+      return;
+    }
+
+    return <> {link && <FormControlDescriptionLink {...link}>{link.text}</FormControlDescriptionLink>}</>;
```
jorgemoya:
```suggestion
    return link && <FormControlDescriptionLink {...link}>{link.text}</FormControlDescriptionLink>;
```

🍹

PR #348, opened 2020-03-14
feat(component): add toggle button component

pedelman:
Adds new Toggle Button component. This just a step towards implementing Toggle Buttons component and is intentionally not exposed to external users of big-design. I had a quick sync up with @deini and @chanceaclark and this was the recommended first step.

Example of different states:
Apr-12-2020 16-58-10

This is my first component in big-design, so I am very interested to learn from the team on any best practices or improvements that can be made!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ packages/docs/.next
 packages/docs/out
 packages/docs/pages/Dev
 .idea
+.DS_Store
--- a/packages/big-design/src/components/ToggleButton/styled.tsx
+++ b/packages/big-design/src/components/ToggleButton/styled.tsx
@@ -1,0 +1,39 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+interface StyledToggleButtonProps {
+  selected: boolean;
+}
+
+export const StyledToggleButton = styled('button')<StyledToggleButtonProps>`
+  background: ${({ selected, theme }) => (selected ? theme.colors.primary20 : theme.colors.white)};
+  border: 1px solid ${({ selected, theme }) => (selected ? theme.colors.primary30 : theme.colors.secondary30)};
+  border-radius: ${({ theme }) => theme.borderRadius.normal};
+  color: ${({ selected, theme }) => (selected ? theme.colors.primary60 : theme.colors.secondary60)};
+  cursor: pointer;
+  font-size: ${({ theme }) => theme.typography.fontSize.medium};
+  line-height: ${({ theme }) => theme.lineHeight.xLarge};
+  margin: 0;
+  outline: none;
+  padding: 0 ${({ theme }) => theme.spacing.medium};
+  position: relative;
+
+  &[disabled] {
+    background: ${({ selected, theme }) => (selected ? theme.colors.secondary10 : theme.colors.white)};
+    border: 1px solid ${({ theme }) => theme.colors.secondary30};
+    color: ${({ theme }) => theme.colors.secondary40};
+    cursor: not-allowed;
+  }
+
+  &:not([disabled]):hover,
+  &:not([disabled]):active {
+    background: ${({ selected, theme }) => (selected ? theme.colors.primary20 : theme.colors.primary10)};
+    color: ${({ selected, theme }) => (selected ? theme.colors.primary60 : theme.colors.primary50)};
+  }
+
+  &:not([disabled]):focus {
+    box-shadow: 0 0 0 4px ${({ theme }) => theme.colors.primary20};
+  }
+`;
+
+StyledToggleButton.defaultProps = { theme: defaultTheme };
--- a/packages/big-design/src/components/ToggleButton/ToggleButton.tsx
+++ b/packages/big-design/src/components/ToggleButton/ToggleButton.tsx
@@ -1,0 +1,17 @@
+import React from 'react';
+
+import { StyledToggleButton } from './styled';
+
+export interface ToggleButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  selected: boolean;
+}
+
+export const ToggleButton: React.FC<ToggleButtonProps> = React.memo(({ className, style, ...props }) => {
+  const { selected } = props;
+
+  return (
+    <StyledToggleButton aria-pressed={selected} selected={selected} {...props}>
+      {props.children}
+    </StyledToggleButton>
+  );
+});
--- a/packages/big-design/src/components/ToggleButton/__snapshots__/spec.tsx.snap
+++ b/packages/big-design/src/components/ToggleButton/__snapshots__/spec.tsx.snap
@@ -1,0 +1,155 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`render selected and disabled 1`] = `
+.c0 {
+  background: #DBE3FE;
+  border: 1px solid #9EB3FC;
+  border-radius: 0.25rem;
+  color: #0B38D9;
+  cursor: pointer;
+  font-size: 1rem;
+  line-height: 2rem;
+  margin: 0;
+  outline: none;
+  padding: 0 1rem;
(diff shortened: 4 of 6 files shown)


jorgemoya:
I'm out of the loop so I'm wondering what is the current goal of this PR and what are the next steps for this component.

packages/big-design/src/components/ToggleButtons/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+import { StyledToggleButton } from './styled';
+
+export interface ToggleButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  label: string;
+  selected: boolean;
+}
+
+export const ToggleButton: React.FC<ToggleButtonProps> = React.memo(props => {
```
chanceaclark: For most/all of our components we want to restrict adding your own custom styling. This should make sure that doesn't happen:
```suggestion
export const ToggleButton: React.FC<ToggleButtonProps> = React.memo(({ className, style, ...props })=> {
```

packages/big-design/src/components/ToggleButtons/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
chanceaclark: The folder path for this component seems off `components/ToggleButtons/ToggleButton` 🤔. I would set the directory structure for this like so:
```
components/
| – ...
| – ToggleButton/
| –––– index.ts
| –––– spec.tsx
| –––– styled.tsx
| –––– ToggleButton.tsx
```
Alongside that, I would remove the `index.ts` file in favor of a `private.ts` file.

packages/big-design/src/components/ToggleButtons/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+export interface ToggleButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  label: string;
+  selected: boolean;
+}
+
+export const ToggleButton: React.FC<ToggleButtonProps> = React.memo(props => {
+  const { label, selected } = props;
+
+  return (
+    <StyledToggleButton aria-label={label} aria-pressed={selected} selected={selected} {...props}>
```
chanceaclark: You don't need an `aria-label` if it's using a valid button element. You could get rid of the `label` prop in that case.

packages/big-design/src/components/ToggleButtons/index.tsx
```diff
@@ -0,0 +1,3 @@
+import { ToggleButtonProps as _ToggleButtonProps } from './ToggleButton/ToggleButton';
```
chanceaclark: Move this into the `private.ts` file, that way you can uncomment below.

packages/big-design/src/components/index.ts
```diff
@@ -23,5 +23,6 @@ export * from './StatefulTable';
 export * from './Table';
 export * from './Tabs';
 export * from './Textarea';
+export * from './ToggleButtons';
```
chanceaclark: Remove.

packages/big-design/src/components/ToggleButtons/ToggleButton/spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const { container } = render(
+    <ToggleButton label="This is a selected toggle button" selected={true} onClick={onClick}></ToggleButton>,
+  );
+  const button = container.firstChild as HTMLElement;
+
+  fireEvent.click(button);
+
+  expect(onClick).toHaveBeenCalled();
+});
```
chanceaclark: Add a `does not forward styles` test.

packages/big-design/src/components/ToggleButtons/ToggleButton/spec.tsx
```diff
@@ -0,0 +1,48 @@
+import { fireEvent, render } from '@test/utils';
+import 'jest-styled-components';
+import React from 'react';
+
+import { ToggleButton } from './ToggleButton';
+
+test('render selected and enabled', () => {
+  const { container } = render(<ToggleButton label="This is a selected toggle button" selected={true}></ToggleButton>);
```
chanceaclark: When you get rid of label, you'll probably want to push it here:
```suggestion
  const { container } = render(<ToggleButton selected={true}>This is a selected toggle button</ToggleButton>);
```

pedelman:
@chanceaclark the colors arent really coming through on the gif if thats what you are referring to.

packages/big-design/src/components/ToggleButtons/ToggleButton/ToggleButton.tsx
pedelman: My thought was that by requiring it, we are helping developers remember this needs to be added when using icon buttons (which are used as examples in the designs).

packages/big-design/src/components/ToggleButtons/ToggleButton/ToggleButton.tsx
pedelman: I wasnt sure whether we wanted to keep ToggleButton at top level of components or treat it as implementation details of ToggleButtons, happy to more though.

chanceaclark: We should keep it at the top level for now. When we get to the group of `ToggleButtons` we can refactor, if it makes sense.

packages/big-design/src/components/ToggleButtons/ToggleButton/ToggleButton.tsx
chanceaclark: You gain accessibility by the content of a button. This is baked in browser accessibility for all `<button>`'s.

In order to add accessibility to an icon in a button, just use the icons `title` prop:
```tsx
<ToggleButton iconOnly={<FormatAlignLeftIcon title="Align content left." />} />
```

chanceaclark:
> @chanceaclark the colors arent really coming through on the gif if thats what you are referring to.

They look like they don't have the same padding compared to the designs:
Screen Shot 2020-03-16 at 5 23 28 PM
vs.
Screen Shot 2020-03-16 at 5 24 04 PM

pedelman:
@chanceaclark whats the recommended approach for support of condensed ui mode?

deini:
@pedelman Condensed ui will probably be its own theme.

pedelman:
Apologies for the delay in getting these comments addressed, I believe I have fixed everything up. Ready for another round of review!

♻️ @chanceaclark @deini

packages/big-design/src/components/ToggleButton/styled.tsx
```diff
@@ -0,0 +1,39 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+interface StyledToggleButtonProps {
+  selected: boolean;
+}
+
+export const StyledToggleButton = styled('button')<StyledToggleButtonProps>`
```
chanceaclark, line 8: Try extending from `StyleableButton` located in `components/Button/private.ts`. You'll be able to write less CSS and use some of the baked in features our `Button` component has.

.gitignore
```diff
@@ -7,3 +7,4 @@ packages/docs/.next
 packages/docs/out
 packages/docs/pages/Dev
 .idea
+.DS_Store
```
icatalina, line 10: This is specific to macos (same for the `.idea` being specific to an editor), they should probably be setup as global `ignores`.

https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

pedelman:
@chanceaclark and I chatted on this a while back, unlikely to get around to this in the short term. Feel free to take over this branch / revisit any of the code! Apologies for keeping this dragging on.

PR #349, opened 2020-03-17
feat(component): add description attribute for Checkbox

capsula4:
Adds `description` attribute to Checkbox component.  (https://github.com/bigcommerce/big-design/pull/345)

Screen Shot 2020-03-17 at 2 23 12 pm

Screen Shot 2020-03-17 at 2 23 20 pm


packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -3,14 +3,21 @@ import React, { forwardRef, useMemo, Ref } from 'react';
 
 import { useUniqueId } from '../../hooks';
 import { typedMemo, warning } from '../../utils';
+import { FormControlDescription as Description, FormControlDescriptionLinkProps } from '../Form';
```
chanceaclark: We should probably be consistent with the other form fields and not rename the imports.
```suggestion
import { FormControlDescription, FormControlDescriptionLinkProps } from '../Form';
```

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -58,6 +66,14 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
     warning('label must be either a string or a CheckboxLabel component.');
   }, [disabled, hiddenLabel, id, label, labelId]);
 
+  const renderedDescription = useMemo(() => {
+    if (!description) {
+      return null;
+    }
+
+    return <Description link={description.link}>{description.text}</Description>;
+  }, [description]);
```
chanceaclark: I can see this always rerendering this memo since description is a value by reference. To prevent that, you could possibly modify it like so:
```suggestion
  }, [description.link, description.text]);
```

packages/big-design/src/components/Checkbox/spec.tsx
```diff
@@ -20,6 +20,27 @@ describe('render Checkbox', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('description', () => {
```
chanceaclark: 🍹 
```suggestion
  test('with description', () => {
```

packages/docs/pages/Radio/RadioPage.tsx
```diff
@@ -52,7 +52,7 @@ export default () => (
                   link: {
                     text: 'Learn more',
                     target: '_blank',
-                    href: 'http://google.com',
+                    href: 'http://www.bigcommerce.com',
```
chanceaclark: ❤️

packages/docs/pages/Checkbox/CheckboxPage.tsx
```diff
@@ -25,7 +25,26 @@ export default () => (
...
+                disabled={true}
+                description={{
+                  text: 'Description for off.',
+                  link: {
+                    text: 'Learn more',
+                    target: '_blank',
+                    href: 'http://www.bigcommerce.com',
+                  },
+                }}
+              />
```
chanceaclark: Should have caught this one too, but instead of adding `description` on the main example, could we create another example above/below the "Indeterminate" one?

packages/docs/PropTables/CheckboxPropTable.tsx
```diff
@@ -39,3 +72,21 @@ export const CheckboxPropTable: React.FC<PropTableWrapper> = props => (
     <PropTable title="Checkbox" propList={checkboxProps} {...props} />
   </>
 );
+
+export const CheckboxDescriptionPropTable: React.FC<PropTableWrapper> = props => (
+  <PropTable
+    title="CheckboxDescription"
```
chanceaclark: Should have caught this in the `Radio` update, but can we change these prop table `title`'s to match our other prop tables like `Message[MessageItem]`? This one would be `Checkbox[CheckboxDescription]`.

packages/big-design/src/components/Form/Description/spec.tsx
```diff
@@ -17,6 +17,7 @@ test('renders description with link', () => {
         text: 'more',
         href: 'f',
         target: 't',
+        external: true,
```
chanceaclark: This will not show an external link icon unless `target="_blank"`. Was this intended for this addition?

packages/docs/PropTables/CheckboxPropTable.tsx
capsula4: I followed also what it was already there. We currently have `CheckboxLabel`, not `Checkbox[CheckboxLabel]`

I wonder if the "parent wrapping" adds any value

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -3,14 +3,21 @@ import React, { forwardRef, useMemo, Ref } from 'react';
...
 
-import { CheckboxContainer, HiddenCheckbox, StyledCheckbox } from './styled';
+import { CheckboxContainer, CheckboxLabelContainer, HiddenCheckbox, StyledCheckbox } from './styled';
 import { CheckboxLabel } from './Label';
 
 interface Props {
   hiddenLabel?: boolean;
   isIndeterminate?: boolean;
   label: React.ReactChild;
+  description?: CheckboxDescription;
```
deini: Since most of the time we won't have links, what do you guys think having this as `string | CheckboxDescription`?

chanceaclark: I agree

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -58,6 +66,14 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
     warning('label must be either a string or a CheckboxLabel component.');
   }, [disabled, hiddenLabel, id, label, labelId]);
 
+  const link = description && description.link;
+  const text = description && description.text;
+
+  const renderedDescription = useMemo(() => <FormControlDescription link={link}>{text}</FormControlDescription>, [
+    link,
+    text,
+  ]);
```
chanceaclark: IMO, this is a little weird. We might want `renderedDescription` to return `null` if `description` is invalid, and use the non-destructured deps i.e. `[..., description.link, description.text]`. That will also play well with @deini  proposed change above.

capsula4: I could use the non-destructured way, but `description` might still be undefined

capsula4: I have pushed a change to allow description to be either string or object, but I think I still need to destructure the variables. To avoid further confusions, maybe you could code a snippet of what you have in mind for this?

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -58,6 +66,14 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
...
   }, [disabled, hiddenLabel, id, label, labelId]);
 
+  const descriptionlink = typeof description === 'object' ? description.link : undefined;
+  const descriptionText = typeof description === 'object' ? description.text : description;
+
+  const renderedDescription = useMemo(
+    () => descriptionText && <FormControlDescription link={descriptionlink}>{descriptionText}</FormControlDescription>,
+    [descriptionlink, descriptionText],
+  );
+
```
chanceaclark: @capsula4 something like this:
```suggestion
  const renderedDescription = useMemo(() => {
      if (!description) {
        return null;
      }
      
      const text = typeof description === 'string' : description ? description.text;

      return <FormControlDescription link={description.link}>{text}</FormControlDescription>
    },
    [description, description.link, description.text],
  );

```

packages/big-design/src/components/Radio/Radio.tsx
```diff
@@ -60,13 +60,13 @@ const RawRadio: React.FC<RadioProps & PrivateProps> = ({
...
-    }
+  const descriptionlink = typeof description === 'object' ? description.link : undefined;
+  const descriptionText = typeof description === 'object' ? description.text : description;
 
-    return <Description link={description.link}>{description.text}</Description>;
-  }, [description]);
+  const renderedDescription = useMemo(
+    () => descriptionText && <FormControlDescription link={descriptionlink}>{descriptionText}</FormControlDescription>,
+    [descriptionlink, descriptionText],
+  );
```
chanceaclark: Same as above.

packages/big-design/src/components/Checkbox/Checkbox.tsx
capsula4: I don't think that would work, as in order to access description.link and descripion.text would have to first check the type of description.

chanceaclark: 🤔 I think you're right. This is what I tried and worked locally, which is basically your first commit anyways. :dunce-hat:
```tsx
  const renderedDescription = useMemo(() => {
    if (!description) {
      return null;
    }

    const text = typeof description === 'object' ? description.text : description;
    const link = typeof description === 'object' ? description.link : undefined;

    return <FormControlDescription link={link}>{text}</FormControlDescription>;
  }, [description]);
```

capsula4: saying that, I think that what I have now might be better as I'm passing the reference of link instead of description object ? 🤷‍♂ 
I also think it might be quite unlikely that those values could change but 🤷‍♂

chanceaclark: Even if those values change, the parent could cause a rerender on it's children.

So I did some playing around inside the performance tab, and this is what I found to actually avoid unnecessary rerenders:
```tsx
  const renderedDescription = useMemo(() => {
    if (!description) {
      return null;
    }

    if (typeof description === 'string') {
      return <FormControlDescription>{description}</FormControlDescription>;
    }

    return <FormControlDescription link={description.link}>{description.text}</FormControlDescription>;
  }, [description]);
```

chanceaclark: Yeah, I tried playing around more in the performance tab, but yeah, no matter what `link` is always going to cause a rerender since it's a reference, but specifically when `typeof description === 'object'`. There's no way to get around it, whether the assign the variables inside or outside the `useMemo`.

IMO, I'd rather have `text` and `link` assignments inside the `useMemo` scope, but I'll leave that up to you 🍹.

capsula4: Thanks for the taking the time in exploring the performance of both approaches! 🙇

capsula4: Just pushed the change 👍

PR #350, opened 2020-03-24
feat(component): adds grouping options to Select component

jordan-massingill:
Similar to recently updated Dropdown component, adds ability to create groups in a Select component with labels. 

Screen Shot 2020-03-26 at 12 32 09 PMScreen Shot 2020-03-26 at 12 31 47 PM

### DIFF
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -6,10 +6,11 @@ import { useUniqueId } from '../../hooks';
 import { Flex } from '../Flex';
 import { FlexItem } from '../Flex/Item';
 import { List } from '../List';
+import { ListGroupHeader } from '../List/GroupHeader';
 import { ListItem } from '../List/Item';
 import { Tooltip } from '../Tooltip';
 
-import { StyledLink, StyledListGroupHeader } from './styled';
+import { StyledLink } from './styled';
 import { DropdownItem, DropdownItemGroup, DropdownLinkItem, DropdownProps } from './types';
 
 export const Dropdown = memo(
@@ -144,7 +145,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
-            <StyledListGroupHeader role="group">{group.label.toUpperCase()}</StyledListGroupHeader>
+            <ListGroupHeader>{group.label.toUpperCase()}</ListGroupHeader>
             {renderItems(group.items)}
           </>
         );
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -144,7 +144,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
-            <StyledListGroupHeader>{group.label.toUpperCase()}</StyledListGroupHeader>
+            <StyledListGroupHeader role="group">{group.label.toUpperCase()}</StyledListGroupHeader>
             {renderItems(group.items)}
           </>
         );
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -6,11 +6,10 @@ import { useUniqueId } from '../../hooks';
 import { Flex } from '../Flex';
 import { FlexItem } from '../Flex/Item';
 import { List } from '../List';
-import { ListGroupHeader } from '../List/GroupHeader';
 import { ListItem } from '../List/Item';
 import { Tooltip } from '../Tooltip';
 
-import { StyledLink } from './styled';
+import { StyledLink, StyledListGroupHeader } from './styled';
 import { DropdownItem, DropdownItemGroup, DropdownLinkItem, DropdownProps } from './types';
 
 export const Dropdown = memo(
@@ -145,7 +144,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
-            <ListGroupHeader>{group.label}</ListGroupHeader>
+            <StyledListGroupHeader>{group.label.toUpperCase()}</StyledListGroupHeader>
             {renderItems(group.items)}
           </>
         );
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -145,7 +145,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
-            <ListGroupHeader>{group.label.toUpperCase()}</ListGroupHeader>
+            <ListGroupHeader>{group.label}</ListGroupHeader>
             {renderItems(group.items)}
           </>
         );
--- a/packages/big-design/src/components/Dropdown/styled.ts
+++ b/packages/big-design/src/components/Dropdown/styled.ts
@@ -2,6 +2,7 @@ import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import styled from 'styled-components';
 
 import { withTransition } from '../../mixins/transitions';
+import { StyledGroupHeader } from '../List/GroupHeader/styled';
 
 export const StyledLink = styled.a`
   ${withTransition(['background-color', 'color'])}
@@ -18,4 +19,12 @@ export const StyledLink = styled.a`
   }
 `;
 
+export const StyledListGroupHeader = styled(StyledGroupHeader)`
+  font-weight: ${({ theme }) => theme.typography.fontWeight.semiBold};
+  font-size: ${({ theme }) => theme.typography.fontSize.small};
+  margin-top: ${({ theme }) => theme.spacing.medium};
+  color: ${({ theme }) => theme.colors.secondary50};
+`;
+
 StyledLink.defaultProps = { theme: defaultTheme };
+StyledListGroupHeader.defaultProps = { theme: defaultTheme };
--- a/packages/big-design/src/components/Dropdown/styled.ts
+++ b/packages/big-design/src/components/Dropdown/styled.ts
@@ -2,7 +2,6 @@ import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import styled from 'styled-components';
 
 import { withTransition } from '../../mixins/transitions';
-import { StyledGroupHeader } from '../List/GroupHeader/styled';
 
 export const StyledLink = styled.a`
   ${withTransition(['background-color', 'color'])}
@@ -19,12 +18,4 @@ export const StyledLink = styled.a`
   }
 `;
 
-export const StyledListGroupHeader = styled(StyledGroupHeader)`
-  font-weight: ${({ theme }) => theme.typography.fontWeight.semiBold};
-  font-size: ${({ theme }) => theme.typography.fontSize.small};
-  margin-top: ${({ theme }) => theme.spacing.medium};
-  color: ${({ theme }) => theme.colors.secondary50};
-`;
-
 StyledLink.defaultProps = { theme: defaultTheme };
-StyledListGroupHeader.defaultProps = { theme: defaultTheme };
--- a/packages/big-design/src/components/List/GroupHeader/ListGroupHeader.tsx
+++ b/packages/big-design/src/components/List/GroupHeader/ListGroupHeader.tsx
@@ -5,7 +5,7 @@ import { StyledGroupHeader } from './styled';
 export const ListGroupHeader: React.FC<React.LiHTMLAttributes<
   HTMLLIElement
 >> = memo(({ className, style, value, ...rest }) => (
-  <StyledGroupHeader {...rest} tabIndex={-1} onMouseDown={preventFocus} aria-hidden />
+  <StyledGroupHeader {...rest} tabIndex={-1} onMouseDown={preventFocus} aria-hidden={true} />
 ));
 
 function preventFocus(event: React.MouseEvent<HTMLLIElement, MouseEvent>) {
--- a/packages/big-design/src/components/List/GroupHeader/ListGroupHeader.tsx
+++ b/packages/big-design/src/components/List/GroupHeader/ListGroupHeader.tsx
@@ -5,7 +5,7 @@ import { StyledGroupHeader } from './styled';
 export const ListGroupHeader: React.FC<React.LiHTMLAttributes<
   HTMLLIElement
 >> = memo(({ className, style, value, ...rest }) => (
-  <StyledGroupHeader {...rest} tabIndex={-1} onMouseDown={preventFocus} role="group" />
+  <StyledGroupHeader {...rest} tabIndex={-1} onMouseDown={preventFocus} aria-hidden />
 ));
 
 function preventFocus(event: React.MouseEvent<HTMLLIElement, MouseEvent>) {
--- a/packages/big-design/src/components/List/GroupHeader/styled.tsx
+++ b/packages/big-design/src/components/List/GroupHeader/styled.tsx
@@ -7,13 +7,14 @@ export const StyledGroupHeader = styled.li<React.HTMLProps<HTMLLIElement>>`
   color: ${({ theme }) => theme.colors.secondary50};
   cursor: default;
   display: flex;
-  font-weight: ${({ theme }) => theme.typography.fontWeight.semiBold};
   font-size: ${({ theme }) => theme.typography.fontSize.small};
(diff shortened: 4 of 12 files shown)


jorgemoya:
Not sure why `yarn.lock` changed if `package.json` is the same. Any idea?

packages/big-design/src/components/Select/types.ts
```diff
@@ -38,3 +38,8 @@ export interface SelectOption<T> extends Omit<BaseItem, 'actionType'> {
 export interface SelectAction extends BaseItem {
   onActionClick(inputText: string | null): void;
 }
+
+export interface SelectOptionGroup {
+  optionsLabel: string;
+  options: Array<SelectOption<any>>;
+}
```
jorgemoya: I believe `SelectOptionGroup` should accept a `<T>` type that is passed down to `SelectOption` so that we don't lose the type of the value.

rtalvarez:
> Not sure why `yarn.lock` changed if `package.json` is the same. Any idea?

My guess is this is the `yarn.lock` from another branch, that had a modified `package.json` .. or something like that?

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -366,6 +403,22 @@ const renderIcon = <T extends any>(item: SelectOption<T> | SelectAction, isHighl
...
+    : (items as Array<SelectOption<any>>);
+};
+
+const isGroup = (item: SelectOption<any> | SelectOptionGroup) => {
+  return 'options' in item && !('value' in item);
+};
+
+const isOption = (item: SelectOption<any> | SelectOptionGroup) => {
+  return 'value' in item && !('options' in item);
+};
```
jorgemoya: Don't want `any` as types. Will need to move this inside the FC.

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -256,7 +262,7 @@ export const Select = typedMemo(
 
     const renderAction = useCallback(
       (actionItem: SelectAction) => {
-        const index = options.length;
+        const index = onlyOptions.length - 1;
```
jorgemoya: I believe this should be `selectOptions.length - 1`.

packages/big-design/src/components/Select/types.ts
jordan-massingill: Had some trouble passing `<T>` type to `SelectOptionGroup` which is why I passed `<any>`, assumed you **_could_** pass `<any>` type and then the `SelectOptionGroup` itself would apply the generic typing to it - happy to go through it and try something different though.

packages/big-design/src/components/Select/styled.tsx
```diff
@@ -19,6 +20,13 @@ export const StyledDropdownIcon = styled(ArrowDropDownIcon)<{ disabled?: boolean
     `}
 `;
 
+export const StyledListGroupHeader = styled(StyledGroupHeader)`
+  font-family: 'Source Sans Pro SemiBold', 'Helvetica Neue', Arial, sans-serif;
+  font-size: ${({ theme }) => theme.helpers.remCalc(14)};
+  margin-top: 1rem;
```
chanceaclark: Can we use the `theme.spacing` values instead?

packages/big-design/src/components/Select/styled.tsx
```diff
@@ -19,6 +20,13 @@ export const StyledDropdownIcon = styled(ArrowDropDownIcon)<{ disabled?: boolean
     `}
 `;
 
+export const StyledListGroupHeader = styled(StyledGroupHeader)`
+  font-family: 'Source Sans Pro SemiBold', 'Helvetica Neue', Arial, sans-serif;
```
chanceaclark: You probably want to use `font-weight` instead of `font-family` to change this type of styling.

packages/big-design/src/components/Select/styled.tsx
```diff
@@ -19,6 +20,13 @@ export const StyledDropdownIcon = styled(ArrowDropDownIcon)<{ disabled?: boolean
...
+  font-size: ${({ theme }) => theme.helpers.remCalc(14)};
```
chanceaclark: 🤔I feel like there might be this value on `theme.typography.fontSize` already.

packages/big-design/src/components/Select/types.ts
```diff
@@ -38,3 +38,8 @@ export interface SelectOption<T> extends Omit<BaseItem, 'actionType'> {
 export interface SelectAction extends BaseItem {
   onActionClick(inputText: string | null): void;
 }
+
+export interface SelectOptionGroup<T> {
+  groupLabel: string;
```
chanceaclark: What does the `Dropdown` group name this?

packages/big-design/src/components/Select/styled.tsx
```diff
@@ -19,6 +20,13 @@ export const StyledDropdownIcon = styled(ArrowDropDownIcon)<{ disabled?: boolean
...
+  margin-top: 1rem;
+  color: ${({ theme }) => theme.colors.secondary50};
```
chanceaclark: This might be a question for design, but should we unify all these stylings for both `Dropdown` and `Select` group headers?

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -37,14 +37,39 @@ export const Select = typedMemo(
...
+    const isOption = useCallback((item: SelectOption<T> | SelectOptionGroup<T>) => {
+      return 'value' in item && !('options' in item);
+    }, []);
+
+    const flattenOptions = useCallback(
+      (items: Array<SelectOption<T> | SelectOptionGroup<T>>): Array<SelectOption<T>> => {
+        return items.every(isGroup)
+          ? (items as Array<SelectOptionGroup<T>>)
+              .map((group: SelectOptionGroup<T>) => group.options)
+              .reduce((acum, curr) => acum.concat(curr), [])
```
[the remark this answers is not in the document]
jordan-massingill: Good call. Was repurposing function from `Dropdown`, will update.

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -37,14 +37,39 @@ export const Select = typedMemo(
...
+          : (items as Array<SelectOption<T>>);
+      },
+      [isGroup],
+    );
+
     // Merge options and action
-    const initialOptions = useMemo(() => (action ? [...options, action] : options), [action, options]);
+    const onlyOptions = useMemo(() => (action ? [...flattenOptions(options), action] : flattenOptions(options)), [
```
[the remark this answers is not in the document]
jordan-massingill: Another repurpose from `Dropdown` - trying to keep naming consistent. I'm fine with renaming this variable.

packages/big-design/src/components/Select/types.ts
jordan-massingill: `Dropdown` names this `DropdownItemGroup`. 

Since `Dropdown` has the `item` attribute while `Select` has the `option` attribute, I took the naming convention from `Dropdown` and replaced `Item` with `Option`. Suggestions?

packages/big-design/src/components/Select/styled.tsx
jordan-massingill: I'll confer w/design and ask for clarification. Should that be a separate ticket for `Dropdown`?

packages/big-design/src/components/Select/Select.tsx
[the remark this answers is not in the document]
jordan-massingill: Actually a) `concat` does not mutate the original array, and b) doing it the way suggested here throws a type error. Will play around with it to see if I can fix the error.

packages/big-design/src/components/Select/types.ts
chanceaclark: I was more thinking about the naming for `groupLabel`.

packages/big-design/src/components/Select/styled.tsx
chanceaclark: No, we should probably tackle that within this one.

packages/big-design/src/components/Select/types.ts
jordan-massingill: AH! Gotcha. Yeah, `Dropdown` calls this `label`, however since the `Select` component has a `label` attribute, I was attempting to avoid confusion. Is it your opinion that I should go ahead and call this `label`?

chanceaclark: We should probably keep these consistent so that we don't have to reteach consumers how to use a component.

jorgemoya: Agree 👍

packages/big-design/src/components/Dropdown/styled.ts
```diff
@@ -18,4 +19,12 @@ export const StyledLink = styled.a`
   }
 `;
 
+export const StyledListGroupHeader = styled(StyledGroupHeader)`
+  font-weight: ${({ theme }) => theme.typography.fontWeight.semiBold};
+  font-size: ${({ theme }) => theme.typography.fontSize.small};
+  margin-top: ${({ theme }) => theme.spacing.medium};
```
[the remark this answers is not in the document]
jordan-massingill: This was requested by design, see comments on the ticket.

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -37,14 +37,37 @@ export const Select = typedMemo(
     value,
     ...rest
   }: SelectProps<T>): ReturnType<React.FC<SelectProps<T>>> => {
+    const isGroup = useCallback((item: SelectOption<T> | SelectOptionGroup<T>) => {
+      return 'options' in item && !('value' in item);
+    }, []);
```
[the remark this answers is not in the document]
jordan-massingill: It was originally outside of the component, however @jorgemoya helped me troubleshoot a typing issue and suggested moving inside the component in order to receive the `<T>` generic type. Would it be better to move outside as I had it originally and instead use the `<T extends any>` generic type?

packages/big-design/src/components/Dropdown/styled.ts
```diff
@@ -18,4 +19,12 @@ export const StyledLink = styled.a`
   }
 `;
 
+export const StyledListGroupHeader = styled(StyledGroupHeader)`
```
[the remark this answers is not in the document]
jordan-massingill: I'm not sure I understand your suggestion - Do you mean update the `ListGroupHeader` component, or do you mean to do `styled(ListGroupHeader)` instead of `styled(StyledGroupHeader)`? The latter can't be done due to `className` and `style` being destructured and removed from the `ListGroupHeader`, and I didn't want to affect the `List` component by changing the original `ListGroupHeader` without direction from design. Or did you mean something else?

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -145,7 +144,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
-            <ListGroupHeader>{group.label}</ListGroupHeader>
+            <StyledListGroupHeader role="group">{group.label.toUpperCase()}</StyledListGroupHeader>
```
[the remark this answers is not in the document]
jordan-massingill: I added it for the tests as they were failing without it (but they can easily be updated), and because the `ListGroupHeader` element has that attribute added while the `StyledListGroupHeader` does not.

packages/big-design/src/components/Dropdown/styled.ts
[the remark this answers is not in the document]
jorgemoya: I'm thinking he meant to update `StyledGroupHeader` instead of extending it and creating a new `Component`.

packages/big-design/src/components/Select/Select.tsx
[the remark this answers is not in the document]
jorgemoya: Like @jordan-massingill mentioned, she needs it inside the component to have access to the `<T>` generic binding afaik.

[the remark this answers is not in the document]
jordan-massingill: Per a video chat, I'll be doing the `<T extends any>` typing and move it outside of the component.

packages/big-design/src/components/Dropdown/styled.ts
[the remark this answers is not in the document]
jordan-massingill: Per video conversation, will be moving these styles to the `StyledGroupHeader`

packages/big-design/src/components/Dropdown/styled.ts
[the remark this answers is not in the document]
jordan-massingill: Per video chat with design, will be applying this styling to all but the first group labels.

packages/big-design/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
jordan-massingill: Per video conversation, updating tests and removing this role, also adding `aria-hidden` to the `ListGroupHeader` component

packages/big-design/src/components/List/GroupHeader/ListGroupHeader.tsx
```diff
@@ -5,7 +5,7 @@ import { StyledGroupHeader } from './styled';
 export const ListGroupHeader: React.FC<React.LiHTMLAttributes<
   HTMLLIElement
 >> = memo(({ className, style, value, ...rest }) => (
-  <StyledGroupHeader {...rest} tabIndex={-1} onMouseDown={preventFocus} role="group" />
+  <StyledGroupHeader {...rest} tabIndex={-1} onMouseDown={preventFocus} aria-hidden />
```
chanceaclark: `aria-hidden` actually needs a value, otherwise the user agent determines if the element is exposed to or hidden from the accessibility API.
```suggestion
  <StyledGroupHeader {...rest} tabIndex={-1} onMouseDown={preventFocus} aria-hidden={true} />
```

packages/big-design/src/components/List/GroupHeader/styled.tsx
```diff
@@ -4,14 +4,20 @@ import styled from 'styled-components';
 export const StyledGroupHeader = styled.li<React.HTMLProps<HTMLLIElement>>`
   align-items: center;
   box-sizing: border-box;
-  color: ${({ theme }) => theme.colors.secondary40};
+  color: ${({ theme }) => theme.colors.secondary50};
   cursor: default;
   display: flex;
+  font-weight: ${({ theme }) => theme.typography.fontWeight.semiBold};
+  font-size: ${({ theme }) => theme.typography.fontSize.small};
```
chanceaclark: 🍹 🔤

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -145,7 +145,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
-            <ListGroupHeader>{group.label}</ListGroupHeader>
+            <ListGroupHeader>{group.label.toUpperCase()}</ListGroupHeader>
```
chanceaclark: 🍹 Maybe instead of `.toUpperCase` we might just want to add `text-transform: uppercase;` instead?

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -277,41 +285,96 @@ export const Select = typedMemo(
...
+      [getItemProps, highlightedIndex, isOpen, selectedOption, selectOptions],
+    );
 
-          const isHighlighted = highlightedIndex === index;
-          const isSelected = selectedOption ? 'value' in item && selectedOption.value === item.value : false;
+    const renderGroup = useCallback(
+      (group: SelectOptionGroup<T>) => {
+        return (
+          <>
+            <ListGroupHeader>{group.label.toUpperCase()}</ListGroupHeader>
```
chanceaclark: Same 🍹

packages/big-design/src/components/Dropdown/Dropdown.tsx
jordan-massingill: The element is not recognizing this style for some reason. Troubleshooting now.

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -277,41 +285,96 @@ export const Select = typedMemo(
...
+    const renderOptions = useCallback(
+      (items: Array<SelectOption<T>>) => {
+        return (
+          isOpen &&
+          items.map(item => {
+            if (
+              !selectOptions.find(
+                (option: SelectOption<T> | SelectAction) => 'value' in option && option.value === item.value,
+              )
+            ) {
```
chanceaclark, line 300: 🍹🍹🍹
```suggestion
            if (!selectOptions.find(option => 'value' in option && option.value === item.value)) {
```

PR #359, opened 2020-04-02
feat(component): CHP-6225 adds ability to hide individual table headers

jordan-massingill:
### WHAT
Adds ability to hide individual headers on the `Table` and `Stateful Table` components without hiding all of them.

### Testing
Manual testing, see screenshot below 

Screen Shot 2020-04-02 at 10 41 46 AM

### DIFF
--- a/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
+++ b/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
@@ -65,8 +65,9 @@ const InternalHeaderCell = <T extends TableItem>({
     >
       <StyledFlex alignItems="center" flexDirection="row" hide={hide} align={align}>
         {children}
-        {renderSortIcon()}
+        {!hide && renderSortIcon()}
       </StyledFlex>
+      {hide && renderSortIcon()}
     </StyledTableHeaderCell>
   );
 };
--- a/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
+++ b/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
@@ -10,6 +10,7 @@ import { StyledFlex, StyledTableHeaderCell, StyledTableHeaderCheckbox } from './
 export interface HeaderCellProps<T> extends React.TableHTMLAttributes<HTMLTableCellElement> {
   actionsRef: RefObject<HTMLDivElement>;
   column: TableColumn<T>;
+  hide?: boolean;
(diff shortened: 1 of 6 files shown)


packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -61,7 +63,7 @@ const InternalHeaderCell = <T extends TableItem>({
       width={width}
       stickyHeight={actionsSize.height}
     >
-      <StyledFlex alignItems="center" flexDirection="row" align={align}>
+      <StyledFlex alignItems="center" flexDirection="row" hide={hide} align={align}>
```
chanceaclark, line 66: Just a 💭, but do you think we should disable sorting functionality if a header is visually hidden? Are there any use-cases in which we might want to sort when it's visually hidden?

packages/docs/PropTables/StatefulTablePropTable.tsx
```diff
@@ -82,6 +82,13 @@ const tableColumnsProps: Prop[] = [
     required: true,
     description: 'Header title.',
   },
+  {
+    name: 'hideHeader',
```
chanceaclark, line 86: Can you think of some other names that might be more verbose of what the prop does under the surface?

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
jordan-massingill: I thought about it but couldn't come to a definite decision. Probably a good question for design, but I can see the possibility of wanting to sort a column that doesn't have a label. Might be good to have a group discussion around this.

packages/docs/PropTables/StatefulTablePropTable.tsx
jordan-massingill: Most of mine were similar to this one - `hiddenHeader`, `headerless` (similar to the parent `Table` component), `noHeader` - none of which seemed quite right. Do you have any ideas?

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
jordan-massingill: Per conversation with design + others, simplest solution for the time being is to continue to allow sorting and to render the sort arrow icon in order to make that functionality appear intentional rather than a bug. If, in the future, this solution no longer meets the needs of users or causes problems, it will be addressed then.

PR #371, opened 2020-04-14
feat(component): adds Counter component to bigdesign

jordan-massingill:
### WHAT
Adds a `Counter` component to the component library 

### TESTING
Manual testing

Screen Shot 2020-04-16 at 3 26 30 PMScreen Shot 2020-04-16 at 3 26 41 PM

### DIFF
--- a/packages/big-design/src/components/Counter/Counter.tsx
+++ b/packages/big-design/src/components/Counter/Counter.tsx
@@ -40,14 +40,6 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
     const uniqueCounterId = useUniqueId('counter');
     const id = props.id ? props.id : uniqueCounterId;
 
-    const decreaseIconColor = () => {
-      return value <= min ? 'secondary' : 'secondary60';
-    };
-
-    const increaseIconColor = () => {
-      return value >= max ? 'secondary' : 'secondary60';
-    };
-
     const handleFocus = (event: React.FocusEvent<HTMLInputElement>) => {
       const { onFocus } = props;
 
@@ -194,7 +186,7 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
           <StyledCounterButton
             disabled={value <= min}
             onClick={handleDecrease}
-            iconOnly={<RemoveCircleOutlineIcon title="Decrease count" color={decreaseIconColor()} />}
+            iconOnly={<RemoveCircleOutlineIcon title="Decrease count" />}
           />
           <StyledCounterInput
             {...props}
@@ -211,7 +203,7 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
           <StyledCounterButton
             disabled={value >= max}
             onClick={handleIncrease}
-            iconOnly={<AddCircleOutlineIcon title="Increase count" color={increaseIconColor()} />}
+            iconOnly={<AddCircleOutlineIcon title="Increase count" />}
           />
         </StyledCounterWrapper>
       </div>
--- a/packages/big-design/src/components/Counter/Counter.tsx
+++ b/packages/big-design/src/components/Counter/Counter.tsx
@@ -5,7 +5,7 @@ import { useUniqueId } from '../../hooks';
 import { typedMemo, warning } from '../../utils';
 import { FormControlDescription, FormControlError, FormControlLabel } from '../Form';
 
-import { StyledCounterInput, StyledCounterWrapper } from './styled';
+import { StyledCounterButton, StyledCounterInput, StyledCounterWrapper } from './styled';
 
 export interface CounterProps extends React.InputHTMLAttributes<HTMLInputElement> {
   label?: React.ReactChild;
@@ -41,11 +41,11 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
     const id = props.id ? props.id : uniqueCounterId;
 
     const decreaseIconColor = () => {
-      return value <= min ? 'secondary' : undefined;
+      return value <= min ? 'secondary' : 'secondary60';
     };
 
     const increaseIconColor = () => {
-      return value >= max ? 'secondary' : undefined;
+      return value >= max ? 'secondary' : 'secondary60';
     };
 
     const handleFocus = (event: React.FocusEvent<HTMLInputElement>) => {
@@ -64,7 +64,9 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
       return onBlur && onBlur(event);
     };
 
-    const handleIncrease = () => {
+    const handleIncrease = (event: React.KeyboardEvent | React.MouseEvent) => {
+      event.preventDefault();
+
       if (value + step > max) {
         return;
       }
@@ -78,7 +80,9 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
       }
     };
 
-    const handleDecrease = () => {
+    const handleDecrease = (event: React.KeyboardEvent | React.MouseEvent) => {
+      event.preventDefault();
+
       if (value - step < min) {
         return;
       }
@@ -107,10 +111,10 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
     const handleKeyPress = (event: React.KeyboardEvent) => {
       switch (event.key) {
         case 'ArrowUp':
-          handleIncrease();
+          handleIncrease(event);
           break;
         case 'ArrowDown':
-          handleDecrease();
+          handleDecrease(event);
           break;
         default:
           break;
@@ -187,7 +191,11 @@ export const StylableCounter: React.FC<CounterProps & PrivateProps> = typedMemo(
         {renderedLabel}
         {renderedDescription}
         <StyledCounterWrapper disabled={disabled} error={errors} focus={focus}>
-          <RemoveCircleOutlineIcon onClick={handleDecrease} color={decreaseIconColor()} />
+          <StyledCounterButton
+            disabled={value <= min}
+            onClick={handleDecrease}
+            iconOnly={<RemoveCircleOutlineIcon title="Decrease count" color={decreaseIconColor()} />}
+          />
           <StyledCounterInput
             {...props}
             ref={forwardedRef}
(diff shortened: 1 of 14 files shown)


packages/big-design/src/components/Counter/private.ts
```diff
@@ -0,0 +1,19 @@
+import { remCalc, theme as defaultTheme } from '@bigcommerce/big-design-theme';
```
deini, line 1: Is there a reason why this file is called `private.ts` and not `styled.ts`?

packages/big-design/src/components/Counter/private.ts
```diff
@@ -0,0 +1,19 @@
...
+import styled from 'styled-components';
+
+import { StyledInput, StyledInputWrapper, StyledInputWrapperProps } from '../Input/styled';
```
deini, line 4: Let's make this import from `../Input/private` and not directly from `styled`

chanceaclark:
We will probably want to remove the ticket number from the PR title.

packages/big-design/src/components/Counter/private.ts
jordan-massingill: I misunderstood advice on how to best implement this 🙅

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+      return validateError(error);
+    }, [error]);
+
+    return (
+      <div>
+        {renderedLabel}
+        {renderedDescription}
+        <StyledCounterWrapper disabled={disabled} error={errors} focus={focus}>
+          <RemoveCircleOutlineIcon onClick={handleDecrease} color={decreaseIconColor()} />
```
chanceaclark, line 190: One thing I noticed about these icons is that you can't tab over them/handle keyboard events. To kind of simplify things, you could wrap these in an extended `StyleableButton`, and you'll get baked in accessibility.

packages/big-design/src/components/Counter/styled.ts
```diff
@@ -0,0 +1,19 @@
+import { remCalc, theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+import { StyledInput, StyledInputWrapper, StyledInputWrapperProps } from '../Input/private';
+
+export const StyledCounterWrapper = styled(StyledInputWrapper)<StyledInputWrapperProps>`
+  padding-left: ${({ theme }) => theme.spacing.xxSmall};
+  padding-right: ${({ theme }) => theme.spacing.xxSmall};
+  width: ${remCalc(125)};
+  max-width: ${remCalc(320)};
```
chanceaclark, line 10: Having `max-width` and `width` doesn't really make sense. You'll probably just want to pick one or the other.

packages/big-design/src/components/Counter/styled.ts
```diff
@@ -0,0 +1,19 @@
+import { remCalc, theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+import { StyledInput, StyledInputWrapper, StyledInputWrapperProps } from '../Input/private';
+
+export const StyledCounterWrapper = styled(StyledInputWrapper)<StyledInputWrapperProps>`
+  padding-left: ${({ theme }) => theme.spacing.xxSmall};
+  padding-right: ${({ theme }) => theme.spacing.xxSmall};
```
chanceaclark, line 8: 🍹 
```suggestion
  padding: 0 ${({ theme }) => theme.spacing.xxSmall};
```

packages/big-design/src/components/Counter/styled.ts
```diff
@@ -0,0 +1,19 @@
...
+  width: ${remCalc(125)};
+  max-width: ${remCalc(320)};
+`;
+
+export const StyledCounterInput = styled(StyledInput)`
+  text-align: center;
+  padding: ${({ theme }) => theme.spacing.none};
```
chanceaclark, line 15: 🍹 
```suggestion
  padding: 0;
```

packages/big-design/src/components/Counter/Counter.tsx
chanceaclark: We'll probably also want a `title` on the SVGs.

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+    value,
+    onCountChange,
+    ...props
+  }) => {
+    const [focus, setFocus] = useState(false);
+    const uniqueCounterId = useUniqueId('counter');
+    const id = props.id ? props.id : uniqueCounterId;
+
+    const decreaseIconColor = () => {
+      return value <= min ? 'secondary' : undefined;
```
chanceaclark, line 44: One more thing. From the designs, it looks like the base color of the +/- icons are actually `secondary60`.

packages/docs/pages/Counter/CounterPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+      Supports most native <Code>&lt;input /&gt;</Code> element attributes.
+    </Text>
+
+    <CounterPropTable />
+
+    <H1>Error State</H1>
+
+    <Text>
+      Counters allow you to pass in an <Code primary>error</Code> message that will control the styles of a counter. The
+      logic on the counter can be controlled with the <Code primary>onChange</Code> prop.
```
deini, line 57:
```suggestion
      Counters allow you to pass in an <Code primary>error</Code> message that will control the styles of a counter. 
```
We can get rid of that sentence or correct the prop name to `onCounterChange`

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,227 @@
+import { AddCircleOutlineIcon, RemoveCircleOutlineIcon } from '@bigcommerce/big-design-icons';
+import React, { forwardRef, useMemo, useState, Ref } from 'react';
+
+import { useUniqueId } from '../../hooks';
+import { typedMemo, warning } from '../../utils';
+// import { StyleableButton } from '../Button/private';
```
deini, line 6: Extra comment 🙏

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+    return (
+      <div>
+        {renderedLabel}
+        {renderedDescription}
+        <StyledCounterWrapper disabled={disabled} error={errors} focus={focus}>
+          <StyledCounterButton
+            disabled={value <= min}
+            onClick={handleDecrease}
+            iconOnly={<RemoveCircleOutlineIcon title="Decrease count" color={decreaseIconColor()} />}
```
chanceaclark, line 197: You could possibly just pass it's base color in, and let disabled styles kick in whenever it gets to that state.

jordan-massingill: Since the icon inline style props override the styled component/css styles, I've moved all icon colors into the styled component.

packages/big-design/src/components/Counter/styled.ts
```diff
@@ -0,0 +1,34 @@
+import { remCalc, theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+import { StyleableButton } from '../Button/private';
+import { StyledInput, StyledInputWrapper, StyledInputWrapperProps } from '../Input/private';
+
+export const StyledCounterButton = styled(StyleableButton)`
```
chanceaclark, line 7: There may be some other things you need to override here (`height`, `width`, `padding`, etc.). I would also check to see what this looks like on mobile.

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,218 @@
...
+      return validateError(error);
+    }, [error]);
+
+    return (
+      <div>
+        {renderedLabel}
+        {renderedDescription}
+        <StyledCounterWrapper disabled={disabled} error={errors} focus={focus}>
+          <StyledCounterButton
+            disabled={value <= min}
```
chanceaclark, line 187: You might also want to check to see if you disabled the component. Also, add/modify some tests around this.

jordan-massingill: Done 👍

packages/big-design/src/components/Counter/styled.ts
jordan-massingill: I don't see any obvious reasons to override most of the styles, however you're right that I needed to change the width on mobile. Fixed.

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,230 @@
...
+    max = 100,
+    step = 1,
+    forwardedRef,
+    label,
+    labelId,
+    description,
+    error,
+    disabled,
+    value,
+    onCountChange,
```
chanceaclark, line 36: 🤔What are your thoughts on using this vs `onChange`?

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+      return validateError(error);
+    }, [error]);
+
+    return (
+      <div>
+        {renderedLabel}
+        {renderedDescription}
+        <StyledCounterWrapper disabled={disabled} error={errors} focus={focus}>
+          <StyledCounterButton
```
chanceaclark, line 200: It seems when you are using a keyboard and you press enter on a button it triggers a form submit.

packages/big-design/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,230 @@
...
+  }) => {
+    const [focus, setFocus] = useState(false);
+    const uniqueCounterId = useUniqueId('counter');
+    const id = props.id ? props.id : uniqueCounterId;
+
+    useEffect(() => {
+      if (!Number.isInteger(value)) {
+        onCountChange(Math.round(value));
+      }
+    });
```
chanceaclark, line 47: I think you might want to include a dependency array on this guy.

packages/big-design/src/components/Counter/Counter.tsx
jordan-massingill: Ah yeah, the onClick was on the wrong component. Fixed.

packages/big-design/src/components/Counter/Counter.tsx
jordan-massingill: Using something other than the native `onChange` (in addition to following the style of the `<Select />` component, for example) allows for an `onChange` to be applied directly to the `<input />` component inside the `<Counter />` parent while supplying the change logic to the parent separately. `onCountChange` also has different behavior than a native `onChange` function - it returns a number and not an event.
Screen Shot 2020-04-20 at 1 59 23 PM

PR #378, opened 2020-04-21
fix(component): SHIPPING-1526 adds description to Select & MultiSelect

agchan12:
## What?

Add description prop to Select dropdown

## Why?

We support the description prop, but typescript doesn't recognise it as a prop

## Testing/Proof
Select
Screen Shot 2020-04-24 at 2 50 54 pm

MultiSelect
Screen Shot 2020-04-24 at 2 51 05 pm

ping @deini @animesh1987 @chanceaclark

### DIFF
--- a/packages/big-design/src/components/Select/types.ts
+++ b/packages/big-design/src/components/Select/types.ts
@@ -1,11 +1,12 @@
 import { Placement } from '@popperjs/core';
-import { RefObject } from 'react';
+import React, { RefObject } from 'react';
 
 import { InputProps } from '../Input';
 import { ListItemProps } from '../List/Item';
 
 interface BaseSelect extends Omit<React.HTMLAttributes<HTMLInputElement>, 'children'> {
   action?: SelectAction;
+  description?: React.ReactChild;
   disabled?: boolean;
   error?: InputProps['error'];
   filterable?: boolean;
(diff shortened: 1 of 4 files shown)


packages/big-design/src/components/Select/Select.tsx
```diff
@@ -20,6 +20,7 @@ import { SelectAction, SelectOption, SelectOptionGroup, SelectProps } from './ty
 export const Select = typedMemo(
   <T extends any>({
     action,
+    description,
```
chanceaclark, line 23: You shouldn't need to destructure this, since it gets passed into the `Input` via `rest`.

packages/big-design/src/components/Select/types.ts
```diff
@@ -6,6 +6,7 @@ import { ListItemProps } from '../List/Item';
 
 interface BaseSelect extends Omit<React.HTMLAttributes<HTMLInputElement>, 'children'> {
   action?: SelectAction;
+  description?: string;
```
chanceaclark, line 9: This can actually accept a `string | FormControlDescription`. See our `Input` implements it.

chanceaclark:
Just remembered, we probably want this for `MultiSelect`'s too.

jorgemoya:
Yeah, if we can get this for `MultiSelect` that'd be 💯

PR #391, opened 2020-04-29
fix(component): improve Dropdown & MultiSelect component performance

jorgemoya:
Reduces initial re-renders of components. First steps in fixing performance issues.

### DIFF
--- a/packages/big-design/src/components/List/List.tsx
+++ b/packages/big-design/src/components/List/List.tsx
@@ -1,3 +1,4 @@
+import type { State } from '@popperjs/core';
 import React, { forwardRef, HTMLAttributes, memo, Ref } from 'react';
 
 import { useIsomorphicLayoutEffect, useWindowSize } from '../../hooks';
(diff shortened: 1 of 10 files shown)


packages/big-design/src/components/List/List.tsx
```diff
@@ -18,12 +17,12 @@ const StyleableList: React.FC<ListProps & PrivateProps> = ({ forwardedRef, updat
   const { height, width } = useWindowSize();
 
   useIsomorphicLayoutEffect(() => {
-    async function scheduleUpdate() {
-      await update();
+    function scheduleUpdate() {
+      update();
```
jordan-massingill, line 21: I'm not very familiar with this function - why remove the `async` aspect of it?

jorgemoya: This may actually be incorrect. I have it typed incorrectly as a non `asynchronous` function, which is why I changed it to non async. However, it looks like it does return a `Promise`.

deini: The types were wrong in the previous downshift version, got confused by this a few days ago.

PR #392, opened 2020-04-29
fix(component): prevent select from selecting option when blurring

### LINKED ISSUE #393: Wrong behavior on closing Select component
I have found interesting behavior for the `Select` component.

**To Reproduce**
1. Open dropdown (Click on the input)
2. Hover on another option from the dropdown
3. Click outside of dropdown (exit without changes)

Expected result: Nothing happens
Actual result: An option that was hover was selected

**Code example**
https://codesandbox.io/s/dank-sea-d20ee

**Desktop (please complete the following information, if applicable):**
 - OS: [Mac Os]
 - Browser [chrome]
 - Version [Version 81.0.4044.122 (Official Build) (64-bit)]

jorgemoya:
Bump downshift. 

Set highlightedIndex to `-1` when needed to prevent the option from selecting when blurring out of the component.

Fixes #393

### DIFF
--- a/packages/big-design/src/components/Select/Select.tsx
+++ b/packages/big-design/src/components/Select/Select.tsx
@@ -110,7 +110,7 @@ export const Select = typedMemo(
     const handleOnHighlightedIndexChange = (
       changes: Partial<UseComboboxState<SelectOption<T> | SelectAction | null>>,
     ) => {
-      if (typeof changes.highlightedIndex !== 'undefined' && changes.highlightedIndex > -1) {
+      if (typeof changes.highlightedIndex !== 'undefined') {
         setHighlightedIndex(changes.highlightedIndex);
       }
     };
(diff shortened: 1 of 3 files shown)


packages/big-design/src/components/Select/Select.tsx
```diff
@@ -110,7 +110,7 @@ export const Select = typedMemo(
     const handleOnHighlightedIndexChange = (
       changes: Partial<UseComboboxState<SelectOption<T> | SelectAction | null>>,
     ) => {
-      if (typeof changes.highlightedIndex !== 'undefined' && changes.highlightedIndex > -1) {
+      if (typeof changes.highlightedIndex !== 'undefined') {
```
deini, line 113: Do we also need this fix in Dropdown / Multi?

jorgemoya: Nope, only relevant to Select.

PR #395, opened 2020-04-30
fix: type regression for ts < 3.8

deini:
No longer relying on import/export type only to support ts < 3.8


packages/big-design/src/components/Input/private.ts
```diff
@@ -1,2 +1,4 @@
+import { StyledInputWrapperProps as _StyledInputWrapperProps } from './styled';
+
+export type StyledInputWrapperProps = _StyledInputWrapperProps;
 export { StyledInput, StyledInputWrapper } from './styled';
```
chanceaclark, line 4: 🍹 swap these two lines to be consistent

PR #398, opened 2020-05-07
fix: lock downshift to 5.2.0 to fix behavior of onOptionChange

### LINKED ISSUE #396: Select.onOptionChange works unpredictably
**Describe the bug**
`Select.onOptionChange` prop does not work as expected.

**To Reproduce**
1. https://codesandbox.io/s/lingering-pond-5szw9?file=/src/App.js
2. Select any value from the Select component - **Nothing happens**

**Expected behavior**
Changing the value should trigger the `onOptionChange` event.

**Additional context**
Downgrading to 0.18.0 fixes the problem (can be validated via the left sidebar - changing the version of the package)

jorgemoya:
This will fix the weird behavior of Selects until we can properly figure out why it breaks using the latest versions of Downshift.

The issue is that our `onOptionChange` is being called two times when you click an option, which trigger an infinite loop effect. By expecting a single call we know the code is working as expected.

Tested locally as well.

Fixes #396

### DIFF
--- a/packages/big-design/package.json
+++ b/packages/big-design/package.json
@@ -36,7 +36,7 @@
     "@bigcommerce/big-design-icons": "^0.11.1",
     "@bigcommerce/big-design-theme": "^0.9.0",
     "@popperjs/core": "^2.2.1",
-    "downshift": "^5.2.2",
+    "downshift": "5.2.0",
     "focus-trap": "^5.1.0",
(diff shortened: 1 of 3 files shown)


Borales:
@jorgemoya adding `"downshift": "5.2.2"` to my package.json helped to solve the problem (I was getting 5.2.6 previously)

PR #406, opened 2020-05-22
feat(component): add Tree component

chanceaclark:
# What

Adds the `Tree` component.

## Still left in this PR:
- [x] Improve state data structures (use `Set` and `Map`)
- [x] Rework initial tree generation (performance)
- [x] Update underlying state when async tree mutations happen.
- [x] Subcategory selected indicator.
- [x] Tests
- [x] Documentation

Type-ahead is recommended for accessibility, but for the current scope, I didn't include it.

## Screenshot
screencapture-localhost-3000-tree-2020-06-11-16_50_10


packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,246 @@
...
+  const expandedNodeIds = initializeExpandedNodeIds(nodes);
+  const visibleNodeIds = initializeVisibleNodeIds(nodes, nodeMap);
+
+  return {
+    nodes,
+    nodeMap,
+    expandedNodeIds,
+    flattenedNodeIds: initializeFlattenedNodeIds(nodes),
+    focusedNode: nodes[0].id,
+    selectedValues: [],
```
chanceaclark, line 20: Just caught my attention, I'll actually need to add some logic for these.

valterfatia:
@chanceaclark I think there's something off with the icon sizes/alignments, which weren't perfect on the design either…

Ideally we should have the following:
- All icons/graphics using a 24px canvas
- Icons should align across rows even when indented (for example the expand arrow should be in the middle of the circle)
- The start of text and icon across rows should align too (for example a folder icon should align with the text of the row above)

Unrelated to this I also noticed the sizes of our checkboxes/radio buttons have some inconsistencies (https://github.com/bigcommerce/big-design/issues/407)

eugene-polev:
Screen Shot 2020-05-25 at 11 27 14 AM

In support of what Valter has mentioned.

chanceaclark:
@valterfatia @eugene-polev Thank you guys for these! I'll address it in my next commit.

capsula4:
Is it possible to customize the icon? e.g. channel vs folder

chanceaclark:
> Is it possible to customize the icon? e.g. channel vs folder

@capsula4 Yeah, there is a `icon` prop on a `TreeNode` in which you can override the `FolderIcon`.

packages/big-design/src/components/Tree/types.ts
```diff
@@ -0,0 +1,42 @@
+export interface TreeProps<T> {
+  iconless?: boolean;
+  nodes: TreeNodeProps<T>[];
+  selectable?: 'radio' | 'multi';
+  onExpand?(node: TreeNodeProps<T>): TreeNodeRef<T> | void;
```
deini, line 5: Should this also accept returning a `Promise<TreeNodeRef<T>>`>?

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
```diff
@@ -0,0 +1,303 @@
...
+  const handleNodeToggle = async (e?: React.MouseEvent<Element>) => {
+    dispatch({ type: 'FOCUS', id });
+
+    // Prevents the collapse/expand when clicking on a radio or checkbox
+    // Checks to see if every element inside the selectableRef gets clicked.
+    if (e?.target instanceof Node && selectableRef.current?.contains(e?.target)) {
+      return;
+    }
+
+    if (thisRef.current.children) {
```
deini, line 60: :tropical_drink: to avoid nesting statements, can we return early here? Something like:

```js
if (thisRef.current.children.length === 0) {
  return;
}
```

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
```diff
@@ -0,0 +1,303 @@
...
+      if (!expanded) {
```
deini, line 61: :tropical_drink: Lets swap to `if (expanded)`

packages/big-design/src/components/Tree/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  onCollapse?(node: TreeNodeProps<T>): TreeNodeRef<T> | void;
+  // Based on the selectable prop, it can either
```
deini, line 7: Instaed of this comment, lets make the types act that way. :pray:

packages/big-design/src/components/Tree/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  };
+};
+
+export interface TreeState<T> {
+  nodeMap: NodeMap;
+  expandedNodeIds: TreeNodeId[];
+  flattenedNodeIds: TreeNodeId[];
+  focusedNode: TreeNodeId;
+  selectedValues: Array<T>;
+  visibleNodeIds: TreeNodeId[];
```
deini, line 41: Other than `flattenedNodeIds` I don't think the others need to be arrays. Maybe we can use `Set` here 🤔. This should avoid having to iterate everywhere.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,304 @@
...
+  }, []);
+
+const initializeSelectedValues = <T>(nodes: TreeProps<T>['nodes'], radio: boolean): T[] => {
+  let selectedValues: T[] = [];
+  const queue = [...nodes];
+
+  for (let i = 0; i < queue.length; i++) {
+    const node = queue[i];
+    if (node.children) {
+      // queue = queue.concat(node.children);
```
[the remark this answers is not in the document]
chanceaclark: 🙇

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,304 @@
...
+      queue.splice(i + 1, 0, ...node.children);
```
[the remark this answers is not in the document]
chanceaclark: `nodeMap` doesn't contain any value information in them. Only `id`'s. Maybe I need to make that more clear with the `nodeMap` name.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,304 @@
...
+        return childId;
+      }
+    }
+
+    return acc;
+  }, null);
+
+const getNextVisibleNode = <T>(state: TreeState<T>, id: TreeNodeId): TreeNodeId => {
+  const index = state.flattenedNodeIds.indexOf(id);
+
```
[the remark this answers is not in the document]
chanceaclark: Just refactored these; not relevant anymore.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,304 @@
...
+  if (index === -1 || nextPossibleId === undefined) {
+    return id;
+  }
+
+  return state.visibleNodeIds.includes(nextPossibleId) ? nextPossibleId : getNextVisibleNode(state, nextPossibleId);
+};
+
+const getPreviousVisibleNode = <T>(state: TreeState<T>, id: TreeNodeId): TreeNodeId => {
+  const index = state.flattenedNodeIds.indexOf(id);
+
```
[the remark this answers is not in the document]
chanceaclark: Just refactored these; not relevant anymore.

packages/big-design/src/components/Tree/reducer.ts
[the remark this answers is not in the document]
deini: Good point 🤔 what if we have `nodeMap` be `nodeId: node`? It would be helpful here.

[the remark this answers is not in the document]
chanceaclark: I could just add the `value` to the `nodeMap` 🤷Would definitely be easier and refactoring it to include all the data.

packages/big-design/src/components/Tree/types.ts
chanceaclark: I thought about doing that at the beginning. I dropped the idea, but I'll try it out after resolving the above comments.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,304 @@
...
+// Inserts id in the correct order and returns a new array.
+const insertNode = (array: TreeNodeId[], id: TreeNodeId, flattenedNodeIds: TreeNodeId[]): TreeNodeId[] => {
+  if (array.includes(id)) {
+    return array;
+  }
+
+  const newArr = [...array, id];
+
+  return flattenedNodeIds.filter((nodeId) => newArr.includes(nodeId));
```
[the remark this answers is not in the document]
chanceaclark: `flattenedNodeIds` is in perfect order. In this case I'm returning a new array, in the same order as it would appear in `flattenedNodeIds`.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,304 @@
...
+};
+
+const removeChildNodes = (array: TreeNodeId[], id: TreeNodeId, nodeMap: NodeMap): TreeNodeId[] => {
+  let tempArr = [...array];
+
+  if (nodeMap[id].children.length > 0) {
+    tempArr = nodeMap[id].children.reduce((acc, childId) => {
```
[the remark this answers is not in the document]
chanceaclark: I modified it a bit, let me know that it's kind of what you were thinking.

packages/big-design/src/components/Tree/types.ts
chanceaclark: So after attempting two different methods, this isn't possible without casting types.

Both of the attempts used different strategies; one was blocked by not being able to use `undefined` as an interfaces index type, the other was blocked by the idea of variance in typings ([covariance and contravariance](https://www.stephanboyer.com/post/132/what-are-covariance-and-contravariance)).

Going to leave it as is for now.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  return initialFocusedNode !== null ? initialFocusedNode : nodes[0].id;
+};
+
+const recursiveInitializeFocusNode = <T>(
+  nodes: TreeProps<T>['nodes'],
+  selectedValues: T[],
+  visibleNodeIds: TreeNodeId[],
+): TreeNodeId | null =>
+  nodes.reduce<TreeNodeId | null>((acc, node) => {
```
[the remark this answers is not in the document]
chanceaclark: We kind of do need to keep iterating through to see if we can find a better node to select. This is the criteria for the intially focused node:
>**When a single-select tree receives focus:**
>- If none of the nodes are selected before the tree receives focus, focus is set on the first node.
>- If a node is selected before the tree receives focus, focus is set on the selected node.
>
>**When a multi-select tree receives focus:**
>- If none of the nodes are selected before the tree receives focus, focus is set on the first node.
>- If one or more nodes are selected before the tree receives focus, focus is set on the first selected node.

I went for bottom-up approach here. I could split it up into two steps:
1. Flatten the nodes
2. Find the node.

But either way I'll have to use a `.reduce` function.

packages/docs/pages/Tree/TreePage.tsx
```diff
@@ -0,0 +1,143 @@
...
+          {
+            id: 0,
+            label: 'Category',
+            children: [],
+          },
+          { id: 1, label: 'Category' },
+          { id: 2, label: 'Category' },
+        ];
+
+        const handleOnExpand = ({ children }: TreeNodeProps<undefined>) => {
```
deini, line 128: What do you think about changing this example a bit to showcase a mock async API request with a node id? You could define the "logic" outside of the example so that it doesn't render in the docs and then just use it like... `await fetchChildren(node.id)`.

It should help showcase that you get the full `node` as params and that you can return a promise.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  dfsValue = depthFirstSearch(
+    nodes,
+    (node) => {
+      if (node.selected && !node.disabled) {
+        return true;
+      }
+
+      return false;
+    },
```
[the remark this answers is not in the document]
chanceaclark: That can return undefined if `selected` or `disabled` is undefined.

packages/big-design/src/components/Tree/reducer.ts
```diff
@@ -0,0 +1,276 @@
...
+      return state;
+
+    default:
+      return state;
+  }
+};
+
+const initializeNodeMap = <T>(nodes: TreeProps<T>['nodes'], parent?: TreeNodeId): NodeMap =>
+  nodes.reduce((acc, node) => {
+    const childNodeMap = node.children ? initializeNodeMap(node.children, node.id) : {};
```
davidchin, line 141: I notice that `initializeNodeMap`, `initializeExpandedNodeIds`, `initializeFlattenedNodeIds`, `initializeVisibleNodeIds`, `initializeSelectedValues` and `initializeFocusNode` functions traverse the same tree of `nodes` multiple times. Do you think if we can achieve the same result by traversing the tree only once?

Also I notice that the traversal is not tail-call optimised. It shouldn't be a problem unless we're dealing with a large list. Otherwise, we might have a memory and call-stack problem. So it'd be good to look at either optimising the recursive calls or use a while loop instead.

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
```diff
@@ -0,0 +1,328 @@
...
+      return;
+    }
+
+    let newSelectedValues: T[] = [];
+
+    if (selectable === 'multi') {
+      if (state.selectedValues.includes(value)) {
+        newSelectedValues = state.selectedValues.filter((selectedValue) => selectedValue !== value);
+      } else {
+        newSelectedValues = [...state.selectedValues, value];
```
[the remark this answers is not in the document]
chanceaclark: Pretty much most of the state values can be changed to using a `Set`. Honestly, I probably should've done it initially. I'll go ahead and start convert over since I'm getting into cleanup and performance.

packages/big-design/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,47 @@
...
+          onSelect={onSelect}
+          selectable={selectable}
+          state={state}
+        />
+      )),
+    [iconless, nodes, onCollapse, onExpand, onSelect, selectable, state],
+  );
+
+  return (
+    <StyledUl role="tree" aria-multiselectable={selectable === 'multi'} style={{ overflow: 'hidden' }}>
```
jordan-massingill, line 41: Why have the `overflow: hidden` as an inline style?

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
```diff
@@ -0,0 +1,328 @@
...
+  );
+
+  const renderedChildren = useMemo(
+    () =>
+      thisRef.current.children &&
+      expanded && (
+        <StyledUl role="group">
+          {isLoading && thisRef.current.children.length < 1 ? (
+            <TreeNode
+              id={-1}
```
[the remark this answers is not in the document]
chanceaclark: The `-1` id is only for the loading state node. As you can see I don't pass any of the callback handlers into the `TreeNode` so it should be safe. The only thing I would worry about in this situation is having that id being passed into the underlying state, but luckily it will never do that. This id will also never be passed as an HTML id attribute so we shouldn't need to be concerned about uniqueness.

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
```diff
@@ -0,0 +1,328 @@
...
+          children: thisRef.current.children,
+          disabled,
+          expanded,
+          id,
+          label,
+          selected,
+          value,
+        });
+
+        if (callbackValue) {
```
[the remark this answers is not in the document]
chanceaclark: Yeah we wouldn't want a falsy value anyways. The callback value should be an object containing new children to update.

.github/CODEOWNERS
```diff
@@ -3,4 +3,7 @@
 # Jorge's components:
 packages/big-design/src/components/Dropdown/ @jorgemoya
 packages/big-design/src/components/MultiSelect/ @jorgemoya
-packages/big-design/src/components/Select/ @jorgemoya
\ No newline at end of file
+packages/big-design/src/components/Select/ @jorgemoya
+
+# Chance's components:
+packages/big-design/src/components/Tree/ @chanceaclark
```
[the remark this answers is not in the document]
chanceaclark: Let's chat about this offline.

packages/big-design/src/components/Tree/reducer.ts
chanceaclark: I've been thinking about that since I initially pushed this PR. I'm going to modify the underlying state data structure to use `Set` and `Map` for most of the state. While I'll do that, I'll try to build the state through 1 traversal.

Regarding tail-call optimization, I'll take a look to see what I can do.

packages/big-design/src/components/Tree/Tree.tsx
chanceaclark: Good question! I only need the overflow on that one `ul` so I opted in for this instead of creating a whole new SC for it. I tried to move it into the SC, but there's some weird nesting stuff that causes styles to override the parent styles and they don't need to be complex.

.github/CODEOWNERS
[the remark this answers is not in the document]
davidchin: Cool, yeah let's have a chat about it.

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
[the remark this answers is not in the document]
davidchin: Ah I see, so it's just the loading state node. If that's the case, I think it's fine. Also I just realise `id` is not passed to the `onSelect` callback. I wonder if we should (please see my comment below) 🤔

packages/big-design/src/components/Tree/types.ts
```diff
@@ -0,0 +1,42 @@
+export interface TreeProps<T> {
+  iconless?: boolean;
+  nodes: TreeNodeProps<T>[];
+  selectable?: 'radio' | 'multi';
+  onExpand?(node: TreeNodeProps<T>): Promise<TreeNodeRef<T>> | TreeNodeRef<T> | void;
+  onCollapse?(node: TreeNodeProps<T>): Promise<TreeNodeRef<T>> | TreeNodeRef<T> | void;
+  // Based on the selectable prop, it can either
+  // return and array of values, or just one.
+  onSelect?(values: T[] | T): void;
```
davidchin, line 9: Would it be useful to also return the IDs that are selected? This is because `value` alone cannot sufficiently indicate which items are selected.

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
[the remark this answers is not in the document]
davidchin: Oh I see, I just had another look at the code. So essentially you can swap out elements of a `TreeNode` with another set of elements passed in from the outside. Can you give me an example why we want to do that? Also, after elements are swapped, would they get wiped out if a new or modified list of `nodes` get passed into `Tree`?

packages/big-design/src/components/Tree/types.ts
chanceaclark: In most cases we would map the `id` with the `value`. Could you provide a use-case in which value is not sufficient?

packages/big-design/src/components/Tree/reducer.ts
chanceaclark: After refactoring the state to use `Set` and `Map` I think I resolved this. Take a look and let me know what you think.

chanceaclark:
Pushed a major refactor ♻️ 🙏

packages/big-design/src/components/Tree/reducer.utils.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    if (state.focusedNode === null) {
+      state.focusedNode = node.id;
+    }
+
+    if (node.expanded) {
+      state.expandedNodeIds.add(node.id);
+    }
+
+    const parent = state.nodeMap.get(node.id)?.parent;
```
deini, line 44: 🍹 Can you do 
```js
const parent = getParentId(state.nodeMap, node.id);
```

Then use `parent` when you are setting the `nodeMap` and in the following statements.

packages/big-design/src/components/Tree/types.ts
davidchin: I raised this because `value` is not guaranteed to be unique. As far as I can tell from the doc, only `id` is unique. If it's not unique, then you can't find the corresponding `id` with `value`. And if it's unique, then you don't really need an additional `id` as an identifier.

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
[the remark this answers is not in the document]
davidchin: Ok, in that case, I suggest we should ask the client to pass in a new list of `nodes` if it needs to display additional data. The way I see it is that `nodes` is the data source. If there are changes to the prop, the component should react to those changes. If you want to model `nodes` as the _initial_ data source, then we need to make it clear that the component won't react to any changes to the prop after its initial render. And all updates have to managed through the callbacks by manually constructing the `TreeNode`. At the moment, there are two competing ways of updating the tree (one wipes out the other). And I'd much prefer if we just stick with `nodes` instead of making `TreeNode` as a part of the public API.

packages/big-design/src/components/Tree/reducer.utils.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    if ((parent !== undefined && state.expandedNodeIds.has(parent)) || parent === undefined) {
+      state.visibleNodeIds.push(node.id);
+    }
+
+    if (node.children && node.children.length > 0) {
+      recursiveInitialize(node.children, state, radio);
```
davidchin, line 51: It's nice to see that we only have a single loop instead of the multiple loops we had previously. 👍 Although it's still not tail-call optimised, I think it's sufficient for now because I don't anticipate the tree to be too deep.

packages/big-design/src/components/Tree/reducer.utils.ts
```diff
@@ -0,0 +1,181 @@
...
+  const node = state.nodeMap.get(action.id);
+
+  if (state.expandedNodeIds.has(action.id)) {
+    state.expandedNodeIds.delete(action.id);
+
+    node?.children.forEach((childId) => {
+      const index = state.visibleNodeIds.indexOf(childId);
+
+      if (index > -1) {
+        state.visibleNodeIds.splice(index, 1);
```
davidchin, line 111: `splice` actually mutates the original array. You might want to treat `state` as immutable, because I think `state` refers to a state that's passed in from a `reducer`. I'd return a copy instead.

packages/big-design/src/components/Tree/reducer.utils.ts
```diff
@@ -0,0 +1,181 @@
...
+    state.flattenedNodeIds.push(node.id);
+
+    state.nodeMap.set(node.id, {
+      children: new Set(node.children?.map((child) => child.id)),
+      id: node.id,
+      parent: getParentId(state.nodeMap, node.id),
+    });
+
+    if (state.focusedNode === null) {
+      state.focusedNode = node.id;
```
davidchin, line 37: I think the same comment applies here. I think `state` is an object that gets passed in from a reducer. If we mutate the object, it'd be much harder for the caller to detect changes.

chanceaclark: `focusedNode` is a value, not a reference so I think it would be fine in this case.

packages/big-design/src/components/Tree/reducer.utils.ts
chanceaclark: My only quandary with that is since is that we are using `Set`'s. When we `.add` or `.delete` on a set it's mutating the `Set`, so why should we not do it consistently for our other state objects?

.github/CODEOWNERS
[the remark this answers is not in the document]
chanceaclark: We will resolve this in another PR.

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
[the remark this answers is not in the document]
chanceaclark: After thinking about it a bit more, I think we'll want to recommend people to use the callback method of updating the children. If this ends up being a bad for consumers we can always refactor.

I put a "notice" in the async example to use the callback method rather than updating nodes itself.

packages/big-design/src/components/Tree/reducer.utils.ts
davidchin: Yes, what I meant is, you should treat all objects as immutable (especially when you're using the reducer pattern) - whenever they are passed in as an argument of a method or function. So with `Set`, you should create a new one as well. It is okay to mutate an object if it is created inside the same function / method scope. But once it is passed to another scope, it shouldn't be mutated.

.github/CODEOWNERS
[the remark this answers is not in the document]
chanceaclark: Thanks for catching that, forgot to remove that.

packages/big-design/src/components/Tree/reducer.utils.ts
chanceaclark: I started refactoring it, I'll hopefully have a commit soon for it.

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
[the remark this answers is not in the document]
chanceaclark: So I went ahead and came up with something to continue along the current path (*1.*). I've renamed `nodes` to `initialNodes` and restricted re-rendering through a `usePropOnce` hook.

I do agree that we should change the component for props to passed down declaratively rather than getting set imperatively, so I went ahead and created a SPIKE ticket to explore it further.

My main reasoning behind going down approach 1 is that we already have taken into account a lot of the a11y concerns. If we go the approach where the client has to partly manage that, there's opportunities for some wrong implementations and we want to manage those misses. Not a major factor, but I also don't want to extend the release of this component a couple more weeks because we decided to change implementations. Though, I will be working with the Product team to get the spike into the next sprint or two as I feel it's important to lock it down.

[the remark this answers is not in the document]
davidchin: Ok cool, thank you very much @chanceaclark, I'm happy with what you're proposing 👍

chanceaclark:
Comments should be addressed. Please ♻️ 🙏

PR #408, opened 2020-05-29
feat(component): add Datepicker component

### LINKED ISSUE #255: Feature Request: Date/Time picker component
For scheduling use cases, it's needed to allow a user to select a particular date and/or time that something should happen. There's nothing perfect for this use case with the native input components in BigDesign right now.

A Date and/or Time input component for forms, such as:

https://www.npmjs.com/package/react-moment-datepicker
https://www.npmjs.com/package/react-datetime-picker

... but styled within BigDesign.

animesh1987:
# What
Adds the `Datepicker` component.

## Screenshot
Screen Shot 2020-05-29 at 7 44 28 pm

Fixes #255


chanceaclark:
What version of `yarn` + `npm` are you using? Seems like it's dropping the integrity SHA on the yarn.lock file. Try running `yarn install` again to see if it fixes it, otherwise you might need to bump your `yarn` and/or `npm` versions.

deini:
@animesh1987 Thanks for the contribution, really nice job! Here are a few things I noticed:

- Can add your route to `next.config.js` 🙏. Right now if you go directly to `/datepicker` it will 404.
- I can't use my keyboard to move around the selected day, the library you are using seems to support it, can you please 👀

packages/big-design/package.json
```diff
@@ -64,6 +65,7 @@
     "@types/jest": "^25.2.1",
     "@types/node": "^13.1.8",
     "@types/react": "^16.8.8",
+    "@types/react-datepicker": "^2.11.0",
```
deini, line 68: Since we are using and exporting a type from here, can you move this to regular dependencies 🙏

packages/big-design/src/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,85 @@
...
+  selected?: Date | null;
+  filterDate?(date: Date): boolean;
+  onChange(date: Date): void;
+}
+
+export interface ExtraProps {
+  theme?: ThemeInterface;
+}
+
+export type DatepickerProps = Props;
```
deini, line 30: Can we do `export interface DatepickerProps` and remove this assignment?

packages/big-design/src/components/Datepicker/styled.ts
```diff
@@ -0,0 +1,143 @@
...
+
+import { ExtraProps } from './Datepicker';
+import { HeaderProps } from './Header';
+
+export const StyledHeader = styled.div<HeaderProps>`
+  align-items: center;
+  display: flex;
+  font-weight: ${({ theme }) => theme.typography.fontWeight.semiBold};
+  justify-content: space-between;
+`;
```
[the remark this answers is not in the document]
animesh1987: Done

packages/big-design/src/components/Datepicker/Header.tsx
```diff
@@ -0,0 +1,48 @@
...
+  <StyledHeader date={date}>
+    <Button
+      type="button"
+      iconOnly={<ChevronLeftIcon />}
+      onClick={decreaseMonth}
+      disabled={prevMonthButtonDisabled}
+      variant="subtle"
+    />
+
+    <span>{`${months[date.getMonth()]} ${date.getFullYear()}`}</span>
```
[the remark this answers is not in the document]
animesh1987: Done

packages/docs/PropTables/DatepickerPropsTable.tsx
```diff
@@ -0,0 +1,52 @@
...
+    description: 'Callback called with value of selected date.',
+  },
+  {
+    name: 'label',
+    types: 'string',
+    description: 'Adds a label to the field.',
+  },
+  {
+    name: 'selected',
+    types: 'Date',
```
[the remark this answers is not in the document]
animesh1987: Have updated the API as we discussed offline.

packages/big-design/src/components/Datepicker/styled.ts
```diff
@@ -0,0 +1,143 @@
...
+  .react-datepicker__day {
+    color: ${({ theme }) => theme.colors.secondary60};
+    display: inline-block;
+    line-height: ${({ theme }) => theme.lineHeight.xLarge};
+    margin: ${({ theme }) => theme.spacing.xxSmall};
+    text-align: center;
+    width: ${({ theme }) => theme.lineHeight.xLarge};
+  }
+
+  & .react-datepicker__day:hover {
```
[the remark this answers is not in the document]
animesh1987: The one state that we discussed offline, it's because when the calendar is opened the date is selected by default when we enable keyboard navigation. I might need to check this with design to get some resolution on this.

packages/docs/PropTables/DatepickerPropsTable.tsx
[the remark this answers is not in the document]
chanceaclark: Let's simplify it a bit more. So for your interface we can do something like this:
```ts
interface Props {
  dateFormat?: string;
  error?: React.ReactNode;
  label?: string;
  locale?: string;
  filterDate?(date: Date): boolean;
  onDateChange(date: string): void;
}

export type DatepickerProps = Props & React.InputHTMLAttributes<HTMLInputElement>;
```

You'll notice that some of your defined props are missing, but we can actually use the inherit props being pulled in from `InputHTMLAttributes`. We can then destructure those props and map them into the abstracted interfaces for `react-datepicker`.

Here's the map I had in mind:
```
// Native prop - Mapped prop
ariaLabel // // Should be brought in from InputHTMLAttributes
min - minDate
max - maxDate
placeholder - placeholder // Brought in from InputHTMLAttributes
required - required // Brought in from InputHTMLAttributes
value - selected
```

We might need to destructure some props from `InputHTMLAttributes` but I'll let you figure that out 😉

packages/big-design/src/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,91 @@
+import { enAU, enCA, enGB, enUS } from 'date-fns/locale';
+import React, { forwardRef, memo, Ref, useEffect, useState } from 'react';
+import ReactDatePicker, { registerLocale } from 'react-datepicker';
+
+registerLocale('en-US', enUS);
```
chanceaclark, line 5: After thinking about this a bit, instead of manually setting these inside BigDesign, what do you think about exposing an API to register them outside of the component?

My reasoning behind this is so that we can drop the `date-fns` package - one less dependency. Plus I think it should be the application that determines the locales it needs, not BigDesign.

packages/big-design/src/components/Datepicker/Header.tsx
```diff
@@ -0,0 +1,48 @@
...
+  date = new Date(),
+  decreaseMonth,
+  increaseMonth,
+  prevMonthButtonDisabled,
+  nextMonthButtonDisabled,
+}) => (
+  <Flex alignItems="center" justifyContent="space-between">
+    <Button
+      type="button"
+      iconOnly={<ChevronLeftIcon />}
```
chanceaclark, line 30: This should have a `title` as well. Also the title should reflect the action you are taking. So like `View previous month.` (Notice the `.` at the end, this helps create a pause for screen readers).

chanceaclark: Doesn't look like this was resolved. `title` should be on the Icon as a prop.

packages/big-design/src/components/Datepicker/Header.tsx
```diff
@@ -0,0 +1,48 @@
...
+      onClick={decreaseMonth}
+      disabled={prevMonthButtonDisabled}
+      variant="subtle"
+    />
+
+    <Text as="span" marginBottom="none" bold>{`${months[date.getMonth()]} ${date.getFullYear()}`}</Text>
+
+    <Button
+      title="View next month."
```
chanceaclark, line 38: Same. `title` should be a prop on the icon itself.

packages/big-design/src/components/Datepicker/styled.ts
```diff
@@ -0,0 +1,132 @@
...
+    color: ${({ theme }) => theme.colors.secondary70};
+    outline: none;
+  }
+
+  & .react-datepicker__day--today::after {
+    content: '';
+    display: block;
+    position: relative;
+    height: 2px;
+    background: ${({ theme }) => theme.colors.primary};
```
chanceaclark, line 89: 🍹 🔤

packages/big-design/src/components/Datepicker/styled.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  & .react-datepicker__day--outside-month {
+    opacity: 0;
+    visibility: hidden;
+  }
+
+  & .bc-datepicker {
+    background: ${({ theme }) => theme.colors.white};
+    ${({ theme }) => theme.shadow.floating};
```
chanceaclark, line 120: 🍹 we usually throw the shadow first:
```suggestion
    ${({ theme }) => theme.shadow.floating};

    background: ${({ theme }) => theme.colors.white};
```

packages/docs/pages/Datepicker/DatepickerPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+              />
+            </FormGroup>
+          </Form>
+        );
+      }}
+      {/* jsx-to-string:end */}
+    </CodePreview>
+
+    <H1>API</H1>
+
```
chanceaclark, line 38: We should also add this:
```suggestion
    <H1>API</H1>
    
    <Text>
      Supports all native <Code>&lt;input [type="date"] /&gt;</Code> element attributes.
    </Text>

```

packages/big-design/src/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { forwardRef, memo, Ref, useEffect, useState } from 'react';
+import ReactDatePicker, { registerLocale } from 'react-datepicker';
+
+import { Input } from '../Input';
+
+import Header from './Header';
+import { createLocalizationProvider } from './localizationProvider';
+import { StyledDatepicker } from './styled';
+
+export interface Props {
```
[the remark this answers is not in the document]
animesh1987: We export the `DatepickerProps` below in the file.

packages/big-design/src/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,90 @@
...
+  onDateChange(date: string): void;
+}
+
+export interface PrivateProps {
+  forwardedRef: Ref<ReactDatePicker>;
+}
+
+export type DatepickerProps = Props & React.InputHTMLAttributes<HTMLInputElement>;
+
+const RawDatePicker: React.FC<DatepickerProps & PrivateProps> = ({
```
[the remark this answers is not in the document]
animesh1987: Good catch. Should be `Datepicker`.

packages/big-design/src/components/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,90 @@
...
+  required,
+  placeholder,
+  value,
+  ...props
+}) => {
+  const localization = createLocalizationProvider(locale);
+  registerLocale(locale, localization);
+
+  const [selected, setSelected] = useState<Date>();
+  const updateDate: (value: Date) => void = (value) => onDateChange(value.toISOString());
```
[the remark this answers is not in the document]
animesh1987: This value won't be undefined since it comes from when you click on a date.

packages/big-design/src/components/Datepicker/Datepicker.tsx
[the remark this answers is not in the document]
chanceaclark: We should only export what we need to export to the library. It seems like we don't actually need to export this, only `DatepickerProps`, which should also get exported in the `index.ts` file as well.

packages/big-design/src/components/Datepicker/Header.tsx
```diff
@@ -0,0 +1,48 @@
+import { ChevronLeftIcon, ChevronRightIcon } from '@bigcommerce/big-design-icons';
+import React from 'react';
+
+import { Flex } from '..';
+import { Button } from '../Button';
+import { Text } from '../Typography';
+
+export interface HeaderProps {
```
chanceaclark, line 8: I also don't see a need to export this too.

animesh1987: Fixed that.

animesh1987:
> Amazing job man! Thank you a ton!

Thank you for all the helpful comments 🤝

valterfatia:
Awesome work @animesh1987!!! Thank you everyone who helped getting it across the line ;)

PR #410, opened 2020-06-16
chore: set frontend team as codeowners

chanceaclark:
## What

Sets the @bigcommerce/frontend team as CODEOWNERS for `big-design`.

## Why

An ongoing effort to get BigCommerce employees insight and more involved with `big-design`. Also, the previous method of having SMEs as CODEOWNERS can restrict merging if someone is on PTO.


PascalZajac:
🙇

PR #414, opened 2020-06-26
fix(docs): grammar update

jdschmitt:
Proposed fix for grammar issue in `external` description on [Link props page](https://developer.bigcommerce.com/big-design/link).


jdschmitt:
ah, yeah, it was probably originally intended to be `an`. I wasn't sure what was intended. I'll update to that.

chanceaclark:
Before you merge, can you set the commit to `fix(docs): ...`?

PR #417, opened 2020-07-14
feat(component): add fadeAway option to alertsManager

deini:
Auto-dismiss an alert using AlertManager after 5s

Eg usage:
```js
alertsManager.add({ title: 'Hello', fadeAway: true });
```


capsula4:
I wonder if `autoDismiss: number` would have been more self-explanatory and flexible.

deini:
A couple of us chatted a bit about this and the idea is to not make the time flexible for consistency. 

I like the prop name tho, feel feee to open a PR to change it.

PR #423, opened 2020-07-28
feat(component): add description to select option and select action

golcinho:
## What?
Adding a description to the `select option` and `select action`.

## Screenshots
Screenshot 2020-07-28 at 11 15 05
Screenshot 2020-07-30 at 09 45 23
Screenshot 2020-07-30 at 09 45 38
Screenshot 2020-07-28 at 09 38 44
Screenshot 2020-07-28 at 09 39 01

## Testing
Unit tests.

### DIFF
--- a/packages/big-design/src/components/Select/Select.tsx
+++ b/packages/big-design/src/components/Select/Select.tsx
@@ -494,8 +494,12 @@ const iconColor = <T extends any>(item: SelectOption<T> | SelectAction, isHighli
 };
 
 const descriptionColor = <T extends any>(item: SelectOption<T> | SelectAction, isHighlighted: boolean) => {
-  if (!isHighlighted || !('onActionClick' in item)) {
+  if (item.disabled) {
     return 'secondary40';
+  }
+
+  if (!isHighlighted || !('onActionClick' in item)) {
+    return 'secondary60';
   }
 
   return 'actionType' in item ? (item.actionType === 'destructive' ? 'danger30' : 'primary30') : 'primary30';
--- a/packages/big-design/src/components/Select/Select.tsx
+++ b/packages/big-design/src/components/Select/Select.tsx
@@ -24,6 +24,7 @@ import { Input } from '../Input';
 import { List } from '../List';
 import { ListGroupHeader } from '../List/GroupHeader';
 import { ListItem } from '../List/Item';
+import { Small } from '../Typography';
 
 import { DropdownButton, StyledDropdownIcon, StyledInputContainer } from './styled';
 import { SelectAction, SelectOption, SelectOptionGroup, SelectProps } from './types';
@@ -317,8 +318,8 @@ export const Select = typedMemo(
     );
 
     const renderOptions = useCallback(
-      (items: Array<SelectOption<T>>) => {
-        return items.map((item) => {
+      (items: Array<SelectOption<T>>) =>
+        items.map((item) => {
           if (
             !selectOptions.find(
               (option: SelectOption<T> | SelectAction) => 'value' in option && option.value === item.value,
@@ -351,8 +352,7 @@ export const Select = typedMemo(
               {isSelected && <CheckIcon color="primary" size="large" />}
             </ListItem>
           );
-        });
-      },
+        }),
       [getItemProps, highlightedIndex, selectedOption, selectOptions],
     );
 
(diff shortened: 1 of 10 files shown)


valterfatia:
Nice work! Two small things from me:
1. The description text should be in "secondary60" and "small text" – or was there a different direction from another designer?
2. I'm unsure about changing the colour of the description when there's an interaction – currently reviewing within the design team, hoping to have it decided soon ;)

golcinho:
> Nice work! Two small things from me:
> 
> 1. The description text should be in "secondary60" and "small text" – or was there a different direction from another designer?
> 2. I'm unsure about changing the colour of the description when there's an interaction – currently reviewing within the design team, hoping to have it decided soon ;)

Hey @valterfatia !

Thank you for the comments!

I changed color to `secondary60` and description text is `Small` (14px).

Thank you for your assistance with the second point 🤝

golcinho:
> Everything looks good, but do you think we can add these changes to the `MultiSelect` component as well?
> 
> Thanks for adding this! 👍

Good point, thank you 🙇 

I will make it for `MultiSelect` too 👍

packages/docs/pages/Select/SelectPage.tsx
```diff
@@ -342,6 +342,53 @@ const SelectPage = () => (
       </Form>
       {/* jsx-to-string:end */}
     </CodePreview>
+
+    <H2>Option & Action Description</H2>
+
+    <Text>It is possible to add a description for select options and select action</Text>
```
chanceaclark, line 348:
```suggestion
    <Text>It is possible to add a description for select options and actions</Text>
```

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -465,3 +480,15 @@ const iconColor = <T extends any>(item: SelectOption<T> | SelectAction, isHighli
...
+const descriptionColor = <T extends any>(item: SelectOption<T> | SelectAction, isHighlighted: boolean) => {
+  if (item.disabled) {
+    return 'secondary40';
+  }
+
+  if (!isHighlighted || !('onActionClick' in item)) {
+    return 'secondary60';
+  }
+
+  return 'actionType' in item ? (item.actionType === 'destructive' ? 'danger30' : 'primary30') : 'primary30';
```
chanceaclark, line 493: @valterfatia `danger30` color with `danger10` background-color doesn't meet accessibility color contrast ratios. Is there another option?

valterfatia:
Clarifying the last point, from my previous comment lets keep the text as secondary60 on hover and just do the background change as we already have.

packages/big-design/src/components/Select/Select.tsx
valterfatia: I have been reviewing that change within the design team and we have agreed to keep secondary60 as the text colour to meet accessibility requirements.

valterfatia:
Okay I have one more request – which is likely better addressed in a separate PR/issue – but I wanted to call it out here.

All text in a dropdown should remain left aligned to a single "key line" regardless of the presence of icons within the dropdown. In practice this when an item in a dropdown has an icon, then other elements without an icon should keep the same gap so the text aligns in a single line.

Ideally we should aim for having dropdowns items should all have a consistent look and feel, but we also accept in some instances this won't be possible. In those cases our request is for the space to be left empty.

chanceaclark:
> Okay I have one more request – which is likely better addressed in a separate PR/issue – but I wanted to call it out here.
> 
> All text in a dropdown should remain left aligned to a single "key line" regardless of the presence of icons within the dropdown. In practice this when an item in a dropdown has an icon, then other elements without an icon should keep the same gap so the text aligns in a single line.
> 
> Ideally we should aim for having dropdowns items should all have a consistent look and feel, but we also accept in some instances this won't be possible. In those cases our request is for the space to be left empty.

Let's talk about this offline in our Feedback Session.

valterfatia:
> Let's talk about this offline in our Feedback Session.

Sounds good!

golcinho:
@chanceaclark @valterfatia I changed description color to `secondary60` for the select options and actions. (screenshots are updated).

Also, I added a description of the options and actions for the `MultiSelect` component.

cc @jorgemoya

packages/docs/pages/MultiSelect/MultiSelectPage.tsx
```diff
@@ -300,6 +300,53 @@ const MultiSelectPage = () => (
       </Form>
       {/* jsx-to-string:end */}
     </CodePreview>
+
+    <H2>Option & Action Description</H2>
+
+    <Text>It is possible to add a description for select options and select action</Text>
```
chanceaclark, line 306:
```suggestion
    <Text>It is possible to add a description for select options and actions.</Text>
```

packages/docs/pages/Select/SelectPage.tsx
```diff
@@ -342,6 +342,53 @@ const SelectPage = () => (
       </Form>
       {/* jsx-to-string:end */}
     </CodePreview>
+
+    <H2>Option & Action Description</H2>
+
+    <Text>It is possible to add a description for select options and actions</Text>
```
chanceaclark, line 348:
```suggestion
    <Text>It is possible to add a description for select options and actions.</Text>
```

PR #425, opened 2020-07-29
feat(component): add Timepicker component

animesh1987:
# What
Adds the `Timepicker` component.

## Screenshot
Screen Shot 2020-07-29 at 6 01 35 pm

ping @bigcommerce/frontend @chanceaclark

### DIFF
--- a/packages/big-design/src/components/Select/types.ts
+++ b/packages/big-design/src/components/Select/types.ts
@@ -6,6 +6,7 @@ import { ListItemProps } from '../List/Item';
 
 interface BaseSelect extends Omit<React.HTMLAttributes<HTMLInputElement>, 'children'> {
   action?: SelectAction;
+  autoWidth?: boolean;
   description?: React.ReactChild;
   disabled?: boolean;
   error?: InputProps['error'];
--- a/packages/big-design/src/components/Timepicker/Timepicker.tsx
+++ b/packages/big-design/src/components/Timepicker/Timepicker.tsx
@@ -2,6 +2,7 @@ import React, { forwardRef, memo, Ref, useMemo } from 'react';
 
 import { createLocalizationProvider, getTimeIntervals } from '../../utils';
 import { Select } from '../Select';
+
 interface Props {
   error?: React.ReactNode;
   label?: React.ReactChild;
--- a/packages/big-design/src/components/Timepicker/Timepicker.tsx
+++ b/packages/big-design/src/components/Timepicker/Timepicker.tsx
@@ -35,6 +35,7 @@ const RawTimePicker: React.FC<TimepickerProps & PrivateProps> = ({
 }) => {
   return (
     <Select
+      autoWidth={true}
       label={label}
       error={error}
       value={value}
--- a/packages/big-design/src/components/Timepicker/Timepicker.tsx
+++ b/packages/big-design/src/components/Timepicker/Timepicker.tsx
@@ -43,5 +43,7 @@ const RawTimePicker: React.FC<TimepickerProps & PrivateProps> = ({
 };
 
 export const Timepicker = memo(
-  forwardRef<HTMLInputElement, TimepickerProps>((props, ref) => <RawTimePicker {...props} forwardedRef={ref} />),
+  forwardRef<HTMLInputElement, TimepickerProps>(({ className, style, ...props }, ref) => (
+    <RawTimePicker {...props} forwardedRef={ref} />
+  )),
 );
--- a/packages/big-design/src/components/Timepicker/Timepicker.tsx
+++ b/packages/big-design/src/components/Timepicker/Timepicker.tsx
@@ -4,7 +4,7 @@ import { getTimeIntervals } from '../../utils';
 import { Select } from '../Select';
 interface Props {
   error?: React.ReactNode;
-  label?: string;
+  label?: React.ReactChild;
   locale?: string;
   onTimeChange(date: string): void;
 }
--- a/packages/big-design/src/components/Timepicker/Timepicker.tsx
+++ b/packages/big-design/src/components/Timepicker/Timepicker.tsx
@@ -25,7 +25,14 @@ function getTimeIntervals() {
   return times.map((time) => ({ value: time, content: time }));
 }
 
-const RawTimePicker: React.FC<TimepickerProps & PrivateProps> = ({ error, label, onTimeChange, value, ...props }) => {
+const RawTimePicker: React.FC<TimepickerProps & PrivateProps> = ({
+  error,
+  forwardedRef,
+  label,
+  onTimeChange,
+  value,
+  ...props
+}) => {
   return (
     <Select
       label={label}
@@ -34,6 +41,7 @@ const RawTimePicker: React.FC<TimepickerProps & PrivateProps> = ({ error, label,
       onOptionChange={onTimeChange}
       placeholder="hh : mm"
       options={getTimeIntervals()}
+      inputRef={forwardedRef}
       {...props}
     />
   );
--- a/packages/big-design/src/components/Timepicker/Timepicker.tsx
+++ b/packages/big-design/src/components/Timepicker/Timepicker.tsx
@@ -1,10 +1,11 @@
 import React, { forwardRef, memo, Ref } from 'react';
 
+import { createLocalizationProvider } from '../../utils';
 import { Select } from '../Select';
-
 interface Props {
   error?: React.ReactNode;
   label?: string;
+  locale?: string;
   onTimeChange(date: string): void;
 }
 
@@ -14,21 +15,41 @@ interface PrivateProps {
 
 export type TimepickerProps = Props & React.InputHTMLAttributes<HTMLInputElement>;
 
-function getTimeIntervals() {
-  const times = ['12:00 AM'];
+function getTimeIntervals24hr() {
+  const times = ['00:00'];
 
   for (let i = 1; i < 24; i++) {
-    i < 12 ? times.push(`${i}:00 AM`) : times.push(`${i === 12 ? i : i - 12}:00 PM`);
+    times.push(`${i}:00`);
   }
-  times.push('11:59 PM');
+  times.push('23:59');
 
   return times.map((time) => ({ value: time, content: time }));
+}
+
+const defaultTimeIntervals = getTimeIntervals24hr();
+
+function getTimeIntervals(locale: string) {
+  const localization = createLocalizationProvider(locale);
+  const localizedTimeIntervals = defaultTimeIntervals.map((time) => {
+    const baseDate = new Date();
+    const [hour, minute] = time.value.split(':');
+    baseDate.setHours(parseInt(hour, 10));
+    baseDate.setMinutes(parseInt(minute, 0));
+
+    return {
+      content: localization.formatTime(baseDate),
+      value: time.value,
+    };
+  });
+
+  return localizedTimeIntervals;
 }
 
 const RawTimePicker: React.FC<TimepickerProps & PrivateProps> = ({
   error,
   forwardedRef,
   label,
+  locale = 'en-US',
   onTimeChange,
   value,
   ...props
@@ -41,7 +62,7 @@ const RawTimePicker: React.FC<TimepickerProps & PrivateProps> = ({
       value={value}
       onOptionChange={onTimeChange}
       placeholder="hh : mm"
-      options={getTimeIntervals()}
+      options={getTimeIntervals(locale)}
       inputRef={forwardedRef}
       {...props}
     />
--- a/packages/big-design/src/components/Timepicker/Timepicker.tsx
+++ b/packages/big-design/src/components/Timepicker/Timepicker.tsx
@@ -1,6 +1,6 @@
 import React, { forwardRef, memo, Ref } from 'react';
 
-import { createLocalizationProvider } from '../../utils';
+import { getTimeIntervals } from '../../utils';
 import { Select } from '../Select';
 interface Props {
   error?: React.ReactNode;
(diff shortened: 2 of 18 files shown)


valterfatia:
Can we use a smaller size for the input? Also is there a min-size on the width of the dropdown?

animesh1987:
I doubt there is a way to specify the input size. Happy to be corrected in this case.

valterfatia:
> I doubt there is a way to specify the input size. Happy to be corrected in this case.

You're correct… we placed the props for sizing within the form control, combined with the form group :|

animesh1987:
Yes, and from what I have observed once we have the Datetimepicker combined into one then we will have the design outcome as per the figma file.

packages/big-design/src/components/Timepicker/Timepicker.tsx
```diff
@@ -0,0 +1,52 @@
...
+  forwardedRef: Ref<HTMLInputElement>;
+}
+
+export type TimepickerProps = Props & React.InputHTMLAttributes<HTMLInputElement>;
+
+function getTimeIntervals() {
+  const times = ['12:00 AM'];
+
+  for (let i = 1; i < 24; i++) {
+    i < 12 ? times.push(`${i}:00 AM`) : times.push(`${i - 12 === 0 ? i : i - 12}:00 PM`);
```
jordan-massingill, line 21: 🍹 just a nit, but why not `{i === 12 ? i : i-12}` instead of `i - 12 === 0`?

animesh1987: Thanks for the feedback. Now I look at it, not sure what it was done in the first place. Updated with the suggestion. Thanks.

packages/big-design/src/components/Timepicker/Timepicker.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <Select
+      label={label}
+      error={error}
+      value={value}
+      onOptionChange={onTimeChange}
+      placeholder="hh : mm"
+      options={getTimeIntervals()}
+      inputRef={forwardedRef}
+      {...props}
+    />
```
chanceaclark, line 46: 🤔 We might want a `StyleableSelect` component (for internal consumption only) if we want to get smaller `Select` width styles. @deini thoughts on this, or should a normal `Select` be fine?

packages/big-design/src/components/Timepicker/Timepicker.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onTimeChange(date: string): void;
+}
+
+interface PrivateProps {
+  forwardedRef: Ref<HTMLInputElement>;
+}
+
+export type TimepickerProps = Props & React.InputHTMLAttributes<HTMLInputElement>;
+
+function getTimeIntervals() {
```
chanceaclark, line 17: Other countries use 24hr time, what are your thoughts on making this (and I say this loosely) "locale aware"? We will down the line use some sort of locale service that gives us how to display customer information (date, time, currency, ltr/rtl, etc...), so how can we give it the option to use 24hr vs. 12hr times?

packages/big-design/src/components/Timepicker/Timepicker.tsx
deini: I'm not opposed to an internal `StyleableSelect`, however, is the size of the select an issue? Quoting @animesh1987:
>From what I have observed once we have the Datetimepicker combined into one then we will have the design outcome as per the figma file.

Seems like we are able to do what we want without any changes. It can also be wrapped in a container that controls the width. 🤔

chanceaclark: That's in regards to the width of the field. I should've clarified, my comment is in regards to the width of the dropdown:
Screen Shot 2020-07-30 at 15 38 47

deini: Gotcha, yeah a styleable internal one or adding a prop for it to be driven by content.

chanceaclark: We might want to consider that @animesh1987

jorgemoya: A prop driven by content is a better idea imo.

packages/big-design/src/components/Timepicker/Timepicker.tsx
jorgemoya: @chanceaclark brings a good point, we should consider this.

animesh1987: I did look at it quickly, based on locales and on using `https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat`. We get time back in AM/PM format by default unless we specify `hour12: false` in options. And that doesn't seem to be locale dependent.

packages/big-design/src/components/Timepicker/Timepicker.tsx
animesh1987: Thanks for the feedback.

So to summarise it we are fine with adding a prop to existing select component?

chanceaclark: Maybe `useContentWidth` or `autoWidth`. We've been discussing a similar prop for buttons and we should be uniform on this matter. My card is in for `autoWidth`, but what does everyone else think?

Another thing to take not of is that `fit-content` doesn't have support in a lot of browsers, so I'd rather use `auto` which should work in all.

Otherwise that looks good.

packages/big-design/src/components/Timepicker/Timepicker.tsx
chanceaclark: >And that doesn't seem to be locale dependent.

It is locale dependent. Take a look at which countries use 24hr vs. 12hr. https://en.wikipedia.org/wiki/Date_and_time_representation_by_country

Using the `.format` with France as the country will output it in 24hr time:
```javascript
let d = new Date("2019-01-01T00:00:00.000000Z");
let formattedDate = Intl.DateTimeFormat('fr', {
  year: 'numeric',
  month: 'numeric',
  day: 'numeric',
  hour: 'numeric',
  minute: 'numeric',
  second: 'numeric'
}).format(d);

formattedDate // "31/12/2018 à 18:00:00"
```

animesh1987: I think I didn't explain my concern well, I did take a look at the above. Let me try and figure out a way to get the time values as per locale.

packages/big-design/src/components/Timepicker/Timepicker.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { forwardRef, memo, Ref } from 'react';
+
+import { getTimeIntervals } from '../../utils';
+import { Select } from '../Select';
+interface Props {
+  error?: React.ReactNode;
+  label?: string;
```
chanceaclark, line 7: This needs to accept both a `string` and a `FormControlLabel`. I would also make a note somewhere that the `Select` component being used provides the type guard functionality since it doesn't make sense to check it twice.

packages/docs/pages/Datepicker/DatepickerPage.tsx
```diff
@@ -83,6 +81,25 @@ const DatepickerPage = () => (
...
+            <FormGroup>
+              <Timepicker locale="en-US" value={time} onTimeChange={(value) => setTime(value)} />
+            </FormGroup>
+          </Form>
+        );
+      }}
+      {/* jsx-to-string:end */}
+    </CodePreview>
+
+    <TimepickerPropTable />
```
chanceaclark, line 102: I would move this into its own page.

packages/docs/PropTables/TimepickerPropTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+const timepickerProps: Prop[] = [
+  {
+    name: 'onTimeChange',
+    types: '(value: string) => void',
+    required: true,
+    description: 'Callback called with value of selected time.',
+  },
+  {
+    name: 'label',
+    types: 'string',
```
chanceaclark, line 15: Update the types from comment above.

packages/big-design/src/utils/localizationProvider/localizationProvider.ts
```diff
@@ -21,5 +39,23 @@ export const createLocalizationProvider = (locale: string) => {
     monthsLong,
     // Required by datepicker.
     formatLong: {},
+    formatTime: timeFormatter.format,
   };
 };
+
+export function getTimeIntervals(locale: string) {
+  const localization = createLocalizationProvider(locale);
```
chanceaclark, line 47: I wonder if we want to create the localizationProvider inside the component like we do with `DatePicker`? 

Down the line we eventually would want to be able to pass a localization provider into the components so that we are not duplicating efforts.

packages/big-design/src/components/Select/types.ts
```diff
@@ -6,6 +6,7 @@ import { ListItemProps } from '../List/Item';
 
 interface BaseSelect extends Omit<React.HTMLAttributes<HTMLInputElement>, 'children'> {
   action?: SelectAction;
+  autoWidth?: boolean;
```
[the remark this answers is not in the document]
jorgemoya: ☝️

packages/big-design/src/components/Timepicker/Timepicker.tsx
jorgemoya: Good catch. The `Select` type was changed recently.

packages/docs/pages/Datepicker/DatepickerPage.tsx
animesh1987: I was trying to keep the whole Datepicker/Timepicker/Datetimepicker in one page.

packages/big-design/src/utils/localizationProvider/localizationProvider.ts
animesh1987: Sounds good. But I would defer this for now till we tackle `DateTimePicker`, which does combine the two components. As part of that work we can work out how to abstract localization provider.

packages/big-design/src/components/Timepicker/Timepicker.tsx
animesh1987: Where do you think is a good place to make that note. TimepickerPropTable?

packages/big-design/src/utils/localizationProvider/localizationProvider.ts
chanceaclark: I'd like us to follow suite with what we are doing with `DatePicker` so that it'll be a uniform refactor when we tackle `DateTimePicker`.

packages/docs/pages/Timepicker/TimepickerPage.tsx
```diff
@@ -0,0 +1,30 @@
+import { Form, FormGroup, H0, Timepicker } from '@bigcommerce/big-design';
+import React, { useState } from 'react';
+
+import { CodePreview } from '../../components';
+import { TimepickerPropTable } from '../../PropTables';
+
+const TimepickerPage = () => (
+  <>
+    <H0>Timepicker</H0>
+    <CodePreview>
```
chanceaclark, line 10: Two things here:
1. I would add some sort of fluff text here so you can get a quick synopsis of what the component is intended for.
2. 🍹 nit - add spacing.
```suggestion
    <H0>Timepicker</H0>

    <CodePreview>
```

packages/docs/pages/Datepicker/DatepickerPage.tsx
```diff
@@ -8,8 +8,6 @@ const DatepickerPage = () => (
   <>
     <H0>Datepicker</H0>
 
-    <Text>Datepicker is good</Text>
```
chanceaclark, line 11: Same comment about fluff text here.

packages/big-design/src/components/Select/types.ts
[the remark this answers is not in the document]
chanceaclark: Just adding comment here, but make sure to update the documentations prop tables for `Select` and `MultiSelect` as well.

packages/big-design/src/components/Timepicker/Timepicker.tsx
```diff
@@ -0,0 +1,44 @@
...
+      placeholder="hh : mm"
+      options={getTimeIntervals(locale)}
+      inputRef={forwardedRef}
+      {...props}
+    />
+  );
+};
+
+export const Timepicker = memo(
+  forwardRef<HTMLInputElement, TimepickerProps>((props, ref) => <RawTimePicker {...props} forwardedRef={ref} />),
```
chanceaclark, line 43: Also, one quick thing - even though the `Select` will remove `className` and `style` props, I would just get in a good practice and destructure them anyways.
```suggestion
  forwardRef<HTMLInputElement, TimepickerProps>(({ className, style, ...props }, ref) => <RawTimePicker {...props} forwardedRef={ref} />),
```

packages/big-design/src/components/Timepicker/Timepicker.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { forwardRef, memo, Ref, useMemo } from 'react';
+
+import { createLocalizationProvider, getTimeIntervals } from '../../utils';
+import { Select } from '../Select';
+interface Props {
+  error?: React.ReactNode;
+  label?: React.ReactChild;
```
chanceaclark, line 7: Can we add some tests around passing these values into the component?

packages/big-design/src/utils/localizationProvider/localizationProvider.ts
```diff
@@ -1,3 +1,27 @@
...
+interface LocalizationProviderInterface {
+  code?: string;
+  localize?: {
+    month(n: number): string;
+    day(n: number): string;
+  };
+  monthsLong?: string[];
+  formatLong?: {};
+  formatTime(date: Date): string;
+}
```
chanceaclark, line 21: 🍹 nit. Can we have our interfaces/types toward the top of the file?

packages/docs/pages/Timepicker/TimepickerPage.tsx
```diff
@@ -0,0 +1,33 @@
...
+import React, { useState } from 'react';
+
+import { CodePreview } from '../../components';
+import { TimepickerPropTable } from '../../PropTables';
+
+const TimepickerPage = () => (
+  <>
+    <H0>Timepicker</H0>
+
+    <Text>Timepicker is used to selected a time in specific hours and minutes.</Text>
```
chanceaclark, line 11:
```suggestion
    <Text>Timepicker is used to select a time in specific hours and minutes.</Text>
```

packages/big-design/src/components/Timepicker/Timepicker.tsx
animesh1987: Done.

PR #443, opened 2020-08-25
fix(component): allow Dropdown groups to have optional `label`

chanceaclark:
Allows `DropdownItemGroup`'s to have optional `label` key/vals.

#### Without `label`:
Screen Shot 2020-08-25 at 14 55 06

### DIFF
--- a/packages/big-design/src/components/Dropdown/spec.tsx
+++ b/packages/big-design/src/components/Dropdown/spec.tsx
@@ -505,3 +505,26 @@ test('clicking label does not call onItemClick', async () => {
 
   await waitForElement(() => screen.getByRole('option', { name: /option 1/i }));
 });
+
+test('renders appropriate amount of list items', async () => {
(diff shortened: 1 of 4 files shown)


packages/big-design/src/components/Dropdown/spec.tsx
```diff
@@ -505,3 +505,29 @@ test('clicking label does not call onItemClick', async () => {
...
+        {
+          items: [{ content: 'Option 2', onItemClick }],
+        },
+      ]}
+      toggle={<Button>Button</Button>}
+    />,
+  );
+  const toggle = getByRole('button');
+
+  await act(async () => {
```
deini, line 526: You shouldn't need to `act` here or `await` for fireEvent.

Maybe just wait for an item like:
```js
await waitForElement(() => screen.getByRole('option', { name: /option 1/i }));
```

PR #445, opened 2020-08-26
feat(component): add separator to dropdown item groups

bc-williamkwon:
Add hr to dropdown when separated flag is true for dropdown groups

image
image

### DIFF
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -7,6 +7,7 @@ import { Flex } from '../Flex';
 import { FlexItem } from '../Flex/Item';
 import { List } from '../List';
 import { ListGroupHeader } from '../List/GroupHeader';
+import { ListGroupSeparator } from '../List/GroupSeparator';
 import { ListItem } from '../List/Item';
 import { Tooltip } from '../Tooltip';
 
@@ -146,6 +147,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
+            {group.separated && <ListGroupSeparator />}
             {group.label && <ListGroupHeader>{group.label}</ListGroupHeader>}
             {renderItems(group.items)}
           </>
--- a/packages/big-design/src/components/Dropdown/spec.tsx
+++ b/packages/big-design/src/components/Dropdown/spec.tsx
@@ -47,6 +47,22 @@ const GroupedDropdownMock = (
   />
 );
 
+const LineSeparatedGroupedDropdownMock = (
+  <Dropdown
+    items={[
+      {
+        label: 'Label 1',
+        items: [{ content: 'Option 1', onItemClick }],
+      },
+      {
+        separated: true,
+        items: [{ content: 'Option 2', onItemClick }],
+      },
+    ]}
+    toggle={<Button>Button</Button>}
+  />
+);
+
 test('renders dropdown toggle', () => {
   const { getByRole } = render(DropdownMock);
   const toggle = getByRole('button');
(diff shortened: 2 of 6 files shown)


packages/big-design/src/components/Dropdown/spec.tsx
```diff
@@ -528,3 +528,27 @@ test('renders appropriate amount of list items', async () => {
...
+          items: [{ content: 'Option 2', onItemClick }],
+        },
+      ]}
+      toggle={<Button>Button</Button>}
+    />,
+  );
+  const toggle = getByRole('button');
+  fireEvent.click(toggle);
+
+  const hrItems = await waitForElement(() => container.querySelectorAll('HR'));
```
chanceaclark, line 551:
```suggestion
  const hrItems = await waitForElement(() => container.querySelector('hr'));
```

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -146,6 +147,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
+            {group.separated && <HR margin="none" />}
```
chanceaclark, line 150: You'll need to throw this into a `aria-hidden` `li` to make it semantic & accessible. Follow a similar pattern to `ListGroupHeader`.

packages/docs/PropTables/DropdownPropTable.tsx
```diff
@@ -192,6 +192,11 @@ const dropdownItemGroupProps: Prop[] = [
     types: 'string',
     description: 'Sets the label for the group.',
   },
+  {
+    name: 'separated',
+    types: 'boolean',
+    description: 'If true adds a line separator above label',
```
chanceaclark, line 198:
```suggestion
    description: 'If true, adds a line separator above the group.',
```

packages/big-design/src/components/List/GroupSeparator/ListGroupSeparator.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { memo } from 'react';
+
+import { HR } from '../../Typography';
+
+export const ListGroupSeparator: React.FC<React.LiHTMLAttributes<HTMLLIElement>> = memo(() => (
+  <li aria-hidden={true}>
+    <HR />
+  </li>
+));
```
chanceaclark, line 9: We need to add some of the similar features `ListGroupHeader` has for a11y.

```suggestion
export const ListGroupSeparator: React.FC<React.LiHTMLAttributes<HTMLLIElement>> = memo(() => (
  <li tabIndex={-1} onMouseDown={preventFocus} aria-hidden={true}>
    <HR />
  </li>
));

function preventFocus(event: React.MouseEvent<HTMLLIElement, MouseEvent>) {
  event.preventDefault();
}
```

packages/big-design/src/components/Dropdown/spec.tsx
```diff
@@ -528,3 +528,27 @@ test('renders appropriate amount of list items', async () => {
 
   expect(listItems.length).toBe(3);
 });
+
+test('renders appropriate amount of line separators (HR)', async () => {
```
chanceaclark, line 532: We should also test for correct accessibility roles are enabled for this list item. Essentially test if it has a proper `aria-hidden`, `tabindex`, and can't focus on it.

bc-chaz:
@bc-williamkwon could we include designs (screenshot) on this PR since it's UI?

jordan-massingill:
Would you mind adding some before/after screenshots?

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -146,6 +147,7 @@ export const Dropdown = memo(
       (group: DropdownItemGroup) => {
         return (
           <>
+            {group.separated && <ListGroupSeparator />}
```
bc-chaz, line 150: One issue that we're going to run into using groupings is the space between the groups.  If you look at the designs for storefront manager ui, you'll notice that there isn't really any added space around the line separator.

Storefront UI designs:
Screen Shot 2020-08-27 at 10 45 37 AM

Big Design dropdown w/o group labels:
Screen Shot 2020-08-27 at 10 47 30 AM

bc-chaz: here's the Figmas for the new UI: https://www.figma.com/file/uFeZH9uVYkPxBcQgCr1ASC/MSF-Merchandising-concept?node-id=1994%3A65969 - Theme Actions -> Step 2 -> Channel Context

chanceaclark: I believe I fixed that gap in #443.

bc-chaz: Interesting - when I remove the label in the BD sandbox (https://developer.bigcommerce.com/big-design/dropdown), it still shows the gap. Having some screenshots on this PR would def help

chanceaclark: Yeah, we haven't release a new version/updated the docs since that change.

bc-chaz:
@bc-williamkwon when you add the additional tests, could you also rebase? looks like we might be out of sync

packages/big-design/src/components/Dropdown/spec.tsx
```diff
@@ -528,3 +528,29 @@ test('renders appropriate amount of list items', async () => {
...
+    />,
+  );
+  const toggle = getByRole('button');
+  fireEvent.click(toggle);
+
+  const hrListItem = await waitForElement(() => container.querySelectorAll('hr')[0].parentElement as HTMLElement);
+  expect(hrListItem.getAttribute('aria-hidden')).toBe('true');
+  expect(hrListItem.getAttribute('tabindex')).toBe('-1');
+  fireEvent.mouseOver(hrListItem);
+  expect(document.activeElement).not.toEqual(hrListItem);
```
chanceaclark, line 555: Let's move this accessibility test into its own test case.

PR #447, opened 2020-09-01
feat(component): rename popover offset props and defaults

deini:
- Make Popover and Tooltips have the same default offset as dropdowns. 
- Rename offset props for Popover

### DIFF
--- a/packages/docs/PropTables/PopoverPropTable.tsx
+++ b/packages/docs/PropTables/PopoverPropTable.tsx
@@ -54,7 +54,7 @@ const popoverProps: Prop[] = [
     description: 'Determines the offset along the anchorElement.',
   },
   {
-    name: 'offsetY',
+    name: 'distance',
     types: 'number',
     defaultValue: '4',
     description: 'Determines the offset away from the anchorElement.',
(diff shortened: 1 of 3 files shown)


packages/docs/PropTables/PopoverPropTable.tsx
```diff
@@ -48,16 +48,16 @@ const popoverProps: Prop[] = [
...
   {
-    name: 'offsetX',
+    name: 'skidding',
     types: 'number',
     defaultValue: '0',
-    description: 'Determines the popover offset on the X axis.',
+    description: 'Determines the offset along the anchorElement.',
   },
   {
     name: 'offsetY',
```
rtalvarez, line 57: Should you rename this to `distance` also?

packages/docs/PropTables/PopoverPropTable.tsx
```diff
@@ -48,16 +48,16 @@ const popoverProps: Prop[] = [
     description: 'If set to true, the Popover will have the same width as its anchor element.',
   },
   {
-    name: 'offsetX',
+    name: 'skidding',
     types: 'number',
     defaultValue: '0',
-    description: 'Determines the popover offset on the X axis.',
+    description: 'Determines the offset along the anchorElement.',
```
rtalvarez, line 54: 🍹 I'd specify the direction of the offset, eg for skidding `Determines the horizontal offset along the anchorElement`
and for distance I'd say `Determines the vertical offset along the anchorElement`. I think it's a bit more clear

deini: In this case, there is no concept of horizontal/vertical since a Popover can be placed in any direction (top, right, etc).

image

PR #453, opened 2020-09-15
feat(component): adding actions to alert component

golcinho:
## What?
Adding `actions` prop to the `InlineAlert` and `Message` component.

- [x] Updated unit tests
- [x] Updated docs

## Screenshots
Screenshot 2020-09-29 at 12 15 28
Screenshot 2020-09-29 at 12 16 42
Screenshot 2020-09-29 at 12 16 53
Screenshot 2020-09-29 at 12 17 14

## Proof
Unit tests, local testing.

ping @chanceaclark

### DIFF
--- a/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
+++ b/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
@@ -1,7 +1,7 @@
 import { CloseIcon } from '@bigcommerce/big-design-icons';
 import React, { memo, useMemo } from 'react';
 
-import { excludePaddingProps, MarginProps } from '../../mixins';
+import { excludeMarginProps, excludePaddingProps, MarginProps } from '../../mixins';
 import { getActionVariant, getMessagingIcon, SharedMessagingProps } from '../../utils';
 import { Box } from '../Box';
 import { Button } from '../Button';
@@ -10,7 +10,6 @@ import { GridItem } from '../Grid';
 
 import {
   StyledActionsWrapper,
-  StyledActionWrapper,
   StyledHeader,
   StyledInlineAlert,
   StyledLink,
@@ -40,11 +39,16 @@ export const InlineAlert: React.FC<InlineAlertProps> = memo(({ className, style,
       props.actions && (
         <StyledActionsWrapper marginTop="xSmall">
           {props.actions.map(({ text, variant = 'secondary', ...actionProps }, index) => (
-            <StyledActionWrapper key={index}>
-              <Button mobileWidth="auto" variant={getActionVariant(variant)} {...actionProps}>
-                {text}
-              </Button>
-            </StyledActionWrapper>
+            <Button
+              key={index}
+              marginBottom="xSmall"
+              marginHorizontal="xxSmall"
+              mobileWidth="auto"
+              variant={getActionVariant(variant)}
+              {...excludeMarginProps(actionProps)}
+            >
+              {text}
+            </Button>
           ))}
         </StyledActionsWrapper>
       ),
--- a/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
+++ b/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
@@ -34,12 +34,12 @@ export const InlineAlert: React.FC<InlineAlertProps> = memo(({ className, style,
         <StyledActionsWrapper flexDirection="row" flexWrap="wrap" marginTop="xSmall">
           {props.actions.map(({ text, variant = 'secondary', ...actionProps }, index) => (
             <Button
+              {...excludeMarginProps(actionProps)}
               key={index}
               marginBottom="xSmall"
               marginHorizontal="xxSmall"
               mobileWidth="auto"
               variant={getActionVariant(variant)}
-              {...excludeMarginProps(actionProps)}
             >
               {text}
             </Button>
--- a/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
+++ b/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
@@ -8,13 +8,7 @@ import { Button } from '../Button';
 import { MessagingButton } from '../Button/private';
 import { GridItem } from '../Grid';
 
-import {
-  StyledActionsWrapper,
-  StyledHeader,
-  StyledInlineAlert,
-  StyledLink,
-  StyledMessageItem,
-} from './styled';
+import { StyledActionsWrapper, StyledHeader, StyledInlineAlert, StyledLink, StyledMessageItem } from './styled';
 
 export type InlineAlertProps = SharedMessagingProps & MarginProps;
 
--- a/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
+++ b/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
@@ -31,7 +31,7 @@ export const InlineAlert: React.FC<InlineAlertProps> = memo(({ className, style,
   const renderedActions = useMemo(
     () =>
       props.actions && (
-        <StyledActionsWrapper marginTop="xSmall">
+        <StyledActionsWrapper flexDirection="row" flexWrap="wrap" marginTop="xSmall">
           {props.actions.map(({ text, variant = 'secondary', ...actionProps }, index) => (
             <Button
               key={index}
--- a/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
+++ b/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
@@ -8,7 +8,14 @@ import { Button } from '../Button';
 import { MessagingButton } from '../Button/private';
 import { GridItem } from '../Grid';
 
-import { StyledHeader, StyledInlineAlert, StyledLink, StyledMessageItem } from './styled';
+import {
+  StyledActionsWrapper,
+  StyledActionWrapper,
+  StyledHeader,
+  StyledInlineAlert,
+  StyledLink,
+  StyledMessageItem,
+} from './styled';
 
 export type InlineAlertProps = SharedMessagingProps & MarginProps;
 
@@ -31,13 +38,15 @@ export const InlineAlert: React.FC<InlineAlertProps> = memo(({ className, style,
   const renderedActions = useMemo(
     () =>
       props.actions && (
-        <Box marginTop="xSmall">
+        <StyledActionsWrapper marginTop="xSmall">
           {props.actions.map(({ text, ...actionProps }, index) => (
-            <Button key={index} {...actionProps}>
-              {text}
-            </Button>
+            <StyledActionWrapper key={index}>
+              <Button mobileWidth="auto" {...actionProps}>
+                {text}
+              </Button>
+            </StyledActionWrapper>
           ))}
-        </Box>
+        </StyledActionsWrapper>
       ),
     [props.actions],
   );
--- a/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
+++ b/packages/big-design/src/components/InlineAlert/InlineAlert.tsx
@@ -4,6 +4,7 @@ import React, { memo, useMemo } from 'react';
 import { excludePaddingProps, MarginProps } from '../../mixins';
 import { getMessagingIcon, SharedMessagingProps } from '../../utils';
 import { Box } from '../Box';
+import { Button } from '../Button';
 import { MessagingButton } from '../Button/private';
 import { GridItem } from '../Grid';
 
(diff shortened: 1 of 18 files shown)


andriiVitvitskyi1990 (approved):

chanceaclark:
So everyone's is aware, we need to have a discussion with the Design team about this. We'll talk about it in the next feedback session.

golcinho:
ping @valterfatia

packages/big-design/src/components/Message/spec.tsx
```diff
@@ -105,3 +105,24 @@ test('does not forward styles', () => {
...
+  const { container, getAllByRole } = render(<Message actions={actions} messages={[{ text: 'Success' }]} />);
+  const buttons = getAllByRole('button') as HTMLButtonElement[];
+
+  expect(container.firstChild).toMatchSnapshot();
+
+  fireEvent.click(buttons[0]);
+
+  expect(fn).toHaveBeenCalledTimes(1);
+
+  fireEvent.click(buttons[0]);
```
deini, line 125: Should this be trying to click on `button[1]`? To avoid this and make it easier to follow maybe do something like:

```jsx
const primaryAction = getByRole('button', { name: /first action/i } );
const secondaryAction = getByRole('button', { name: /first action/i } );
```

And trigger the click on those.

golcinho: Great proposal! 👍

packages/big-design/src/components/InlineAlert/InlineAlert.tsx
```diff
@@ -27,12 +28,27 @@ export const InlineAlert: React.FC<InlineAlertProps> = memo(({ className, style,
 
   const renderedHeader = useMemo(() => header && <StyledHeader>{header}</StyledHeader>, [header]);
 
+  const renderedActions = useMemo(
+    () =>
+      props.actions && (
+        <Box marginTop="xSmall">
+          {props.actions.map(({ text, ...actionProps }, index) => (
```
chanceaclark, line 35: We probably want to ensure `children` can't get rendered.
```suggestion
          {props.actions.map(({ text, children, ...actionProps }, index) => (
```

Same applies for the other components.

packages/big-design/src/components/InlineAlert/InlineAlert.tsx
```diff
@@ -27,12 +28,27 @@ export const InlineAlert: React.FC<InlineAlertProps> = memo(({ className, style,
...
+            <Button key={index} {...actionProps}>
+              {text}
+            </Button>
```
chanceaclark, line 38: Make sure the intended behavior is for these to go to full-width on mobile.

packages/big-design/src/components/Message/Message.tsx
```diff
@@ -27,12 +28,27 @@ export const Message: React.FC<MessageProps> = memo(({ className, style, header,
...
 
+  const renderedActions = useMemo(
+    () =>
+      props.actions && (
+        <Box marginTop="xSmall">
+          {props.actions.map(({ text, ...actionProps }, index) => (
+            <Button key={index} {...actionProps}>
+              {text}
+            </Button>
+          ))}
```
chanceaclark, line 39: Same comments as above.

packages/docs/PropTables/InlineAlertPropTable.tsx
```diff
@@ -1,9 +1,26 @@
...
+      <>
+        Accepts an array of objects with{' '}
+        <NextLink href="/Button/ButtonPage" as="/button">
+          Button
+        </NextLink>{' '}
+        props and an additional <Code>text</Code> prop. See example for usage.
+      </>
+    ),
+  },
+];
```
chanceaclark, line 22: 🍹 Can we 🔤 this, by swapping the spread operator?
```suggestion
const inlineAlertProps: Prop[] = [
  {
    name: 'actions',
    types: 'object[]',
    description: (
      <>
        Accepts an array of objects with{' '}
        <NextLink href="/Button/ButtonPage" as="/button">
          Button
        </NextLink>{' '}
        props and an additional <Code>text</Code> prop. See example for usage.
      </>
    ),
  },
  ...sharedMessagingProps,
];
```

packages/docs/PropTables/MessagePropTable.tsx
```diff
@@ -1,9 +1,26 @@
 import React from 'react';
 
-import { PropTable, PropTableWrapper } from '../components';
+import { Code, NextLink, Prop, PropTable, PropTableWrapper } from '../components';
 
 import { sharedMessagingProps } from './shared';
 
+const messageProps: Prop[] = [
+  ...sharedMessagingProps,
```
chanceaclark, line 8: 🍹 Same comment as above.

packages/big-design/src/utils/messagingHelpers.tsx
```diff
@@ -25,6 +27,10 @@ export type MessageLinkItem = Pick<LinkProps, 'external' | 'href' | 'target'> &
   text: string;
 };
 
+export interface MessageAction extends Omit<ButtonProps, 'children'> {
```
chanceaclark, line 30: Do you think we should limit the Button's `variant` prop to `secondary` and `subtle`? There should only be 1 `primary` button on a page, and I would assume it wouldn't be in a messaging component.

cc. @valterfatia @eugene-polev

chanceaclark (commented):
@golcinho thanks for this! Here's a few changes.

packages/big-design/src/components/InlineAlert/InlineAlert.tsx
golcinho: thank you for the comment 🙇 

in the current implementation, it works correctly!

As you see `children` from props is overrides by the content inside `Button` and if we will pass `undefined` as `text` and pass `children` from action props it's will be anyway overridden by the `undefined`.

You don't mind, I prefer to leave it as is.

packages/big-design/src/components/InlineAlert/InlineAlert.tsx
golcinho: 👍 
Screenshot 2020-09-23 at 12 13 12

eugene-polev: I think we shouldn't stretch buttons to the full width but use their initial size even on mobile. See examples below 
image

golcinho: @eugene-polev I fixed it 👍 

Screenshot 2020-09-25 at 11 26 17
Screenshot 2020-09-25 at 11 26 40

packages/big-design/src/utils/messagingHelpers.tsx
eugene-polev: I think it makes sense

golcinho: Guys, please let me know if everyone agrees with the proposal from @chanceaclark and I will make changes to the current PR 🙇

golcinho: @chanceaclark @eugene-polev 
I'm done with this topic so, now available only 2 variants: `secondary` and `subtle`(docs & screenshots were updated).

packages/big-design/src/utils/messagingHelpers.tsx
```diff
@@ -25,6 +27,11 @@ export type MessageLinkItem = Pick<LinkProps, 'external' | 'href' | 'target'> &
   text: string;
 };
 
+export interface MessageAction extends Omit<ButtonProps, 'children' | 'mobileWidth' | 'variant'> {
+  text: string;
+  variant?: Extract<ButtonProps['variant'], 'secondary' | 'subtle'>;
```
chanceaclark, line 32: 🍹 Maybe use `Exclude<ButtonProps['variant'], 'primary'>` in case we add more variants to buttons in the future?

packages/big-design/src/components/Message/styled.ts
```diff
@@ -44,7 +45,23 @@ export const StyledLink = styled(Link)`
   vertical-align: middle;
 `;
 
+export const StyledActionsWrapper = styled(Flex)`
+  flex-direction: row;
+  flex-wrap: wrap;
```
chanceaclark, line 50: We should apply these styles to the usage.

packages/big-design/src/components/Message/styled.ts
```diff
@@ -44,7 +45,23 @@ export const StyledLink = styled(Link)`
...
+  margin-bottom: -${({ theme }) => theme.spacing.xSmall};
+  margin-left: -${({ theme }) => theme.spacing.xxSmall};
+  margin-right: -${({ theme }) => theme.spacing.xxSmall};
+`;
+
+export const StyledActionWrapper = styled(FlexItem)`
+  margin-bottom: ${({ theme }) => theme.spacing.xSmall};
+  margin-left: ${({ theme }) => theme.spacing.xxSmall};
+  margin-right: ${({ theme }) => theme.spacing.xxSmall};
+`;
```
chanceaclark, line 60: Why not get rid of the `ActionWrapper` and let the `Buttons` be the flex items, they have the margins already:

```tsx
<Button
  key={index}
  variant={getActionVariant(variant)}
  {...excludeMarginProps(actionProps)}
  mobileWidth="auto"
>
```

Note: I included the `excludeMarginProps` utility so that users can't add their own margins. You might want to enforce this on the type too.

packages/big-design/src/components/Message/Message.tsx
```diff
@@ -27,12 +35,29 @@ export const Message: React.FC<MessageProps> = memo(({ className, style, header,
...
+            </StyledActionWrapper>
+          ))}
+        </StyledActionsWrapper>
+      ),
+    [props.actions],
+  );
+
   return (
-    <StyledMessage {...filteredProps} role="alert">
+    <StyledMessage {...filteredProps} backgroundColor="white" role="alert">
```
chanceaclark, line 55: 🙇🏼‍♂️ Thanks for this!

packages/big-design/src/components/Message/Message.tsx
```diff
@@ -27,12 +35,29 @@ export const Message: React.FC<MessageProps> = memo(({ className, style, header,
...
+  const renderedActions = useMemo(
+    () =>
+      props.actions && (
+        <StyledActionsWrapper marginTop="xSmall">
+          {props.actions.map(({ text, variant = 'secondary', ...actionProps }, index) => (
+            <StyledActionWrapper key={index}>
+              <Button mobileWidth="auto" variant={getActionVariant(variant)} {...actionProps}>
+                {text}
+              </Button>
+            </StyledActionWrapper>
```
chanceaclark, line 47: See comment above.

packages/big-design/src/components/InlineAlert/InlineAlert.tsx
```diff
@@ -27,12 +35,29 @@ export const InlineAlert: React.FC<InlineAlertProps> = memo(({ className, style,
...
+  const renderedActions = useMemo(
+    () =>
+      props.actions && (
+        <StyledActionsWrapper marginTop="xSmall">
+          {props.actions.map(({ text, variant = 'secondary', ...actionProps }, index) => (
+            <StyledActionWrapper key={index}>
+              <Button mobileWidth="auto" variant={getActionVariant(variant)} {...actionProps}>
+                {text}
+              </Button>
+            </StyledActionWrapper>
```
chanceaclark, line 47: See comment above.

packages/big-design/src/components/InlineAlert/styled.ts
```diff
@@ -46,7 +47,23 @@ export const StyledLink = styled(Link)`
...
+  margin-bottom: -${({ theme }) => theme.spacing.xSmall};
+  margin-left: -${({ theme }) => theme.spacing.xxSmall};
+  margin-right: -${({ theme }) => theme.spacing.xxSmall};
+`;
+
+export const StyledActionWrapper = styled(FlexItem)`
+  margin-bottom: ${({ theme }) => theme.spacing.xSmall};
+  margin-left: ${({ theme }) => theme.spacing.xxSmall};
+  margin-right: ${({ theme }) => theme.spacing.xxSmall};
+`;
```
chanceaclark, line 62: See comment above.

chanceaclark (commented):
Few more comments around the updates. Let me know if it doesn't make sense!

packages/big-design/src/utils/messagingHelpers.tsx
golcinho: I was thought that we should support only `secondary` and `subtle` variants 😄  So, yeah, it's not a problem to change it 🤝

chanceaclark (approved):

PR #455, opened 2020-09-28
fix(component): form controls error states

golcinho:
## What? / Why?
Fixing the error state of the `Input` and `Textarea` components.

Cases when we don't show error state:
- Empty array;
- When we get an empty `error` array after firing `validationError` in the case when we receive error property as an `array` or an `array` with the "invalid" elements.

### Before:
Screenshot 2020-09-28 at 13 20 40
Screenshot 2020-09-28 at 13 21 18
Screenshot 2020-09-29 at 11 19 43
Screenshot 2020-09-30 at 12 00 43
Screenshot 2020-09-30 at 12 01 14
Screenshot 2020-09-30 at 12 01 25

### After:
Screenshot 2020-09-28 at 13 30 00
Screenshot 2020-09-28 at 13 30 11
Screenshot 2020-09-29 at 11 19 59
Screenshot 2020-09-30 at 12 01 46
Screenshot 2020-09-30 at 12 02 02
Screenshot 2020-09-30 at 12 02 12

## Proof
- Existing unit tests;
- Tested locally.

ping @chanceaclark

### DIFF
--- a/packages/big-design/src/components/Input/Input.tsx
+++ b/packages/big-design/src/components/Input/Input.tsx
@@ -143,7 +143,7 @@ const StyleableInput: React.FC<InputProps & PrivateProps> = ({
     if (Array.isArray(error)) {
       error.forEach(validateError);
 
-      return error;
+      return error.length > 0 ? error : null;
     }
 
     return validateError(error);
--- a/packages/big-design/src/components/Input/Input.tsx
+++ b/packages/big-design/src/components/Input/Input.tsx
@@ -141,9 +141,13 @@ const StyleableInput: React.FC<InputProps & PrivateProps> = ({
     };
 
     if (Array.isArray(error)) {
-      error.forEach(validateError);
+      const nextError = error.reduce<Array<Props['error']>>((acc, errorItem) => {
+        const nextErrorItem = validateError(errorItem);
 
-      return error.length > 0 ? error : null;
+        return nextErrorItem ? [...acc, nextErrorItem] : acc;
+      }, []);
+
+      return nextError.length > 0 ? nextError : null;
     }
 
     return validateError(error);
--- a/packages/big-design/src/components/Input/Input.tsx
+++ b/packages/big-design/src/components/Input/Input.tsx
@@ -141,7 +141,7 @@ const StyleableInput: React.FC<InputProps & PrivateProps> = ({
     };
 
     if (Array.isArray(error)) {
-      const nextError = error.reduce<Array<Props['error']>>((acc, errorItem) => {
+      const nextError = error.reduce<Array<React.ReactNode>>((acc, errorItem) => {
         const nextErrorItem = validateError(errorItem);
 
         return nextErrorItem ? [...acc, nextErrorItem] : acc;
--- a/packages/big-design/src/components/Input/spec.tsx
+++ b/packages/big-design/src/components/Input/spec.tsx
@@ -1,4 +1,5 @@
 import { AddIcon } from '@bigcommerce/big-design-icons';
+import { theme } from '@bigcommerce/big-design-theme';
 import React, { createRef } from 'react';
 import 'jest-styled-components';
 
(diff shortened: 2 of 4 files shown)


andriiVitvitskyi1990 (approved):

VitalyBesk (approved):

deini (approved):

chanceaclark (commented):
Can we add a test or two around this? Especially with an empty string error.

golcinho:
> Can we add a test or two around this? Especially with an empty string error.

I see that all error cases already covered... I will add some tests to check the correct styles related to the `input` state

packages/big-design/src/components/Input/Input.tsx
```diff
@@ -141,9 +141,13 @@ const StyleableInput: React.FC<InputProps & PrivateProps> = ({
...
-      error.forEach(validateError);
+      const nextError = error.reduce<Array<Props['error']>>((acc, errorItem) => {
+        const nextErrorItem = validateError(errorItem);
 
-      return error;
+        return nextErrorItem ? [...acc, nextErrorItem] : acc;
+      }, []);
+
+      return nextError.length > 0 ? nextError : null;
     }
```
golcinho, line 151: ℹ️ Logic updated, be aware 👀

andriiVitvitskyi1990 (approved):

BC-YIvashchenko (approved):

VitalyBesk (approved):

packages/big-design/src/components/Input/spec.tsx
```diff
@@ -224,27 +234,58 @@ test('error shows with valid string', () => {
...
+    const { container } = render(
+      <FormGroup>
+        <Input error={[]} />
+      </FormGroup>,
+    );
+    const styledInputWrapper = container.querySelector('[class*="StyledInputWrapper"]');
 
-  errors.forEach((error) => expect(getByText(error)).toBeInTheDocument());
+    expect(container.querySelector('[class*="StyledError"]')).not.toBeInTheDocument();
+    expect(styledInputWrapper).toHaveStyleRule('border', basicBorderStyle);
```
chanceaclark, line 262: I don't think we need to test if there if the wrapper has the stylings 🤔  Querying to see if the `StyledError` container is in the document should be sufficient.

packages/big-design/src/components/Input/Input.tsx
```diff
@@ -141,9 +141,13 @@ const StyleableInput: React.FC<InputProps & PrivateProps> = ({
     };
 
     if (Array.isArray(error)) {
-      error.forEach(validateError);
+      const nextError = error.reduce<Array<Props['error']>>((acc, errorItem) => {
```
chanceaclark, line 144: This type is saying that it could be an `Array<React.ReactNode> | Array<Array<React.ReactNode>>`. I think we could just change it to this:
```suggestion
      const nextError = error.reduce<Array<React.ReactNode>>((acc, errorItem) => {
```

chanceaclark:
Also, just thought of this, but was there a reason for not including this feature in all the other FormControl components (`Textarea` and `Select`)?

packages/big-design/src/components/Input/spec.tsx
golcinho: Without this fix we have cases when the input has an error state (red border etc.) but, the `StyledError` is not in the document. So, this PR was created to fix these cases.

golcinho:
> Also, just thought of this, but was there a reason for not including this feature in all the other FormControl components (`Textarea`, `Select`, `Datepicker`, and `Timepicker`)?

- `Select`, `Multiselect`, `Datepicker` and `Timepricker` use `Input` inside so, this issue was fixed in the scope of `Input` fix;
- Added fix for the `Textarea`.

🤝

packages/big-design/src/components/Input/spec.tsx
chanceaclark: Ah yeah, good callout!

chanceaclark (approved):

PR #456, opened 2020-10-06
feat: update downshift and fix tests

jorgemoya:
Update `downshift` to use the latest version and make the changes necessary to make it work. Fixed breaking tests due to new changes.

There is a still a performance pass pending but will be done at a different time.


chanceaclark (approved):
Looks good to me.

deini (approved):

PR #462, opened 2020-10-15
feat(component): show icon on selected item

jorgemoya:
## What
Show icon on selected item

## Screenshot
Screen Shot 2020-10-15 at 11 21 42 AM

Closes #400

PR #463, opened 2020-10-16
fix(component): add missing prop to open dropdown

jorgemoya:
Dropdown's `<List />` component was missing the `isOpen` prop.

Will lurk the `Downshift` library to figure out why `getMenuProps` returns `any`, which caused an oversight by me thinking `ts` would catch an issue like this.

### DIFF
--- a/packages/big-design/src/components/Select/Select.tsx
+++ b/packages/big-design/src/components/Select/Select.tsx
@@ -298,6 +298,7 @@ export const Select = typedMemo(
       isOpen,
       onOptionChange,
       openMenu,
+      selectedOption,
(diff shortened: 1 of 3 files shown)


packages/big-design/src/components/Select/Select.tsx
```diff
@@ -298,6 +298,7 @@ export const Select = typedMemo(
       isOpen,
       onOptionChange,
       openMenu,
+      selectedOption,
```
jorgemoya, line 301: Missed adding this dependency on #462

chanceaclark (approved):

PR #470, opened 2020-10-28
fix(docs): alert docs - alert management import clarification

bc-maksym-konohorov:
Slightly changed docs for `Alerts` page.

Added some information regarding how to use `createAlertsManager` instance:
Screen Shot 2020-10-28 at 15 31 21

Changed code in `CodeSnippet` to be string, as it didn't render correctly, in was rendering extra `}` character, which could confuse consumers of this example:
Screen Shot 2020-10-28 at 15 30 53

### DIFF
--- a/packages/docs/pages/Alert/AlertPage.tsx
+++ b/packages/docs/pages/Alert/AlertPage.tsx
@@ -1,4 +1,4 @@
-import { AlertProps, AlertsManager, Button, H0, H1, H2, Message, Text } from '@bigcommerce/big-design';
+import { AlertProps, Button, H0, H1, H2, Message, Text } from '@bigcommerce/big-design';
 import React from 'react';
 
 import { Code, CodePreview, CodeSnippet } from '../../components';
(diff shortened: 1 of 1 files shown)


jdschmitt (approved):
This is a great update @bc-maksym-konohorov. As a consumer of BigDesign, this does make it more clear for me. Thanks for taking the time to improve our docs!

chanceaclark (approved):

jorgemoya (approved):

packages/docs/pages/Alert/AlertPage.tsx
```diff
@@ -69,27 +69,44 @@ const AlertPage = () => (
...
-          <>
-            {/* ... */}
-            <AlertsManager manager={alertsManager} />
-            {/* ... */}
-          </>
-        );
-      }}
-
-      {/* jsx-to-string:end */}
+      {`
```
deini, line 77: We should prob investigate the extra `}` you are mentioning.

packages/docs/pages/Alert/AlertPage.tsx
```diff
@@ -69,27 +69,44 @@ const AlertPage = () => (
...
     </CodeSnippet>
 
     <Text>
-      This works in conjunction with an instance created by <Code>createAlertsManager</Code> function below.
+      This works in conjunction with an instance created by <Code>createAlertsManager</Code> function below. You need to
+      export <Code>alertsManager</Code> instance and then you can import it in child components in order to trigger
+      alert:
     </Text>
 
+    <CodeSnippet>
```
deini, line 99: Maybe we should add an example on how to create a `useAlerts` hook.

packages/docs/pages/Alert/AlertPage.tsx
bc-maksym-konohorov: My suggestion - jsx-to-string transformer works in a weird way, it depends on indentation in code, I moved two closing curly braces `}}` back, and it rendered it right. but there was a problem when you add just regular js code line before function.  it rendered it incorrectly, and also linter on commit hook was breaking formatting.

packages/docs/pages/Alert/AlertPage.tsx
bc-maksym-konohorov: Agree it's a good idea, but I need to play with hook to become familiar with it, and unfortunately I'm not sure If I will have time for that in this sprint.

PR #474, opened 2020-11-23
docs(component): updates switch docs not for use in forms

jordan-massingill:
### WHAT
Updates `Switch` documentation to state that this component is for immediate "toggle" changes and is thus not appropriate for use in submittable forms. Removes `Form` from code example, replaces with `Box` - background changes color for visual representation of component usage, but I'm happy to use better examples.
Screen Shot 2020-11-23 at 2 58 04 PMScreen Shot 2020-11-23 at 4 40 52 PM

### DIFF
--- a/packages/docs/pages/Switch/SwitchPage.tsx
+++ b/packages/docs/pages/Switch/SwitchPage.tsx
@@ -18,7 +18,7 @@ const SwitchPage = () => (
       {function Example() {
         const [checked, setChecked] = useState(false);
         const handleChange = () => setChecked(!checked);
-        const boxColor = checked ? 'brand' : 'secondary20';
+        const boxColor = checked ? 'white' : 'secondary20';
 
         return (
           <Box display="inline-block" padding="xLarge" backgroundColor={boxColor}>
--- a/packages/docs/pages/Switch/SwitchPage.tsx
+++ b/packages/docs/pages/Switch/SwitchPage.tsx
@@ -1,4 +1,4 @@
-import { Box, H0, H1, Switch, Text } from '@bigcommerce/big-design';
+import { H0, H1, Switch, Text } from '@bigcommerce/big-design';
 import React, { useState } from 'react';
 
 import { Code, CodePreview } from '../../components';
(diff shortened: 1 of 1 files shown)


packages/docs/pages/Switch/SwitchPage.tsx
```diff
@@ -9,21 +9,21 @@ const SwitchPage = () => (
...
+      Switches are a stylized <Code>input[type="checkbox"]</Code> with controllable checked/unchecked states. Switches
+      are intended to be used for immediate toggle actions and are therefore not intended to be used in forms.
     </Text>
 
     <CodePreview>
       {/* jsx-to-string:start */}
       {function Example() {
         const [checked, setChecked] = useState(false);
         const handleChange = () => setChecked(!checked);
+        const boxColor = checked ? 'brand' : 'secondary20';
```
chanceaclark, line 21: I don't think the `brand` color on `checked` is very accessible 🤔 Could we just leave the normal background color?

jordan-massingill: Which background color is normal? I was going to use `primary` but then the switch blends in.

chanceaclark: Just the inherit white background should be good.

packages/docs/pages/Switch/SwitchPage.tsx
```diff
@@ -9,21 +9,21 @@ const SwitchPage = () => (
...
         const handleChange = () => setChecked(!checked);
+        const boxColor = checked ? 'white' : 'secondary20';
 
         return (
-          <Form>
-            <FormGroup>
-              <Switch checked={checked} onChange={handleChange} />
-            </FormGroup>
-          </Form>
+          <Box display="inline-block" padding="xLarge" backgroundColor={boxColor}>
```
chanceaclark, line 24: Let's just remove this altogether.
```suggestion
          <Box padding="xLarge">
```

jordan-massingill: I understand removing the `display` prop (I forgot to take it out after some experimentation), but the intent of the box color was to show the `Switch` as an immediate state toggle. Do you have any suggestions of how that can be made visually clear/explicit, or do you think the description alone will be enough for anyone to understand the intent?

chanceaclark: I think the description along should be :gucci: enough. We just want to make sure we have it documented in text that the intent on using this component is not for forms.

chanceaclark (approved):

PR #475, opened 2020-12-01
fix(component): performance improvements for Dropdown/Select components

jorgemoya:
### What
Long time coming, but we noticed every single item would rerender whenever hovering an option, so we knew performance could be improved.

This changes fix that by memoizing the props that were triggering a rerender, thus only items with true changes will rerender. I added some props to some components for feature parity (like groups to MultiSelect). A lot of things were moved around, hoping this helps with clarity of these components.

### Some area of opportunity:
- Typescript stuff. Having some issues with this and had to drop some `as`. Let me know if we can fix them some other way!
- When selection happens in MultiSelect component, all items rerender.
- All wrappers rerender atm thanks to popper when scrolling, **FIXED**.
- In the future, I would like to take the time to transform the MultiSelect component to use `useMultipleSelection` downshift hook.

### Screenshots
Kapture 2020-12-01 at 14 56 29


packages/big-design/src/components/List/Item/Item.tsx
```diff
@@ -1,32 +1,193 @@
...
 }
 
-const StyleableListItem: React.FC<ListItemProps & PrivateProps> = ({
-  autoWidth = false,
-  actionType = 'normal' as 'normal',
-  forwardedRef,
-  isAction = false,
-  ...props
-}) => {
+type Items = DropdownItem | DropdownLinkItem | SelectOption<unknown> | SelectAction;
```
chanceaclark, line 34: I think you can add a generic to `Items`
```suggestion
type Items<T> = DropdownItem | DropdownLinkItem | SelectOption<T> | SelectAction;
```

packages/big-design/src/components/List/List.tsx
```diff
@@ -1,34 +1,246 @@
...
-  maxHeight: number;
-  update(): Promise<Partial<State> | null>;
+  items: DropdownProps['items'] | SelectProps<T>['options'];
+  isDropdown?: boolean;
+  maxHeight?: number;
+  selectedItem?: SelectOption<T> | null;
+  selectedItems?: Array<SelectOption<T>> | null;
+  addItem?(item: SelectOption<T>): void;
+  getItemProps: UseSelectPropGetters<any>['getItemProps'];
+  getMenuProps: UseSelectPropGetters<any>['getMenuProps'];
```
chanceaclark, line 29:
```suggestion
  getItemProps: UseSelectPropGetters<DropdownProps['items'] | SelectProps<T>['options']>['getItemProps'];
  getMenuProps: UseSelectPropGetters<DropdownProps['items'] | SelectProps<T>['options']>['getMenuProps'];
```

packages/big-design/src/components/List/List.tsx
```diff
@@ -1,34 +1,246 @@
...
+      >
+        {isOpen && renderChildren}
+      </StyledList>
+    );
+  },
+);
 
 export const List = memo(
-  forwardRef<HTMLUListElement, ListProps>((props, ref) => <StyleableList {...props} forwardedRef={ref} />),
+  forwardRef<HTMLUListElement, ListProps<unknown>>((props, ref) => <StyleableList {...props} forwardedRef={ref} />),
```
chanceaclark, line 220: Can we add a comment on why we need this `unknown`?

packages/big-design/src/components/List/styled.tsx
```diff
@@ -19,4 +16,9 @@ export const StyledList = styled.ul<Partial<ListProps>>`
   z-index: ${({ theme }) => theme.zIndex.popover};
 `;
 
+export const StyledMenuContainer = styled.div`
```
chanceaclark, line 19: `Box` has a `zIndex` prop, could we use that instead?

packages/big-design/src/components/Select/Select.tsx
```diff
@@ -1,9 +1,8 @@
-import { CheckIcon } from '@bigcommerce/big-design-icons';
+import { useUniqueId } from '@bigcommerce/big-design-icons';
```
chanceaclark, line 1: Can we import this from the local package instead?

packages/big-design/src/components/Select/styled.tsx
```diff
@@ -34,4 +34,3 @@ export const DropdownButton = styled(StyledButton)`
 StyledStatusMessage.defaultProps = { theme: defaultTheme };
 StyledDropdownIcon.defaultProps = { theme: defaultTheme };
 StyledInputContainer.defaultProps = { theme: defaultTheme };
-DropdownButton.defaultProps = { theme: defaultTheme };
```
chanceaclark, line 37: Why is this getting removed?

packages/big-design/src/components/List/List.tsx
jorgemoya: This is the only change that isn't working. I can't set directly from the Dropdown or Select props because those items contain groups, which Downshift doesn't get. But even setting a type like `DropdownItem | DropdownLinkItem | SelectOption<T> | SelectAction` returns an error on all components, because `Select` (for example) only passes in `SelectOption<T> | SelectAction`.

packages/big-design/src/components/Select/styled.tsx
jorgemoya: I keep getting this error on my type 

```
Type of property 'defaultProps' circularly references itself in mapped type 'Pick<ForwardRefExoticComponent<Pick<Pick<any, Exclude<keyof ReactDefaultizedProps<StyledComponentInnerComponent<WithC>, ComponentPropsWithRef<StyledComponentInnerComponent<WithC>>>, StyledComponentInnerAttrs<...> | ... 1 more ... | StyledComponentInnerAttrs<...>> | Exclude<...> | Exclude<...> | Exclude<...>> & Parti...'.
```

jorgemoya: nvm fixed I was using the wrong component

packages/big-design/src/components/List/List.tsx
chanceaclark: I tried to play around with this for a bit, and yeah it doesn't seem easily possible without separating the `List` component like we've chatted about. Should be good for now.

packages/big-design/src/components/Select/styled.tsx
```diff
@@ -3,7 +3,7 @@ import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import { hideVisually } from 'polished';
 import styled, { css } from 'styled-components';
 
-import { StyledButton } from '../Button/styled';
+import { StyleableButton } from '../Button/Button';
```
chanceaclark, line 6: 🍹 
```suggestion
import { StyleableButton } from '../Button/private';
```

chanceaclark (approved):
Looks good, one last 🍹 comment. I would also get 1 more set of 👀 on this.

jorgemoya:
We're looking for one more set of eyes @bigcommerce/frontend

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,298 +1,119 @@
...
+    // Popper
+    const referenceRef = useRef(null);
+    const popperRef = useRef(null);
+
+    const { attributes, styles, update } = usePopper(referenceRef.current, popperRef.current, {
+      modifiers: [
+        {
+          name: 'eventListeners',
+          options: {
+            scroll: isOpen,
```
deini, line 71: nice

deini (approved):

jorgemoya:
There is an issue with Downshift and memoized components. I have opened a PR for a potential fix, I would advise not to merge until the fix is released. https://github.com/downshift-js/downshift/pull/1205

PR #481, opened 2020-12-10
refactor(component): refactor Tree component API to StatefulTree

chanceaclark:
## What

Refactors `Tree` component to be a more internal component and adds a new `StatefulTree` component.

## Why

The old `Tree` API dealt with a lot of the state management and caused use usage issues:
1. The old API wasn't very "React-y", in the sense that you couldn't re-render nodes based on any API updates.
2. Focus on the page went directly to the `Tree` component due to some state management requirements.
3. `onClick` handlers were being requested for certain DOM interactions.

## Notes

- We are not going to be exposing the `Tree` component anymore until we find we need more granular usages (async node fetches, complex node manipulation, etc...). Until that point in time, you should be using `StatefulTree`.
- Hopefully the `Tree` component is now in a good state that we can add drag-n-drop functionality. This functionality will be for another PR.

```
BREAKING CHANGE:
`Tree` component and now `StatefulTree` with new API changes.
See dev docs for new API.
```

ping @bigcommerce/frontend


jorgemoya (commented):
Looking good so far. I liked what you did with the hooks and context to clean it up. 👍

packages/big-design/src/components/Tree/types.ts
```diff
@@ -1,46 +1,66 @@
...
-  expanded?: boolean;
   icon?: React.ReactChild | null;
   id: TreeNodeId;
   label: string;
-  selected?: boolean;
   value?: T;
 }
 
-export type TreeNodeRef<T> = Pick<TreeNodeProps<T>, 'children'>;
+export interface TreeExandable {
```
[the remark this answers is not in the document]
chanceaclark: Good catch! Thanks!

jorgemoya (approved):
Looks great, good job! I couldn't find any major issues, it's a great example of how to build complex components for the future. Lets get another pair of eyes to look at it.

packages/big-design/src/components/Tree/hooks/spec.ts
```diff
@@ -0,0 +1,347 @@
...
+      const { result } = renderHook(() => useTreeKeyEvents({ onFocus, onSelect, onToggle, nodeMap, visibleNodes }));
+
+      result.current(
+        ({
+          currentTarget: 1,
+          target: 1,
+          key: 'ArrowLeft',
+          preventDefault,
+          stopPropagation,
+        } as unknown) as React.KeyboardEvent<HTMLLIElement>,
```
rtalvarez, line 272: 🍹  could we use `new KeyboardEvent(...)` instead of manually constructing the object, to get rid of the `as` castings? never tried this before so i'm not sure if it'd work

packages/big-design/src/components/Tree/hooks/useNodeMap.ts
```diff
@@ -0,0 +1,52 @@
...
+const getParentId = (nodeMap: NodeMap, id: TreeNodeId) => {
+  const iterator = nodeMap.entries();
+
+  for (const [, value] of iterator) {
+    if (value.children && value.children.includes(id)) {
+      return value.id;
+    }
+  }
+
+  return undefined;
```
rtalvarez, line 22: this `return undefined` seems to be unnecessary?

rtalvarez (commented):
looks pretty good to me overall, just a couple of questions

packages/big-design/src/components/Tree/hooks/spec.ts
chanceaclark: Good idea, let me try it out!

chanceaclark: So I checked it out, not sure the types are going to work cause `React.KeyboardEvent !== KeyboardEvent`. The react type provides a generic element to be passed in, while the native `KeyboardEvent` doesn't and relies on `EventTarget` as it's "element".

BC-YIvashchenko (approved):
Great 🥇

packages/big-design/src/components/StatefulTree/hooks/useSelectable.ts
```diff
@@ -0,0 +1,104 @@
...
+        } else {
+          setSelectedNodes([...selectedNodes, nodeId]);
+          setSelectedValues([...selectedValues, value]);
+        }
+      } else if (type === 'radio' && !selectedNodes.includes(nodeId)) {
+        setSelectedNodes([nodeId]);
+        setSelectedValues([value]);
+      }
+    }
+  };
```
deini, line 98: 🍹 

```suggestion
  const onSelect: TreeSelectable<T>['onSelect'] = (nodeId, value) => {
    if (disabledNodes?.includes(nodeId)) {
      return;
    }

    if (type === 'multi') {
      if (selectedNodes.includes(nodeId)) {
        setSelectedNodes((prevNodes) => prevNodes.filter((prevNodeId) => prevNodeId !== nodeId));
        setSelectedValues((prevValues) => prevValues.filter((prevValue) => prevValue !== value));
      } else {
        setSelectedNodes([...selectedNodes, nodeId]);
        setSelectedValues([...selectedValues, value]);
      }
    }

    if (type === 'radio' && !selectedNodes.includes(nodeId)) {
      setSelectedNodes([nodeId]);
      setSelectedValues([value]);
    }
  };
```

packages/big-design/src/components/StatefulTree/hooks/useVisibleNodes.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  return recursiveSearch({
+    nodes: parentNodes,
+    expandedNodes,
+    nodeMap,
+  });
+};
+
+export const useVisibleNodes = ({ expandedNodes, nodeMap }: UseVisibleNodesProps) => {
+  const [visibleNodes, setVisibleNodes] = useState(buildVisibleNodes({ expandedNodes, nodeMap }));
```
deini, line 48:
```suggestion
  const [visibleNodes, setVisibleNodes] = useState(() => buildVisibleNodes({ expandedNodes, nodeMap }));
```

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
```diff
@@ -21,268 +21,134 @@ import {
...
+      // Checks to see if every element inside the selectableRef gets clicked.
+      if ((e?.target instanceof Node && selectableRef.current?.contains(e?.target)) || children === undefined) {
+        return;
+      }
 
-        if (callbackValue) {
-          thisRef.current = callbackValue;
-        }
+      if (typeof expandable.onToggle === 'function') {
+        await expandable.onToggle(id, isExpanded);
```
deini, line 66: `await` is not doing much here, unless the types are off for `onToggle`, `onCollapse`, and `onExpand`.

rtalvarez:
Btw did you check the profiler to get an overall feel on the re-rendering performance of this component? Some real usecases for this component might get a bit heavy (eg category trees)

chanceaclark:
@rtalvarez I did and it's going to rerender frequently due to the use of the context API. I had some discussions with @deini about this, and the only real solution to circumvent this use a state library (Redux/RxJS/etc.). We may need to find a solution for this for all of our stateful/complex components in the future.

---

I'll keep this PR open a bit longer for comments, and in the meantime I will stress test it with a larger dataset.

packages/big-design/src/components/Tree/Tree.tsx
```diff
@@ -1,53 +1,56 @@
...
-          onExpand={onExpand}
-          onSelect={onSelect}
-          selectable={selectable}
-          state={state}
-          treeId={treeId}
-        />
-      )),
-    [iconless, nodes, onCollapse, onExpand, onSelect, selectable, state, treeId],
-  );
+  useEffect(() => {
```
deini, line 41: I see you are using this pattern in multiple places. Consider abstracting it to a `useDerivedState` hook.

You can also optimize it a bit, `useEffect` runs after browser paint, you can use another `useState` to keep track of the previous props value. This will trigger a re-render before commiting the changes to the DOM. [More info](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops).

packages/big-design/src/components/StatefulTree/hooks/useSelectable.ts
```diff
@@ -0,0 +1,104 @@
...
+
+interface GetDefaultSelectedValuesProps<T> extends Pick<UseSelectableProps<T>, 'nodes' | 'type'> {
+  selectedNodes: TreeNodeId[];
+}
+
+const getDefaultSelectedValues = <T>({ nodes, selectedNodes, type }: GetDefaultSelectedValuesProps<T>) => {
+  if (type === 'radio') {
+    const selectedNode = depthFirstSearch(nodes, ({ id, value }) => selectedNodes.includes(id) && value !== undefined);
+
+    if (selectedNode && selectedNode.value) {
```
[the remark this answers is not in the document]
chanceaclark: Good catch! Thanks!

packages/big-design/src/components/StatefulTree/hooks/useSelectable.ts
```diff
@@ -0,0 +1,104 @@
...
+      return [selectedNode.value];
+    }
+  }
+
+  if (type === 'multi') {
+    return (
+      depthFirstSearch(nodes, ({ id, value }) => selectedNodes.includes(id) && value !== undefined, false)?.reduce<T[]>(
```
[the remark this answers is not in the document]
chanceaclark: In this case, we know that `value` is defined, but if we map, TS still won't know that - so the output type would be 
```ts
(T | undefined)[]
```

That was the reasoning behind reducing, to ensure the type was alway `T[]`.

Good catch on the falsey values. Thanks!

packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
chanceaclark: Remnant of the old method of lazy-loading children. I'll most likely rip that out for now.

deini (approved):

packages/big-design/src/components/Tree/Tree.tsx
chanceaclark: We tried this out and were running into infinite rerenders. Just noting we can probably come back to this if performance is a huge issue with this component.

PR #482, opened 2020-12-10
fix(component): ensure autocomplete works on Select component

bc-sumi:
## What?
Ensure autocomplete works on Select component. 
Screen Shot 2020-12-10 at 3 57 58 PM

## Why?
The Select component is currently using `autoComplete: 'no'`, when it should be `autoComplete: 'off'`.
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

### DIFF
--- a/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
@@ -273,7 +273,7 @@ export const MultiSelect = typedMemo(
               <Input
                 {...rest}
                 {...getInputProps({
-                  autoComplete: 'no',
+                  autoComplete: 'off',
                   disabled,
                   onFocus: () => {
                     !isOpen && openMenu();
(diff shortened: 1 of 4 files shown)


chanceaclark (approved):
Looks good! Thanks for the fix!

jorgemoya:
Why do I remember we had as `off` at some point? Looks good, but can you change the `autoComplete` attribute in `MultiSelect` as well?

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -307,7 +307,7 @@ export const MultiSelect = typedMemo(
                   label: option.content,
                   onDelete: () => removeItem(option),
                 }))}
-                autoComplete="no"
+                autoComplete="off"
```
jorgemoya, line 310: I think this line was added by accident. Should be fine if we remove, can you test? Sorry about it!

jorgemoya (approved):

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
bc-sumi: @jorgemoya Should I remove the autoComplete completely?

jorgemoya: This line, yes

jordan-massingill (approved):

PR #492, opened 2021-01-08
chore(ci): split workflows

deini:


deini:
Closing this, we don't really gain much. To run our lint / tests we first need to build the libraries. No real benefit in splitting them.

PR #495, opened 2021-01-14
feat(component): Add drag and drop support to stateful table component

### LINKED ISSUE #426: Feature Request: Draggable Rows on Stateful Table
The current UI for tables allows for draggable rows but the Stateful Table does not support this feature.

I would like to have draggable rows part of Stateful Table like the ones shown here. 
https://www.figma.com/file/Rj63WSTZQc2H4XI9Oi3r5o/MSF-design-(LATEST)?node-id=10702%3A0

Screen Shot 2020-07-29 at 9 35 54 PM

animesh1987:
## What
Add drag and drop support to `StatefulTable` component.

## Why
[Feature request for Drag and Drop](https://github.com/bigcommerce/big-design/issues/426)

Closes #426.

ping @bigcommerce/frontend @deini @chanceaclark

### DIFF
--- a/packages/big-design/src/components/StatefulTable/StatefulTable.tsx
+++ b/packages/big-design/src/components/StatefulTable/StatefulTable.tsx
@@ -16,12 +16,20 @@ export interface StatefulTableColumn<T> extends Omit<TableColumn<T>, 'isSortable
 export interface StatefulTableProps<T>
   extends Omit<TableProps<T>, 'columns' | 'pagination' | 'selectable' | 'sortable'> {
   columns: Array<StatefulTableColumn<T>>;
-  dragAndDrop?: boolean;
+  draggable?: boolean;
   pagination?: boolean;
   selectable?: boolean;
   defaultSelected?: T[];
   onSelectionChange?: TableSelectable<T>['onSelectionChange'];
 }
+
+const swapArrayElements = <T extends unknown>(array: T[], from: number, to: number) => {
+  if (from > to) {
+    return [...array.slice(0, to), array[from], ...array.slice(to, from), ...array.slice(from + 1, array.length)];
+  } else {
+    return [...array.slice(0, from), ...array.slice(from + 1, to), array[from], ...array.slice(to, array.length)];
+  }
+};
 
 const InternalStatefulTable = <T extends TableItem>({
   columns = [],
@@ -30,7 +38,7 @@ const InternalStatefulTable = <T extends TableItem>({
   items = [],
   keyField,
   onSelectionChange,
-  dragAndDrop = false,
+  draggable = false,
   pagination = false,
   selectable = false,
   stickyHeader = false,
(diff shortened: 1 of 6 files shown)


packages/big-design/src/components/StatefulTable/StatefulTable.tsx
```diff
@@ -86,6 +88,16 @@ const InternalStatefulTable = <T extends TableItem>({
     onSort,
   ]);
 
+  const onDragEnd = useCallback(
+    (from: number, to: number) => {
+      const itemsClone = [...items];
+      const item = itemsClone.splice(from, 1);
+      itemsClone.splice(to, 0, ...item);
```
chanceaclark, line 95: 🍹 Could we potentially change this to not rely on mutating an array? Maybe with a combination of `slice` and `spread` operations?

packages/docs/PropTables/StatefulTablePropTable.tsx
```diff
@@ -19,6 +19,11 @@ const statefulTableProps: Prop[] = [
     description: 'The array of items to display.',
     required: true,
   },
+  {
+    name: 'dragAndDrop',
```
chanceaclark, line 23: 🍹 Just want to have a discussion about this prop. Should this potentially be something different? Maybe just `draggable`?

chanceaclark (commented):
A few things to point out:

- How do know what the new order is when a user has changed it? Seems like we might need a callback.
- Looking at the `Table` documentation, doesn't seem like the new drag and drop props listed. Could you add those in this PR?
- In the **Usage with drag and drop** example in the `Table` documentation, it has some erroneous keys, could we potentially remove `keyField`, `itemName`, and any `isSortable` key/vals to clean up the example?

chanceaclark:
Hey @animesh1987, everything looks great, but could you address these comments?

> * How do know what the new order is when a user has changed it? Seems like we might need a callback.
> * Looking at the `Table` documentation, doesn't seem like the new drag and drop props listed. Could you add those in this PR?
> * In the **Usage with drag and drop** example in the `Table` documentation, it has some erroneous keys, could we potentially remove `keyField`, `itemName`, and any `isSortable` key/vals to clean up the example?

packages/big-design/src/components/StatefulTable/StatefulTable.tsx
```diff
@@ -14,21 +14,36 @@ export interface StatefulTableColumn<T> extends Omit<TableColumn<T>, 'isSortable
 }
 
 export interface StatefulTableProps<T>
-  extends Omit<TableProps<T>, 'columns' | 'pagination' | 'selectable' | 'sortable'> {
+  extends Omit<TableProps<T>, 'columns' | 'pagination' | 'selectable' | 'sortable' | 'onRowDrop'> {
   columns: Array<StatefulTableColumn<T>>;
   pagination?: boolean;
   selectable?: boolean;
   defaultSelected?: T[];
+  onRowDrop?(items: T[]): void;
```
chanceaclark, line 22: @deini what do you think of this approach?

deini: The approach is similar to `onSelectionChange`, should be ok 👍

chanceaclark (approved):
LGTM!

PR #497, opened 2021-01-27
fix(component): remove tabindex and role from Button

chanceaclark:
## What

Removes the `tabIndex` and `role` props from `Button` components.

## Why

`<button>` elements already have a default `tabindex="0"` and `role="button"` applied to them via the user-agent. These are unnecessary and can be omitted.

### DIFF
--- a/packages/big-design/src/components/Alert/__snapshots__/spec.tsx.snap
+++ b/packages/big-design/src/components/Alert/__snapshots__/spec.tsx.snap
@@ -646,8 +646,6 @@ exports[`renders close button 1`] = `
   >
     <button
       class="c7 c8"
-      role="button"
-      tabindex="0"
     >
       <span
         class="c9"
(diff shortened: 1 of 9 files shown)


exeDog (approved):

jorgemoya:
Does this occur in all browsers? Is there a site with this info?

jordan-massingill (approved):
Good call. Wonder why these were initially added 🤔

chanceaclark:
> Does this occur in all browsers? Is there a site with this info?

For `role` see **Base Concept** under _Characteristics_ https://www.w3.org/TR/wai-aria-1.1/#button tl'dr it inherits `HTML Button`

For `tabindex`, see **If the value is null** https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex There should be a list of elements that the user-agents should be considered "sequentially focusable".

PR #499, opened 2021-01-27
feat(component): add headerId prop to Panel

chanceaclark:
## What

Adds `headerId` as a prop to `Panel` components.

## Why

Gives users the ability to add an `id` attribute to the heading. Useful in situation where you want to use URL hashes to link to `Panel` headings, en lieu of using the `id` on the `Panel` itself.


jordan-massingill (approved):
👍

jorgemoya (approved):
Good idea

deini (approved):

PR #504, opened 2021-02-12
feat(theme): add currentColor, inherit, and transparent props to colors

jorgemoya:
Add inherit and transparent props to colors object for use in BD components.

### DIFF
--- a/packages/big-design-theme/src/system/colors.ts
+++ b/packages/big-design-theme/src/system/colors.ts
@@ -1,4 +1,8 @@
 export interface Colors {
+  currentColor: string;
+  inherit: string;
+  transparent: string;
+
   white: string;
   brand: string;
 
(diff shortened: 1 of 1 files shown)


chanceaclark:
Can we add a description for this PR?

jorgemoya:
> Can we add a description for this PR?

Done!

packages/big-design-theme/src/system/colors.ts
```diff
@@ -49,6 +52,9 @@ export interface Colors {
 }
 
 export const colors: Colors = {
+  inherit: 'inherit',
+  transparent: 'transparent',
```
chanceaclark, line 56: 🍹 What do you think about adding `currentColor` as well?

chanceaclark (approved):

packages/big-design-theme/src/system/colors.ts
jorgemoya: Might as well

PR #507, opened 2021-03-05
fix: hide arrow if Node children is an empty array

Meztimuro:
## What
Don't show arrow if `TreeNode` children prop is an empty array

## Why
Now TreeNode with an empty array as children is displayed with an arrow, and when node is "expanded" no sub-nodes are displayed and it may seem like a bug

First Node has an empty array as a children prop, second has no children prop, third has a child
image

## Proof
Same Nodes with my proposed changes
image

### DIFF
--- a/packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
+++ b/packages/big-design/src/components/Tree/TreeNode/TreeNode.tsx
@@ -133,7 +133,7 @@ const InternalTreeNode = <T extends unknown>({
 
   const renderedChildren = useMemo(
     () =>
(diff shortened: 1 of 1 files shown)


chanceaclark:
The original intention for checking for length is if a node potentially had children that we need to asynchronously fetch. I believe we removed that functionality when we changed the exposed component from `Tree` -> `StatefulTree`. I'm in favor of **not** removing this, since we may need it later. We have plans to expose the `Tree` component when the need to asynchronously fetch nodes comes up.

I recommend just removing `children` from your tree if it's empty for now.

PR #509, opened 2021-03-08
chore(deps): support React v17

chanceaclark:
## What

Adds React@17 support and updates related dependencies.

### Other notes:
- Added missing `@babel/plugin-proposal-nullish-coalescing-operator` into devDeps.
- With the `react-datepicker` and `react-popper` version update, `fireEvent.focus()` and `fireEvent.blur()` support stopped working so I changed them over to `fireEvent.click()` which made a bit more sense.
- Alongside that issue ^ I had to remove some of our blur/focus tests. RTL proposes [another alternative](https://testing-library.com/docs/guide-events/#focusblur) to use but for now I would like to leave out those tests till there's something a bit better.

## Why

To the future and beyond! Also because our React application are now starting to use React@17.


packages/big-design/src/components/Select/spec.tsx
```diff
@@ -223,18 +222,6 @@ test('input button has aria-label', async () => {
...
-  const input = getByTestId('select');
-
-  expect(queryByRole('listbox')).toBeEmptyDOMElement();
-
-  fireEvent.focus(input);
-
-  expect(queryByRole('listbox')).not.toBeEmptyDOMElement();
-  await waitForElement(() => getByRole('option', { name: /mex/i }));
-});
-
```
jorgemoya, line 237: I think we should keep this one and just change wording to "click" and use `click()`.

jorgemoya (commented):
Thanks for updating this. 👍

packages/big-design/src/components/Datepicker/spec.tsx
```diff
@@ -65,9 +65,7 @@ test('calls onDateChange function when a date cell is clicked', () => {
 
   const input = container.querySelector('input');
 
-  act(() => {
-    fireEvent.focus(input as HTMLInputElement);
-  });
+  fireEvent.click(input as HTMLInputElement);
```
jordan-massingill, line 68: Why `fireEvent` instead of `userEvent`?

packages/big-design/src/components/Select/spec.tsx
chanceaclark: Good idea

jorgemoya (approved):

chibicode:
@chanceaclark would you be able to cut a new release for this? Thank you 🙏

chanceaclark:
Hey @chibicode! Just getting back from vacation. Let me get caught up on a few things and I'll see about cutting a release soon.

chanceaclark:
@chibicode just released a new patch. Try `@bigcommerce/big-design@0.29.1` for `react@17` support!

PR #510, opened 2021-03-10
feat(component): add display prop to TableColumns

### LINKED ISSUE #496: Add responsive display props for Table Columns
It would be nice to have props to control the responsive  display of the table columns in the Table/StatefulTable Components. Currently have added styling to control this (mobile, tablet, desktop responsive display), but adding props into big-design would be nice.  

Table/StatefulTable Components props could be able to control the display of the responsive table column display.

**Describe alternatives you've considered**
Currently have added styling around the Table Component to control the responsive display of the colums - 

```
export const StyledTable = styled.div`
    tr td:nth-child(3) {
        display: none;
    }
    tr th:nth-child(3) {
        display: none;
    }
    word-break: break-word;
    ${({ theme }) => theme.breakpoints.tablet} {
        tr td:nth-child(3) {
            display: table-cell;
        }
        tr th:nth-child(3) {
            display: table-cell;
        }
    }
`;
``` 
**Additional context**
CC: @chanceaclark

jorgemoya:
Allows responsive prop in `TableColumns` to hide/show columns responsively.

Fixes #496

## Screenshots

Third column hides/show responsively.

Screen Shot 2021-03-10 at 12 27 37 PM

Screen Shot 2021-03-10 at 12 27 44 PM

ping @bigcommerce/frontend

### DIFF
--- a/packages/big-design/src/components/Table/index.ts
+++ b/packages/big-design/src/components/Table/index.ts
@@ -1,3 +1,2 @@
 export { Table, TableFigure } from './Table';
-export * from './mixins';
 export * from './types';
--- a/packages/big-design/src/components/Table/index.ts
+++ b/packages/big-design/src/components/Table/index.ts
@@ -1,2 +1,3 @@
 export { Table, TableFigure } from './Table';
+export * from './mixins';
 export * from './types';
--- a/packages/big-design/src/components/Table/mixins/display/display.tsx
+++ b/packages/big-design/src/components/Table/mixins/display/display.tsx
@@ -1,13 +1,13 @@
 import { Breakpoints, breakpointsOrder, ThemeInterface } from '@bigcommerce/big-design-theme';
 import { css, FlattenSimpleInterpolation } from 'styled-components';
 
-import { DisplayOverload, DisplayProps } from './types';
+import { TableColumnDisplayOverload, TableColumnDisplayProps } from './types';
 
-export const withDisplay = () => css<DisplayProps>`
+export const withTableColumnDisplay = () => css<TableColumnDisplayProps>`
   ${({ display, theme }) => display && getDisplayStyles(display, theme, 'display')};
 `;
 
-const getDisplayStyles: DisplayOverload = (
+const getDisplayStyles: TableColumnDisplayOverload = (
   displayProp: any,
   theme: ThemeInterface,
   cssKey: any,
@@ -27,7 +27,7 @@ const getSimpleDisplay = (displayProp: string | number, cssKey: string): Flatten
   ${cssKey}: ${displayProp}
 `;
 
-const getResponsiveDisplay: DisplayOverload = (
+const getResponsiveDisplay: TableColumnDisplayOverload = (
   displayProp: any,
   theme: ThemeInterface,
   cssKey: string,
--- a/packages/big-design/src/components/StatefulTable/StatefulTable.tsx
+++ b/packages/big-design/src/components/StatefulTable/StatefulTable.tsx
@@ -110,7 +110,7 @@ const InternalStatefulTable = <T extends TableItem>({
         onRowDrop(updatedItems);
       }
     },
-    [onRowDrop, state.currentItems, pagination],
+    [state.currentItems, onRowDrop, pagination],
   );
 
   return (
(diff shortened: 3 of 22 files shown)


packages/big-design/src/components/Table/index.ts
```diff
@@ -1,2 +1,3 @@
 export { Table, TableFigure } from './Table';
+export * from './mixins';
```
jorgemoya, line 2: Do I need to expose this?

chanceaclark (approved):
Looks great, minus that export!

packages/big-design/src/components/Table/mixins/display/display.tsx
```diff
@@ -0,0 +1,49 @@
...
+  if (typeof displayProp === 'object') {
+    return getResponsiveDisplay(displayProp, theme, cssKey);
+  }
+
+  if (typeof displayProp === 'string' || typeof displayProp === 'number') {
+    return getSimpleDisplay(displayProp, cssKey);
+  }
+
+  return [];
+};
```
golcinho, line 24: Any chance to avoid `any` type here? 

I also see many places around this file with `any` type. Could you please check all the same cases?

Also, what if `displayProp` will be an array? Code below will be broken.
```
if (typeof displayProp === 'object') {
    return getResponsiveDisplay(displayProp, theme, cssKey);
}
```

golcinho: Also, I see that most of those mixins already exists in `big-design/src/mixins/display`, can we re-use them?

jorgemoya: Originally I had the code reusing the mixins, the reason why I chose to duplicate the logic was for readability. Reusing the mixins means adding more type variants, and I got concerned that this felt like a snowflake scenario overall. But if we feel strongly against this I can certainly modify it.

cc @chanceaclark

chanceaclark: The use-case for this mixin is `Table`-specific, so I'd rather keep it private and close to it's usage.

golcinho: @chanceaclark @jorgemoya Thank you for your answers 🙇 

I'm not against your vision but, I just want to explain my position and understand your idea more clear.

I will start from `mixin`, IMO I'm not sure that duplicating is the correct way... We have two concepts of mixins: from OOP and SASS(or other pre-processor, in short, mixin concept that relates to styles) and the main idea of both is `reuse` so, technically people who will look at this code will be confused cause mixin should be reusable and not be duplicated.

I'm not sure that these mixins are `table-specific` cause they have completely the same implementations, the only difference is types(but this problem could be resolved easily). Also, I think we need to follow the base `DRY` principle.

Also, if we follow the rule to keep mixins close to it's usage, why we keeping general mixins in `src/mixins/display/display.tsx`?

I think that we need to reuse generic mixins and create a `table-specific` mixins only in case when they should really specific for the table component and not for any other.

Please, understand me correctly, I'm not trying to start a holly-war here 😄 I'm important to explain my position and get your thoughts about it 🤝

jorgemoya: You bring good points. I agree that the only issue was types, and this can certainly be resolved. I can see your point of `mixins` being, by concept, reusable, which is why I actually first started my implementation that way.

However, the idea behind duplicating it to the table was so that we don't add any extra layer of logic (or types) to the existing one. I honestly don't see anything else using this duplicated `mixin`, or having to snowflake another use case for the display `mixin`, so I wanted to keep it separate. 

I'm not against going back to adding the `types` to the general `mixin`, I can certainly fix that quick. At this point I'm okay with both approaches, it really depends on what you guys think.

chanceaclark: We did spike on making these "mixins" (responsive properties) a little better/DRYer, but TypeScript makes it difficult and would add layers of hard to understand complexity. I'm 100% okay with things not being DRY because coupling things may lead to issues in the future and for readability/complexity sake.

However, I'm okay changing the terminology of the `mixins` folder/files since it's not descriptive of what's in there. Let's pull this conversation into a gh-issue. Seems like it would need another PR if we wanted to update it.

golcinho: Thank you for your answer and thoughts! 🙇 

I apologize for delaying the delivery of this feature

✅

golcinho (approved):

PR #516, opened 2021-04-05
feat(component): add Worksheet component

jorgemoya:
## Beautiful Preview
Here it is people (or at least the foundation).

Be amazed by the beauty:

Kapture 2021-04-08 at 11 26 57

Wow very nice.

## What

The idea of this PR is to review the foundation for the Worksheet component but without exposing/releasing it to the public. This is due to the complexity of this component. There are still a lot of features we need to implement before public release.

Proposed API:
```ts
const columns: WorksheetColumn[] = [
  { hash: 'productName', header: 'Product name' },
  { hash: 'otherField', header: 'Other field', type: 'number', validation: (value) => value !== 2 },
];

const items = [
  {
    productName: 'Epic shoes',
    otherField: 1,
  },

  {
    productName: 'Not so nice shoes',
    otherField: 2,
  },
]

<Worksheet columns={columns} items={items} onChange={handleChange} onErrors={handleErrors} />;
```

## How
I'm using the [zustand](https://github.com/pmndrs/zustand) library to handle state to prevent unnecessary rerendering.

I'm trying to think ahead and many of the functions I build support multiple cell parameters even thought we're currently constrained to a single cell selection.

## Features

- Only Text/Number editor for now
- Cell visual selection
- Validation
- Returns edited cells on change
- Returns invalid cells on first pass and after a change

## Notes

- As you can see, there are a lot of missing features expected from a Worksheet component (click-n-drag cells, select all, undo/redo, row/column select, adjustable column width, etc...), but it is WIP in progress.
- No keyboard listeners other than enter and escape when editing a cell.
- No mouse multi selection.
- I haven't thought about accessibility.
- For launch we definitely need at least two other field types `select` & `checkbox`.
- Virtualization is planned.
- No docs.


packages/big-design/src/components/Worksheet/hooks/spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+//   // beforeEach(() => {
+//   //   jest.unmock('./useUpdateItems');
+//   // });
+
+//   test('throws if is not wrapped in UpdateItemsProvider', () => {
+//     const { result } = renderHook(() => useUpdateItems());
+
+//     expect(result.error.message).toContain('UpdateItemsProvider');
+//   });
+// });
```
jorgemoya, line 109: I couldn't figure out how to add this test since I mock `useUpdateItems` for the other component above.

packages/big-design/src/components/Worksheet/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  items: T[];
+  onChange(items: Cell[]): void;
+  onErrors?(items: Cell[]): void;
+}
+
+export interface Cell {
+  columnIndex: number;
+  hash: string;
+  rowIndex: number;
+  value: string | number;
```
chanceaclark, line 19: Let's see if we can infer this type through a generic. I took a quick look, and you might explore this:
```ts
interface Product {
    productName: string;
    categories: string;
    otherField: string;
    otherField2: string;
    otherField3: number;
}

interface Worksheet<Item> {
    items: Item[];
    onChange(items: Array<Cell<Item[keyof Item]>>): void;
}

interface Cell<Value> {
    value: Value;
}

const ws: Worksheet<Product> = {
    items: [],
    onChange: items => {
        console.log(items);
    },
};
```
Screen Shot 2021-04-05 at 16 09 41

packages/big-design/src/components/Worksheet/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { Cell } from './types';
+
+export const mergeCells = (oldCells: Cell[], newCells: Cell[]) =>
+  newCells.reduce(
+    (accum, cell) => {
+      const index = oldCells.findIndex((c) => c.rowIndex === cell.rowIndex && c.columnIndex === cell.columnIndex);
```
chanceaclark, line 6: Same with above, could we make `c` named more developer friendly?

packages/big-design/src/components/Worksheet/styled.tsx
```diff
@@ -0,0 +1,22 @@
...
+  table-layout: fixed;
+  width: 100%;
+`;
+
+export const Header = styled.th`
+  background-color: ${({ theme }) => theme.colors.secondary10};
+  border: ${({ theme }) => `${remCalc(0.5)} solid ${theme.colors.secondary30}`};
+  box-sizing: border-box;
+  color: ${({ theme }) => theme.colors.secondary60};
+  font-weight: 600; //TODO: fix?
```
chanceaclark, line 16: `semiBold`?

packages/big-design/src/components/Worksheet/styled.tsx
```diff
@@ -0,0 +1,22 @@
+import { theme as defaultTheme, remCalc } from '@bigcommerce/big-design-theme';
```
chanceaclark, line 1: I think we should be trying to use `remCalc` from the `theme` object instead of importing it 🤔  Did you see any other usage like this?

chanceaclark (commented):
Initial review, once we get it in a gucci state, I'll ♻️  again.

packages/big-design/src/components/Worksheet/RowStatus/RowStatus.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { memo, useMemo } from 'react';
+
+import { useStore } from '../hooks';
+
+import { Status } from './styled';
+
+export interface RowStatusProps {
+  rowIndex: number;
+}
```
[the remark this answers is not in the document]
jorgemoya: In this example I dont need to `export` it anymore (fixed), but do you mean move all types to the parent `./types` file?

My idea was to inline `ComponentProps` or local types and add only "global" type to `/Worksheet/types.ts` like `Cell`.

packages/big-design/src/components/Worksheet/editors/TextEditor/TextEditor.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { EditableCellKeyDown } from '../../hooks';
+import { Cell } from '../../types';
+
+import { StyledInput } from './styled';
+
+export interface TextEditorProps {
+  cell: Cell;
+  handleBlur(): void;
+  handleKeyDown: EditableCellKeyDown;
```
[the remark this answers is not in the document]
jorgemoya: You mean change them to `onBlur` and `onKeyDown`?

packages/big-design/src/components/Worksheet/editors/TextEditor/styled.ts
```diff
@@ -0,0 +1,27 @@
...
+  line-height: ${({ theme }) => theme.lineHeight.small};
+
+  &:focus {
+    outline: none;
+  }
+
+  ${({ isEdited }) =>
+    isEdited &&
+    css`
+      background-color: ${({ theme }) => theme.colors.warning20};
```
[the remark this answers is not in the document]
jorgemoya: What designs you looking at? The ones I see at this [figma](https://www.figma.com/file/zcJRWjXwgLGRcpxKrdFkwZ/Starter-App-Outline?node-id=1994%3A11015) file has them as warning20.

packages/big-design/src/components/Worksheet/styled.tsx
jorgemoya: Honestly I don't remember using it from the `theme` object in BigDesign before, I might need to take a look at other components at a later time.

packages/big-design/src/components/Worksheet/types.ts
jorgemoya: This is super gucci but the problem I see is that I use the `Cell` type in a lot of places. What would be a good solution for that?

packages/big-design/src/components/Worksheet/RowStatus/RowStatus.tsx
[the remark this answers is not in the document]
chanceaclark: >do you mean move all types to the parent ./types file?

Sort of yeah. With the Tree component types, I sort of mix-n-matched where I threw them. I don't know what the best approach is, but was wondering your thoughts on the matter?

packages/big-design/src/components/Worksheet/editors/TextEditor/TextEditor.tsx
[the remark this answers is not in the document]
chanceaclark: Yep!

packages/big-design/src/components/Worksheet/types.ts
chanceaclark: Yeah, `unknown` would be the best bet. Is there a problem with that approach?

packages/big-design/src/components/Worksheet/RowStatus/RowStatus.tsx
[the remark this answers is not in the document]
jorgemoya: I like the idea of keeping `ComponentProps` types inline and the rest in `./types.ts`

packages/big-design/src/components/Worksheet/hooks/spec.tsx
jorgemoya: I feel like this test isn't really necessary. Do you feel strongly we need it?

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const { handleDoubleClick, handleBlur, handleKeyDown, isEditing, Editor } = useEditableCell(cell);
+  const setSelectedRows = useStore(useMemo(() => (state) => state.setSelectedRows, []));
+  const setSelectedCells = useStore(useMemo(() => (state) => state.setSelectedCells, []));
+  const addInvalidCells = useStore(useMemo(() => (state) => state.addInvalidCells, []));
+  const removeInvalidCells = useStore(useMemo(() => (state) => state.removeInvalidCells, []));
+
+  const isSelected = useStore(
+    useMemo(
+      () => (state) =>
+        state.selectedCells.reduce(
```
jordan-massingill, line 31: I like this use of reduce, but jw is there a reason not to use `Array.some()` here instead?

```suggestion
        state.selectedCells.some(selectedCell => selectedCell.columnIndex === cell.columnIndex && selectedCell.rowIndex === cell.rowIndex)
```

packages/big-design/src/components/Worksheet/RowStatus/RowStatus.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface RowStatusProps {
+  rowIndex: number;
+}
+
+export const RowStatus: React.FC<RowStatusProps> = memo(({ rowIndex }) => {
+  const isSelected = useStore(useMemo(() => (state) => state.selectedRows.includes(rowIndex), [rowIndex]));
+
+  const isInvalid = useStore(
+    useMemo(() => (state) => Boolean(state.invalidCells.find((cell) => cell.rowIndex === rowIndex)), [rowIndex]),
```
jordan-massingill, line 15: 🍹  This line may benefit from a `.some()` as well since it already returns a boolean value.

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
jorgemoya: Good call

packages/big-design/src/components/Worksheet/Worksheet.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }, [editedCells, onChange]);
+
+  useEffect(() => {
+    if (typeof onErrors === 'function' && invalidCells.length) {
+      onErrors(invalidCells);
+    }
+  }, [invalidCells, onErrors]);
+
+  const renderHeaders = useMemo(
```
rtalvarez, line 39: 🍹 nitpicky comment, since this const holds an actual value and not a callback, i'd suggest renaming this to `renderedHeaders` 

(same for below)

packages/big-design/src/components/Worksheet/hooks/spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  test('enter', () => {
+    const { result } = renderHook(() => useEditableCell(cell));
+
+    act(() => {
+      result.current.handleDoubleClick();
+      result.current.handleKeyDown(
+        ({
+          currentTarget: 1,
```
rtalvarez, line 54: curious, what are these `currentTarget` and `target` fields for in a React.KeyboardEvent? Can we just not supply them?

Also, can we not construct this object using the native API `new KeyboardEvent('Foo', { key: 'Enter' })` ?

packages/big-design/src/components/Worksheet/styled.tsx
```diff
@@ -0,0 +1,22 @@
...
+export const Table = styled.table`
+  border-collapse: collapse;
+  border-spacing: 0;
+  table-layout: fixed;
+  width: 100%;
+`;
+
+export const Header = styled.th`
+  background-color: ${({ theme }) => theme.colors.secondary10};
+  border: ${({ theme }) => `${theme.helpers.remCalc(0.5)} solid ${theme.colors.secondary30}`};
```
rtalvarez, line 13: 🍹  I wonder if we need a "tableBorder" internal theme prop or something to provide this styling, it'll be easier to enforce, maintain and use

rtalvarez (commented):
looking good

packages/big-design/src/components/Worksheet/hooks/spec.tsx
jorgemoya: To be honest it's a copy pasta from another test. I can simply remove them since they are not relevant to current test, good catch. I couldn't implement your solution, can you expand on it?

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { Cell as TCell, WorksheetColumn } from '../types';
+
+import { StyledCell } from './styled';
+
+interface CellProps<Value> extends TCell<Value> {
+  validation: WorksheetColumn<Value>['validation'];
+}
+
+// We can't know what value type a cell has
+export const Cell: React.FC<CellProps<any>> = memo(({ columnIndex, hash, rowIndex, type, value, validation }) => {
```
chanceaclark, line 14: You can still use generics here:
```suggestion
const InternalCell = <T extends unknown>({
  columnIndex,
  hash,
  rowIndex,
  type,
  value,
  validation,
}: CellProps<T>): React.ReactElement<CellProps<T>> => {
```

And the export will be:
```tsx
export const Cell = typedMemo(InternalCell);
```

I would still try and type these with generics if you can.

packages/big-design/src/components/Worksheet/Row/Row.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useStore } from '../hooks';
+import { RowStatus } from '../RowStatus';
+import { WorksheetColumn } from '../types';
+
+interface RowProps {
+  columns: WorksheetColumn<unknown>[];
+  rowIndex: number;
+}
+
+export const Row: React.FC<RowProps> = memo(({ columns, rowIndex }) => {
```
chanceaclark, line 13: Still think you can use generics here too.

packages/big-design/src/components/Worksheet/Row/Row.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <tr>
+      <RowStatus rowIndex={rowIndex} />
+      {columns.map((column, columnIndex) => (
+        <Cell
+          columnIndex={columnIndex}
+          hash={column.hash}
+          key={`${rowIndex}-${columnIndex}`}
+          rowIndex={rowIndex}
+          type={column.type || 'text'}
```
chanceaclark, line 25:
```suggestion
          type={column.type ?? 'text'}
```

packages/big-design/src/components/Worksheet/hooks/spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+let mockUpdateItems = jest.fn();
+jest.mock('./useUpdateItems', () => ({
+  useUpdateItems: () => ({
+    updateItems: mockUpdateItems,
+  }),
+}));
+
+describe('useEditableCells', () => {
+  const cell: Cell<unknown> = { rowIndex: 0, columnIndex: 0, type: 'text', hash: 'category', value: 'Text' };
```
chanceaclark, line 17: Shouldn't this be `string`?
```suggestion
  const cell: Cell<string> = { rowIndex: 0, columnIndex: 0, type: 'text', hash: 'category', value: 'Text' };
```

chanceaclark (commented):
Few more comments but looks really good man

packages/big-design/src/components/Worksheet/hooks/spec.tsx
chanceaclark: Eh, get rid of it then.

jorgemoya:
Per offline feedback, I have now added a check to only pass value of type `string | number` to `Cells`. So I updated the types to reflect this change.

Most noticeably, `validation` now should always return a `string | number` as the type of the value prop.

Which now makes me think if instead of adding `generics`, should we just expect the value to be `string | number` since we enforce it now anyway?

chanceaclark (approved):
Looks good to me! Great job on this!

PR #520, opened 2021-04-14
feat(component): implements selectable column type

jorgemoya:
## What
Implements selectable column type

## Why
Ongoing development of Worksheet component.

## Known issues
- When clicking on the input directly, there is a small delay on open/closed state change of that component. This can cause an overlap when opening sibling selects. There is no issue if we click on the dropdown button.
- Because of the font resize, the ✔️ icon can be too big, especially when I set the menu as `autoWidth`. Since width is not auto I'm ignoring for now but we might need to revise if we choose to go with that option a later at point.

## Beautiful stuff
Kapture 2021-04-14 at 11 16 48

### DIFF
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -2,24 +2,27 @@ import React, { useCallback, useEffect, useMemo } from 'react';
 
 import { typedMemo } from '../../../utils';
 import { Small } from '../../Typography';
+import { SelectEditor, TextEditor } from '../editors';
 import { useEditableCell, useStore } from '../hooks';
 import { Cell as TCell, WorksheetColumn, WorksheetItem } from '../types';
 
 import { StyledCell } from './styled';
 
 interface CellProps<Item> extends TCell<Item> {
+  options: WorksheetColumn<Item>['options'];
   validation: WorksheetColumn<Item>['validation'];
 }
 
 const InternalCell = <T extends WorksheetItem>({
   columnIndex,
   hash,
+  options,
   rowIndex,
   type,
   value,
   validation,
 }: CellProps<T>) => {
-  const cell = useMemo(() => ({ columnIndex, hash, rowIndex, type, value }), [
+  const cell: TCell<T> = useMemo(() => ({ columnIndex, hash, rowIndex, type, value }), [
     columnIndex,
     hash,
     rowIndex,
@@ -27,7 +30,7 @@ const InternalCell = <T extends WorksheetItem>({
     value,
   ]);
 
-  const { Editor, handleBlur, handleDoubleClick, handleKeyDown, isEditing } = useEditableCell<T>(cell);
+  const { handleBlur, handleChange, handleDoubleClick, handleKeyDown, isEditing } = useEditableCell<T>(cell);
   const setSelectedRows = useStore((state) => state.setSelectedRows);
   const setSelectedCells = useStore((state) => state.setSelectedCells);
   const addInvalidCells = useStore((state) => state.addInvalidCells);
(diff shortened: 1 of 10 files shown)


packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -85,19 +95,23 @@ let handleErrors = jest.fn();
...
+  });
 });
 
-test('renders worksheet', () => {
+test('renders worksheet', async () => {
   const { container } = render(<Worksheet columns={columns} items={items} onChange={handleChange} />);
 
   expect(container.firstChild).toMatchSnapshot();
+
+  await waitForElement(() => screen.getAllByRole('combobox'));
```
jordan-massingill, line 110: 🤔 What is this line for?

jorgemoya: From what I've gathered, the Select component performs some state changes when rendering so we actually need to wait for it to be done with the updates else we get an error due to state changes not wrapped in an act.

jordan-massingill: 🧠 TIL - thanks 😃

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
```diff
@@ -82,15 +85,20 @@ const InternalCell = <T extends WorksheetItem>({
...
+      // Default to TextEditor
+      default:
+        return isEditing ? (
+          <TextEditor cell={cell} isEdited={isEdited} onBlur={handleBlur} onKeyDown={handleKeyDown} />
+        ) : (
+          // In case of NaN casting to string
+          <Small color="secondary70">{value.toString()}</Small>
+        );
+    }
+  }, [cell, handleBlur, handleChange, handleKeyDown, isEdited, isEditing, options, type, value]);
```
jorgemoya, line 101: The previous logic of returning an `Editor` proved to be too tricky for Typescript and/or wasn't the right approach. I moved the switch to the `Cell` for a more transparent approach, and it is typed correctly.

packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
```diff
@@ -0,0 +1,31 @@
...
+      isEdited &&
+      css`
+        background-color: ${({ theme }) => theme.colors.warning10};
+      `}
+  }
+
+  [role='option'] {
+    font-size: ${({ theme }) => theme.typography.fontSize.small};
+  }
+`;
```
jorgemoya, line 29: This approach of overwriting styles saved me the complexity of allowing the `Select` custom input components. Since this is the only scenario were we want a different input, it made sense to me. Thoughts?

packages/big-design/src/components/Worksheet/types.ts
```diff
@@ -8,7 +10,8 @@ export interface Worksheet<Item extends WorksheetItem> {
 export interface WorksheetColumn<Item> {
   hash: keyof Item;
   header: string;
-  type?: 'text' | 'number';
+  options?: SelectOption<unknown>[];
+  type?: 'text' | 'number' | 'select';
```
jorgemoya, line 14: Should I try to split this type into two since ideally we only need `options` when `type === select`? What would be the best approach?

chanceaclark: Yeah, may be better to split the interface into two, maybe `WorksheetTextColumn` & `WorksheetSelectableColumn`? Thoughts?

packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
chanceaclark: lgtm

packages/big-design/src/components/Worksheet/types.ts
jorgemoya: That is what I had in mind, let me try it out.

packages/big-design/src/components/Worksheet/Row/Row.tsx
```diff
@@ -22,6 +22,7 @@ const InternalRow = <T extends WorksheetItem>({ columns, rowIndex }: RowProps<T>
           columnIndex={columnIndex}
           hash={column.hash}
           key={`${rowIndex}-${columnIndex}`}
+          options={column.type === 'select' ? column.options : []}
```
chanceaclark, line 25: Could we potentially allow `options` to be `undefined`? We default it here https://github.com/bigcommerce/big-design/pull/520/files#diff-89a2998a7d4275373c1d9b534a177558d58446e282bc6c05197f9dead4bd8dd7R20

chanceaclark (approved):
Looks good to me, though I have one slight comment. Let me know what you think.

packages/big-design/src/components/Worksheet/Row/Row.tsx
jorgemoya: I will change it to be allow `undefined` along the way and let the default handle the scenario of an empty `options` array.

PR #523, opened 2021-04-21

golcinho:
## What?

## Screenshots & Video
**Buttons:**
Screenshot 2021-04-21 at 11 22 18
Screenshot 2021-04-21 at 11 22 26
Screenshot 2021-04-21 at 11 22 37

**Icon only:**
Screenshot 2021-04-21 at 11 22 56

**Dropdown:**
Screenshot 2021-04-21 at 11 23 47

**Docs:**
Screenshot 2021-05-20 at 13 28 12

## Proof
Unit test + local testing.

### DIFF
--- a/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
@@ -1,0 +1,71 @@
+import React, { ButtonHTMLAttributes, Fragment, HTMLAttributes, memo, ReactNode, useMemo } from 'react';
+
+import { MarginProps } from '../../mixins';
+import { Dropdown, DropdownProps } from '../Dropdown';
+import { Flex } from '../Flex';
+
+import { StyledButton } from './styled';
+
+export interface ButtonGroupItem extends ButtonHTMLAttributes<HTMLButtonElement> {
+  content: string | ReactNode;
+}
+
+export interface ButtonGroupDropdownItem extends Omit<DropdownProps, 'toggle'> {
+  toggle: ButtonGroupItem;
+}
+
+export interface ButtonGroupProps extends HTMLAttributes<HTMLDivElement>, MarginProps {
+  items: Array<ButtonGroupItem | ButtonGroupDropdownItem>;
+}
+
+const isDropdownItem = (item: ButtonGroupItem | ButtonGroupDropdownItem): item is ButtonGroupDropdownItem =>
+  'items' in item && Array.isArray(item.items);
+
+const getButtonProps = (item: ButtonGroupItem | ButtonGroupDropdownItem) => (isDropdownItem(item) ? item.toggle : item);
+
+const ButtonItem: React.FC<ButtonGroupItem & { isFirst: boolean; isLast: boolean }> = memo(
+  ({ content, isFirst, isLast, ...restProps }) => {
+    if (typeof content === 'string') {
+      return (
+        <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast}>
+          {content}
+        </StyledButton>
+      );
+    }
+
+    return <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast} iconOnly={content} />;
+  },
+);
+
+export const ButtonGroup: React.FC<ButtonGroupProps> = memo(({ items, ...marginProps }) => {
+  const renderedItems = useMemo(
+    () =>
+      items.map((item, index) => {
+        const { content, ...buttonProps } = getButtonProps(item);
+        const button = (
+          <ButtonItem
+            {...buttonProps}
+            role="listitem"
+            content={content}
+            isFirst={index === 0}
+            isLast={index + 1 === items.length}
+          />
+        );
+
+        if (isDropdownItem(item)) {
+          const { toggle, ...dropdownProps } = item;
+
+          return <Dropdown {...dropdownProps} key={index} toggle={button} />;
+        }
+
+        return <Fragment key={index}>{button}</Fragment>;
+      }),
+    [items],
+  );
+
+  return (
+    <Flex {...marginProps} flexDirection="row" flexWrap="nowrap" role="list">
+      {renderedItems}
+    </Flex>
+  );
+});
--- a/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
@@ -110,7 +110,13 @@ export const ButtonGroup: React.FC<ButtonGroupProps> = memo(({ actions, ...wrapp
   const renderedActions = useMemo(
     () =>
       actionsState.map((stateObj, index) => (
-        <StyledFlexItem key={index} ref={stateObj.ref} isVisible={stateObj.isVisible} role="listitem">
+        <StyledFlexItem
+          data-testid={`button-group-action-${index}`}
+          key={index}
+          ref={stateObj.ref}
+          isVisible={stateObj.isVisible}
+          role="listitem"
+        >
           <StyledButton
             {...stateObj.action}
             isFirst={index === 0}
--- a/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
@@ -1,71 +1,148 @@
-import React, { ButtonHTMLAttributes, Fragment, HTMLAttributes, memo, ReactNode, useMemo } from 'react';
+import { MoreHorizIcon } from '@bigcommerce/big-design-icons';
+import React, {
+  createRef,
+  HTMLAttributes,
+  isValidElement,
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 
+import { useWindowResizeListener } from '../../hooks';
 import { MarginProps } from '../../mixins';
-import { Dropdown, DropdownProps } from '../Dropdown';
+import { ButtonProps } from '../Button';
+import { Dropdown } from '../Dropdown';
 import { Flex } from '../Flex';
 
-import { StyledButton } from './styled';
+import { StyledButton, StyledFlexItem } from './styled';
 
-export interface ButtonGroupItem extends ButtonHTMLAttributes<HTMLButtonElement> {
-  content: string | ReactNode;
-}
-
-export interface ButtonGroupDropdownItem extends Omit<DropdownProps, 'toggle'> {
-  toggle: ButtonGroupItem;
+export interface ButtonGroupAction extends Omit<ButtonProps, 'children' | 'iconLeft' | 'iconRight'> {
+  text: string;
 }
 
 export interface ButtonGroupProps extends HTMLAttributes<HTMLDivElement>, MarginProps {
-  items: Array<ButtonGroupItem | ButtonGroupDropdownItem>;
+  actions: ButtonGroupAction[];
 }
 
-const isDropdownItem = (item: ButtonGroupItem | ButtonGroupDropdownItem): item is ButtonGroupDropdownItem =>
-  'items' in item && Array.isArray(item.items);
-
-const getButtonProps = (item: ButtonGroupItem | ButtonGroupDropdownItem) => (isDropdownItem(item) ? item.toggle : item);
-
-const ButtonItem: React.FC<ButtonGroupItem & { isFirst: boolean; isLast: boolean }> = memo(
-  ({ content, isFirst, isLast, ...restProps }) => {
-    if (typeof content === 'string') {
-      return (
-        <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast}>
-          {content}
-        </StyledButton>
-      );
-    }
-
-    return <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast} iconOnly={content} />;
-  },
-);
-
-export const ButtonGroup: React.FC<ButtonGroupProps> = memo(({ items, ...marginProps }) => {
-  const renderedItems = useMemo(
-    () =>
-      items.map((item, index) => {
-        const { content, ...buttonProps } = getButtonProps(item);
-        const button = (
-          <ButtonItem
-            {...buttonProps}
-            role="listitem"
-            content={content}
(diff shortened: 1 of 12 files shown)


packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,31 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled, { css } from 'styled-components';
+
+import { ButtonProps } from '../Button';
+import { StyleableButton } from '../Button/private';
+
+export const StyledButton = styled(StyleableButton)<ButtonProps & { isFirst?: boolean; isLast?: boolean }>`
+  border-radius: 0;
```
animesh1987, line 8: Can use `theme.borderRadius.none` here ?

golcinho: Thank you for your comment 🙇 

Sure I can 😄 done ✅

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ({ content, isFirst, isLast, ...restProps }) => {
+    if (typeof content === 'string') {
+      return (
+        <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast}>
+          {content}
+        </StyledButton>
+      );
+    }
+
+    return <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast} iconOnly={content} />;
```
animesh1987, line 36: Can we have an icon and text as part of content in a button?

golcinho: I talk to Eugen Polev about this requirement... so, we can't use `icon` + `text` at the same time

golcinho: Also, we have a style restrictions... only `secondary` variant.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,71 @@
...
+          return <Dropdown {...dropdownProps} key={index} toggle={button} />;
+        }
+
+        return <Fragment key={index}>{button}</Fragment>;
+      }),
+    [items],
+  );
+
+  return (
+    <Flex {...marginProps} flexDirection="row" flexWrap="nowrap" role="list">
```
animesh1987, line 67: Is the flexDirection needed here?

golcinho: yes, I avoides `column` direction on mobile devices.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
eugene-polev: I think we can allow icon and text option but in our guidelines we will recommend using text for all buttons in the group except for the ellipsis for more actions

golcinho: @eugene-polev so, should I add possibility to use `icon` + `text`? or should we restrict it?

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,71 @@
...
+export interface ButtonGroupItem extends ButtonHTMLAttributes<HTMLButtonElement> {
+  content: string | ReactNode;
+}
+
+export interface ButtonGroupDropdownItem extends Omit<DropdownProps, 'toggle'> {
+  toggle: ButtonGroupItem;
+}
+
+export interface ButtonGroupProps extends HTMLAttributes<HTMLDivElement>, MarginProps {
+  items: Array<ButtonGroupItem | ButtonGroupDropdownItem>;
```
chanceaclark, line 18: So I think we should make this API a little more consistent with some of our existing patterns, like the `Modal` `actions` prop: https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Modal/Modal.tsx#L22.

I'll talk a little be more about my vision of `ButtonGroupDropdownItem` in a new comment.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import { StyledButton } from './styled';
+
+export interface ButtonGroupItem extends ButtonHTMLAttributes<HTMLButtonElement> {
+  content: string | ReactNode;
+}
+
+export interface ButtonGroupDropdownItem extends Omit<DropdownProps, 'toggle'> {
+  toggle: ButtonGroupItem;
+}
```
chanceaclark, line 15: We should do this overflow dropdown for the consumer. Essentially get rid of this interface and just make the component consume an array of `actions`. This overflow button could reuse some of the logic we did for the [`PillTabs` component](https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/PillTabs/PillTabs.tsx#L34). I would consider abstracting some of that overflow logic into a utility hook or something and reuse it in both places.

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,31 @@
...
+      border-top-left-radius: ${theme.borderRadius.normal};
+    `}
+
+  ${({ isLast, theme }) =>
+    isLast &&
+    css`
+      border-bottom-right-radius: ${theme.borderRadius.normal};
+      border-top-right-radius: ${theme.borderRadius.normal};
+      margin-right: 0;
+    `}
```
chanceaclark, line 24: Could we possibly use `:first-of-type` and `:last-of-type` here?

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,31 @@
...
+    
+  &:focus {
+    z-index: 1;
+  }
```
chanceaclark, line 28: Curious as to why this is needed?

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,71 @@
...
+          {content}
+        </StyledButton>
+      );
+    }
+
+    return <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast} iconOnly={content} />;
+  },
+);
+
+export const ButtonGroup: React.FC<ButtonGroupProps> = memo(({ items, ...marginProps }) => {
```
chanceaclark, line 40: May also consider adding a `maxWidth` prop? Could be useful for a consumer 🤷

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
chanceaclark: We already have it built into our `Button` component that if the `iconOnly` prop is present, the text content will not show. From my previous comment above, I think we should use the same pattern with `actions` in our `Modal` component to consume an array of `ButtonProps & { text: string; }`. That way the consumer can add the `iconOnly` prop if needed.

In combination with this component handling the overflow, we should enforce `text` to always be present so it can populate the dropdown, even if the button has the `iconOnly` prop.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
jorgemoya: I 100% agree with this

golcinho:
@chanceaclark Thank you for your comments! I missed the `actions` pattern 🤦 Let me re-build my version...

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
golcinho: Got it, thank you, guys!

I have seen the `PillTabs` and I thought that the `ButtonGroup` concept should be a lil bit different but, yeah, I agree with you and re-build my version 🤝 

cc @jorgemoya @chanceaclark

packages/big-design/src/components/ButtonGroup/styled.tsx
golcinho: After re-build I 💯 can do that

packages/big-design/src/components/ButtonGroup/styled.tsx
golcinho: the explanation is pretty simple, I want to achieve correct `box-shadow` behavior, pls see the right border on the example below

**Without `z-index`**
Screenshot 2021-04-22 at 10 29 58

**With `z-index`**
Screenshot 2021-04-22 at 10 29 46

golcinho:
@chanceaclark @eugene-polev 

As I see `ButtonGroup` and `PillTabs` components are pretty similar... Now I see only one difference - and this difference is `styles` 🤔 

After discussion with @eugene-polev, I realized that we want to restrict the ability to use icons with `buttons` inside the `button group` component. Also, `PillTabs` and `ButtonGroup` have the same behavior around the `more` dropdown...

| `PillTabs` | `ButtonGroup` |
| --- | --- |
| Screenshot 2021-04-22 at 11 44 51 | Screenshot 2021-04-22 at 10 29 58 |

Will it be a good idea to combine `PillTabs` and `ButtonGroup` into a single component where we can choose the needed variant for example: `variant="pills"` or `variant="button-group"`? 

It just a proposal, need your thoughts guys 🙏

jorgemoya:
I would definitely prefer for them to be separate components, at least when exposed to the user, since they serve different use cases. But I agree it seems they are very similar in many ways, except Pills hide overflown pills and button group doesn't hide buttons?

chanceaclark:
@golcinho 

> Will it be a good idea to combine `PillTabs` and `ButtonGroup` into a single component where we can choose the needed variant for example: `variant="pills"` or `variant="button-group"`?

Let's try not to introduce a `variant` prop on this. `PillTabs` has a different API on how consumers use it. How I see this component is it either accepts `ButtonProps[]` or, _in the future_, `ToggleButtonProps[]` but no intersection of the two interfaces. `PillTabs` can consume a `ButtonGroup` component, but should not be strictly coupled.

For now, account for  `ButtonGroup` to just consume an array of `ButtonProps` and we can modify the component in the future to support `ToggleButtons`, just keep that in mind when reworking this.

Regarding styles, if we need to expose being able to style the `ButtonGroup` we can always create an internal usage only `private.tsx` file for consumption in the `PillsTabs`.

---

> After discussion with @eugene-polev, I realized that we want to restrict the ability to use icons with `buttons` inside the `button group` component. Also, `PillTabs` and `ButtonGroup` have the same behavior around the `more` dropdown...

Since `actions` extends `ButtonProps`, maybe we can omit `iconLeft` + `iconRight` and add a new "prop"

```ts
export interface ModalAction extends Omit<ButtonProps, 'children' | 'iconLeft' | 'iconRight'> {
  // Enforce text for iconOnly buttons to populate the dropdown
  text: string;
  // Icon for the dropdown item
  dropdownIcon?: React.ReactNode;
}
```

I'm not sold on the name `dropdownIcon` but what do you think about that?

golcinho:
@chanceaclark @jorgemoya Thank you guys for your thoughts 🙇 

Now, I don't do any combining and still follow your concept 🤝 

But, let me clarify my message above...

Requirements that I have: 
- Buttons inside button group should `secondary` style;
- Buttons should not have the possibility to use `icons` and `icon` can only be the `more` dropdown with overflowed buttons but, this is not requirement and it will be a recommendation on the `guidelines` page so, based on this, I can assume that this is not yet fully worked out in the guidelines.

According to all info above we can not say that the `ButtonGroup` actions prop will have `ButtonProps[]` type cause we can not use almost all `ButtonProps` properties:
- We can not use `variant` & `actionType` cause we support only `secondary` variant;
- We can not use `iconLeft`, `iconOnly` & `iconRight` according to guideline recommendations;
- We probably can not use normally `mobileWidth` cause `buttons` will be grouped;
So, probably `ButtonGroup` will have a more similar interface with `PillTab` items instead of `ButtonProps[]`  and I suppose that it can confuse the ended consumers of our library. 

Accroding to all info above, I proposed to merge `PillTab` and `ButtonGroup` but, should not be `PillTab` it could be component with more generic name.

How I see the `ButtonGroup` component: I think it should component which can group `Buttons` with all their props. we can get `bootstrap` for the referance.

Now we will get `PillTab` and `ButtonGroup` components with almost the same interfaces and functionalities 😄 

Sorry for holywar, just want to explain more my position.

golcinho:
> How I see the ButtonGroup component: I think it should be a component which can group Buttons with all their props. we can get bootstrap for the referance.

e.g.

```
<ButtonGroup>
    <Button variant="primary">Primary</Button>
    <Button variant="primary">Primary</Button>
    <Button variant="primary">Primary</Button>
</ButtonGroup>
```

chanceaclark:
@golcinho I disagree that this API should look like `PillTabs` API. We can still make the API generic and let `actions` be an array of "`ButtonProps`", and restrict what we need based off of today's requirements. In the future we can remove some of the restrictions as needed, for example allowing `subtle` variant buttons. We do something like this approach in the `InlineMessage` component, something you added too 😅 . The API should be more about what it looks like to the consumer, assuming a button group a **group of buttons**, rather than what it should look like to us developing it.

---

I also have a tough time acknowledging those are the complete requirements since,
1. There are two separate files that have completely two different designs and guidelines **and...**
2. In one of those files the only thing it says are that `Button Groups` are "**a group of buttons**".

cc. @eugene-polev 

(Reach out to me if you are wondering what files I'm talking about).

---

>How I see the ButtonGroup component: I think it should be a component which can group Buttons with all their props. we can get bootstrap for the referance.

We shouldn't use the same API as Bootstrap. BigDesign and Bootstrap solve completely different things and we should design for [configuration over flexibility](https://intranet.bigcommerce.com/display/CREAT/Tutorial+for+Building+a+BigDesign+React+Component#TutorialforBuildingaBigDesignReactComponent-ConfigurationoverFlexibility). However, I do agree with you that the component should just work with "an array of buttons", per my proposed API above.

---

However, this discussion should be taken offline. @golcinho set up a time to chat about this with me.

golcinho:
Close and reopen this RP for clearing comments log

PR #528, opened 2021-04-30
fix(component): add element check to popover closeOnClickOutside effect

### LINKED ISSUE #519: Popover example code not working
**Describe the bug**
Popover example code does not work on either the playground or a local environment.

**To Reproduce**
1. Setup global styles with the `GlobalStyles` component in the root component.
2. Use the Popover example code on the documentation/playground after importing the required components and `useState` hook. 

**Expected behavior**
I expected something to pop up like the [image here](https://github.com/bigcommerce/big-design/pull/441) but nothing really pops up.

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 89.0.4389.128

chanceaclark:
## What

Fixes #519

## Why

Event click delegation was running when the `Popover` component mounted, causing the `closeOnClickOutside` effect to run.

The idea is to let the anchor elements `onClick` handler take precedence over `closeOnClickOutside`.

## Testing

https://user-images.githubusercontent.com/10539418/116724117-e022b100-a9a5-11eb-9f4c-c037957ba3e0.mov

### DIFF
--- a/packages/big-design/src/components/Popover/Popover.tsx
+++ b/packages/big-design/src/components/Popover/Popover.tsx
@@ -126,6 +126,10 @@ const InternalPopover: React.FC<InternalPopoverProps> = ({
         return;
       }
 
+      if (anchorElement?.contains(event.target)) {
+        return;
+      }
+
       onClose();
     };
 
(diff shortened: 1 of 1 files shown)


jorgemoya (approved):
🙏

PR #532, opened 2021-05-10
feat(component): implements modal column type

jorgemoya:
## ❤️ ❤️ 

https://user-images.githubusercontent.com/196129/117721684-c94e3c80-b1a5-11eb-8620-fa32275a72db.mp4

## What
Implement Modal type. 

## How
Column of type `modal` has a `config` object that will receive a component to render, passing the value of the selected cell and a callback to set a new value.

```  
config: {
  header?: string;
  render(
    value: Item[keyof Item],
    onChange: (value: Item[keyof Item]) => void,
  ): React.ComponentType<Item> | React.ReactElement;
};
type: 'modal';
```

### DIFF
--- a/packages/big-design/src/components/Worksheet/Worksheet.tsx
+++ b/packages/big-design/src/components/Worksheet/Worksheet.tsx
@@ -4,10 +4,11 @@ import { typedMemo } from '../../utils';
 
 import { UpdateItemsProvider } from './context';
 import { useStore } from './hooks';
+import { WorksheetModal } from './Modal/Modal';
 import { Row } from './Row';
 import { Status } from './RowStatus/styled';
 import { Header, Table } from './styled';
-import { WorksheetItem, Worksheet as WorksheetProps } from './types';
+import { WorksheetItem, WorksheetModalColumn, Worksheet as WorksheetProps } from './types';
 import { editedRows, invalidRows } from './utils';
 
 const InternalWorksheet = <T extends WorksheetItem>({
@@ -62,12 +63,21 @@ const InternalWorksheet = <T extends WorksheetItem>({
     [columns, rows],
   );
 
+  const renderedModals = useMemo(
+    () =>
+      columns
+        .filter((column): column is WorksheetModalColumn<T> => column.type === 'modal')
+        .map((column, index) => <WorksheetModal column={column} key={index} />),
+    [columns],
+  );
+
   return (
     <UpdateItemsProvider items={rows}>
       <Table>
         {renderedHeaders}
         {renderedRows}
       </Table>
+      {renderedModals}
     </UpdateItemsProvider>
   );
 };
--- a/packages/big-design/src/components/Worksheet/editors/ModalEditor/styled.ts
+++ b/packages/big-design/src/components/Worksheet/editors/ModalEditor/styled.ts
@@ -1,11 +1,22 @@
 import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import styled from 'styled-components';
 
-import { Link } from '../../../Link';
+import { StyleableButton } from '../../../Button/Button';
 
-export const StyledLink = styled(Link)`
+export const StyledButton = styled(StyleableButton)`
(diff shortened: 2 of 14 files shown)


packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -611,3 +711,30 @@ describe('CheckboxEditor', () => {
...
+    const { getAllByText } = render(<Worksheet columns={columns} items={items} onChange={handleChange} />);
+
+    const buttons = getAllByText('Edit');
+
+    expect(buttons.length).toBe(9);
+
+    await waitForElement(() => screen.getAllByRole('combobox'));
+  });
+
+  // test('CheckboxEditor shows the appropriate state', async () => {
```
chanceaclark, line 726: Comment?

chanceaclark (commented):
Got a few comments for you, but looks really good!

packages/big-design/src/components/Worksheet/types.ts
```diff
@@ -26,12 +31,31 @@ export interface WorksheetSelectableColumn<Item> extends Omit<WorksheetBaseColum
   type: 'select';
 }
 
+export interface WorksheetModalColumn<Item> extends Omit<WorksheetBaseColumn<Item>, 'type'> {
+  config: {
```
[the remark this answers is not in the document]
jorgemoya: Good idea

packages/big-design/src/components/Worksheet/Worksheet.tsx
```diff
@@ -62,12 +63,21 @@ const InternalWorksheet = <T extends WorksheetItem>({
     [columns, rows],
   );
 
+  const renderedModals = useMemo(
+    () =>
+      columns
+        .filter((column): column is WorksheetModalColumn<T> => column.type === 'modal')
+        .map((column, index) => <WorksheetModal column={column} key={index} />),
```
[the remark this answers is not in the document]
jorgemoya: Can you give me an idea? I was under the impression only if needed (aka if there was a column type of `modal`) the Modal would be included with this code.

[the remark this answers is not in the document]
chanceaclark: Talked offline, going to tackle this in another PR

packages/big-design/src/components/Worksheet/editors/ModalEditor/styled.ts
```diff
@@ -0,0 +1,11 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+import { Link } from '../../../Link';
+
+export const StyledLink = styled(Link)`
```
chanceaclark, line 6: Sorry, just caught this. But we probably don't want a link here since it breaks focusability if we don't have an `href`. We probably should use `StyleableButton` instead and make sure keyboard events work properly for this cell.

chanceaclark (commented):
One last thing I caught 🙇

packages/big-design/src/components/Worksheet/editors/ModalEditor/styled.ts
jorgemoya: Ok, good point, let me look into it

packages/big-design/src/components/Worksheet/editors/ModalEditor/styled.ts
```diff
@@ -0,0 +1,22 @@
...
+  height: auto;
+  line-height: ${({ theme }) => theme.lineHeight.small};
+  padding: 0;
+
+  & :hover:not(:active) {
+    background-color: inherit;
+    color: ${({ theme }) => theme.colors.primary70};
+  }
+
+  & :active {
```
chanceaclark, line 17: Do these need to not have spaces?
```suggestion
  &:hover:not(:active) {
    background-color: inherit;
    color: ${({ theme }) => theme.colors.primary70};
  }

  &:active {
```

chanceaclark (approved):
Looks good, thanks for splitting up the tests too ✅

PR #538, opened 2021-05-14
feat(component): allow formatting of columns

jorgemoya:
## What
Allow the Text, Number, and Modal column types to be formatted.

## Why
Allow custom text readability.

## API
```
{
  hash: 'numberField',
  header: 'Number field',
  type: 'number',
  validation: (value: number) => value >= 50,
  formatting: (value: number) => `$${value}.00`,
},
```

## Screenshot

Screen Shot 2021-05-14 at 2 47 00 PM

### DIFF
--- a/packages/big-design/src/components/Worksheet/Row/Row.tsx
+++ b/packages/big-design/src/components/Worksheet/Row/Row.tsx
@@ -1,10 +1,16 @@
-import React, { useMemo } from 'react';
+import React, { useCallback, useMemo } from 'react';
 
 import { typedMemo } from '../../../utils';
 import { Cell } from '../Cell';
 import { useStore } from '../hooks';
 import { RowStatus } from '../RowStatus';
-import { WorksheetColumn, WorksheetItem } from '../types';
+import {
+  WorksheetColumn,
+  WorksheetItem,
+  WorksheetModalColumn,
+  WorksheetNumberColumn,
+  WorksheetTextColumn,
+} from '../types';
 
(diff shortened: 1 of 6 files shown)


packages/big-design/src/components/Worksheet/Row/Row.tsx
```diff
@@ -14,12 +20,20 @@ interface RowProps<Item> {
...
+    return column.type === 'text' || column.type === 'number' || column.type === 'modal';
+  }, []);
+
   return (
     <tr>
       <RowStatus rowIndex={rowIndex} />
       {columns.map((column, columnIndex) => (
         <Cell
           columnIndex={columnIndex}
+          formatting={hasFormatting(column) ? column.formatting : undefined}
```
chanceaclark, line 36: Couldn't we just use optional chaining + nullish coalescing here? Or am I missing something?
```suggestion
          formatting={column?.formatting ?? undefined}
```

chanceaclark (approved):
Left a comment, but otherwise looks good to me 👍

packages/big-design/src/components/Worksheet/Row/Row.tsx
jorgemoya: We can't because `formatting` isn't present in all `column` types.

chanceaclark: Ah I guess it's TS in this case complaining on the checkbox column types.

PR #541, opened 2021-05-17
chore: add @bigcommerce/product-design to codeowners

chanceaclark:
## What

Adds the @bigcommerce/product-design team as codeowners.

## Why

The @bigcommerce/product-design team wants to have more insight on additions/changes from external contributions. Adding the team on all PR's will give more visibility on all development.

**Note: We are still requiring a member of @bigcommerce/frontend team to approve before merging code.**

cc. @valterfatia @eugene-polev

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @bigcommerce/frontend
+* @bigcommerce/frontend @bigcommerce/product-design


deini (approved):

jorgemoya (approved):

PascalZajac:
I don't believe codeowners works unless the team has Write access to the repository (`product-design` currently has Triage). If we grant that, I believe it will force that team to provide approvals, regardless of whether a "co-owner" like Frontend has approved.

It would probably be better if the members of `product-design` simply watched the repository directly.

chanceaclark:
We currently only require 1 approval so if we have two (or multiple) codeowners with Write access, only one member from any team needs to approve to allow the PR to move forward. We currently do this in some of our other internal multi-team repos so it'll not force that team to provide approvals.

PascalZajac:
That sounds fine then. We'll still have to adjust permissions for this to work.

chanceaclark:
I'll do that once I merge this in then 👍

PR #542, opened 2021-05-17
feat(component): add autoComplete prop for selects

### LINKED ISSUE #540: Select autocomplete not disabled in Chrome
**Describe the bug**

When using the [Select](https://developer.bigcommerce.com/big-design/select) component, I would like to disable the autocomplete functionality. More specifically, I would like to stop Chrome from suggesting addresses on a country select because the Browser dropdown overlays the component's list making it impossible to select a dropdown item until the autofill suggestion window has been dismissed which is not trivial.

Additionally, as any autofilled country does not map to a valid option, the autofill functionality fails, adding to the confusion. 

**To Reproduce**

1. In Chrome, Implement a country dropdown as part of an address form
2. Click on country dropdown.
3. See "Manage Addresses" dropdown instead of the expected Select option dropdown.

Here's an example: https://codesandbox.io/s/compassionate-hypatia-6e896?file=/src/index.tsx

**Expected behavior**

This is a bit of a tricky problem, so I'm not entirely sure of the best way forward.

My first step was to try and find a way to force disable the browser suggestions. I believe this would require the ability to override the `autocomplete` attribute. Currently, it's [hardcoded to `off`](https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Select/Select.tsx#L238) with no ability to override by my understanding. This sounds logical but unfortunately `off` doesn't disable this functionality in chrome, but by my tests, setting it to `autocomplete="chrome-off"` will stop the browser address suggestions from showing. This doesn't sound like a great solution from a cross-browser point of view though. 

Any suggestions on ways to solve this on either library or the implementation level?

**Screenshots**

Screenshot 2021-05-17 at 16 52 16
 
Screenshot 2021-05-17 at 16 52 20

**Desktop (please complete the following information, if applicable):**
 - OS: Mac Big Sur 11.2.3
 - Browser: Chrome 90

jorgemoya:
## What
Allow custom values for autoComplete. Defaults to `off`.

## Why
Fixes #540

### DIFF
--- a/packages/big-design/src/components/MultiSelect/types.ts
+++ b/packages/big-design/src/components/MultiSelect/types.ts
@@ -7,6 +7,7 @@ import { SelectOptionGroup } from '../Select/types';
 
 interface BaseSelect extends Omit<React.HTMLAttributes<HTMLInputElement>, 'children'> {
   action?: SelectAction;
+  autoComplete?: string;
   autoWidth?: boolean;
   description?: React.ReactChild;
   disabled?: boolean;
--- a/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
@@ -26,6 +26,7 @@ import { MultiSelectProps } from './types';
 export const MultiSelect = typedMemo(
   <T extends unknown>({
     action,
+    autoComplete = 'off',
     autoWidth = false,
     className,
     disabled,
@@ -314,7 +315,7 @@ export const MultiSelect = typedMemo(
           <Input
             {...getInputProps({
               ...props,
-              autoComplete: 'off',
+              autoComplete,
               disabled,
               onClick: () => {
                 !isOpen && openMenu();
@@ -359,6 +360,7 @@ export const MultiSelect = typedMemo(
         </StyledInputContainer>
       );
     }, [
+      autoComplete,
       disabled,
       filterable,
       getInputProps,
--- a/packages/big-design/src/components/Select/Select.tsx
+++ b/packages/big-design/src/components/Select/Select.tsx
@@ -25,6 +25,7 @@ import { SelectAction } from '../Select/types';
 export const Select = typedMemo(
   <T extends unknown>({
     action,
+    autoComplete = 'off',
     autoWidth = false,
     className,
     disabled,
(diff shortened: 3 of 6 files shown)


packages/big-design/src/components/MultiSelect/types.ts
```diff
@@ -7,6 +7,7 @@ import { SelectOptionGroup } from '../Select/types';
 
 interface BaseSelect extends Omit<React.HTMLAttributes<HTMLInputElement>, 'children'> {
   action?: SelectAction;
+  autoComplete?: string;
```
chanceaclark, line 10: Can you check that this is not on `React.HTMLAttributes<HTMLInputElement>`?

chanceaclark (commented):
Small comment, but could you actually show how it fixes #540?

packages/big-design/src/components/MultiSelect/types.ts
jorgemoya: It isn't

chanceaclark: Ah makes sense. We'd probably want to do this for the `Form` component too then.

jorgemoya: I assume it defaults to `on` on the form?

chanceaclark: Yep

jorgemoya: `FormHTMLAttributes` does have `autoComplete`

jorgemoya: Hmm, I could swap the type of the Selects for `<React.InputHTMLAttributes<HTMLInputElement>` which do include `autoComplete` and a bunch of others

chanceaclark: 🤔 

Only thing we need to make sure if we do, is that the types we say we allow (for example, valid input attributes like any `aria-` ones, or the `multiple` attribute) are actually passed in to the underlying component. We don't want our types to say we have something that our component doesn't actually implement.

[Full list of input attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attributes)

jorgemoya: Lets keep as is since I don't want to expose all those to the user.

chanceaclark (approved):

rtalvarez (approved):

jordan-massingill (approved):

PR #543, opened 2021-05-20

golcinho:
## What?

*Note:* This is a reopen version of https://github.com/bigcommerce/big-design/pull/523 (reopen for clear huge offtopic comments).

## Screenshots & Video
TBD

## Proof
Unit test + local testing.

ping @chanceaclark

### DIFF
--- a/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
@@ -1,0 +1,71 @@
+import React, { ButtonHTMLAttributes, Fragment, HTMLAttributes, memo, ReactNode, useMemo } from 'react';
+
+import { MarginProps } from '../../mixins';
+import { Dropdown, DropdownProps } from '../Dropdown';
+import { Flex } from '../Flex';
+
+import { StyledButton } from './styled';
+
+export interface ButtonGroupItem extends ButtonHTMLAttributes<HTMLButtonElement> {
+  content: string | ReactNode;
+}
+
+export interface ButtonGroupDropdownItem extends Omit<DropdownProps, 'toggle'> {
+  toggle: ButtonGroupItem;
+}
+
+export interface ButtonGroupProps extends HTMLAttributes<HTMLDivElement>, MarginProps {
+  items: Array<ButtonGroupItem | ButtonGroupDropdownItem>;
+}
+
+const isDropdownItem = (item: ButtonGroupItem | ButtonGroupDropdownItem): item is ButtonGroupDropdownItem =>
+  'items' in item && Array.isArray(item.items);
+
+const getButtonProps = (item: ButtonGroupItem | ButtonGroupDropdownItem) => (isDropdownItem(item) ? item.toggle : item);
+
+const ButtonItem: React.FC<ButtonGroupItem & { isFirst: boolean; isLast: boolean }> = memo(
+  ({ content, isFirst, isLast, ...restProps }) => {
+    if (typeof content === 'string') {
+      return (
+        <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast}>
+          {content}
+        </StyledButton>
+      );
+    }
+
+    return <StyledButton {...restProps} variant="secondary" isFirst={isFirst} isLast={isLast} iconOnly={content} />;
+  },
+);
+
+export const ButtonGroup: React.FC<ButtonGroupProps> = memo(({ items, ...marginProps }) => {
+  const renderedItems = useMemo(
+    () =>
+      items.map((item, index) => {
+        const { content, ...buttonProps } = getButtonProps(item);
+        const button = (
+          <ButtonItem
+            {...buttonProps}
+            role="listitem"
+            content={content}
+            isFirst={index === 0}
+            isLast={index + 1 === items.length}
+          />
+        );
+
+        if (isDropdownItem(item)) {
+          const { toggle, ...dropdownProps } = item;
+
+          return <Dropdown {...dropdownProps} key={index} toggle={button} />;
+        }
+
+        return <Fragment key={index}>{button}</Fragment>;
+      }),
+    [items],
+  );
+
+  return (
+    <Flex {...marginProps} flexDirection="row" flexWrap="nowrap" role="list">
+      {renderedItems}
+    </Flex>
+  );
+});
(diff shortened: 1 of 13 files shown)


andriiVitvitskyi1990 (approved):

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,161 @@
...
+            }
+          },
+          hash: stateObj.action.text.toLowerCase(),
+          icon,
+        };
+      });
+
+    return (
+      <StyledFlexItem
+        data-testid="button-group-dropdown-toggle"
```
jorgemoya, line 97: As far as I know, we don't set `data-testid` in component directly, only when passed in as a prop during testing.

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,52 @@
...
+    css`
+      position: absolute;
+      visibility: hidden;
+      z-index: ${({ theme }) => -theme.zIndex.tooltip};
+    `}
+`;
+
+export const StyledButton = styled(StyleableButton)<StyledButtonProps>`
+  border-radius: ${({ theme }) => theme.borderRadius.none};
+  margin-right: -1px;
```
jorgemoya, line 29: Just curious, why do we need this `-1px`?

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,52 @@
...
+    isLast &&
+    css`
+      border-bottom-right-radius: ${theme.borderRadius.normal};
+      border-top-right-radius: ${theme.borderRadius.normal};
+      margin-right: 0;
+    `}
+    
+  &:focus {
+    z-index: 1;
+  }
```
jorgemoya, line 48: Why is this needed when focused? Do you think we can use one of the `theme` zIndex values? 🍹

packages/docs/pages/ButtonGroup/ButtonGroupPage.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { Code, CodePreview } from '../../components';
+import { ButtonGroupPropTable } from '../../PropTables/ButtonGroupPropTable';
+
+const ButtonGroupPage = () => (
+  <>
+    <H0>Button Group</H0>
+
+    <Text>
+      The <Code primary>Button Group</Code> component is used for grouping actions like <Code primary>Button</Code>
```
jorgemoya, line 12: This looks weird to me.

```suggestion
      The <Code primary>Button Group</Code> component is used for grouping actions as <Code primary>Button</Code>
```

Maybe?

packages/docs/pages/ButtonGroup/ButtonGroupPage.tsx
```diff
@@ -0,0 +1,36 @@
...
+    </Text>
+
+    <CodePreview>
+      {/* jsx-to-string:start */}
+      <ButtonGroup actions={[{ text: 'Button' }, { text: 'Button' }, { text: 'Button' }]} />
```
jorgemoya, line 17: I wonder if it is needed to add an example with `onClicks` just for demonstration purposes. Thoughts?

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const StyledFlexItem = styled(FlexItem)<StyledFlexItemProps>`
+  ${(props) =>
+    !props.isVisible &&
+    css`
+      position: absolute;
+      visibility: hidden;
+      z-index: ${({ theme }) => -theme.zIndex.tooltip};
+    `}
+`;
```
jorgemoya, line 25: We tend to destruct the props.

```suggestion
  ${({ isVisible }) =>
    !isVisible &&
    css`
      position: absolute;
      visibility: hidden;
      z-index: ${({ theme }) => -theme.zIndex.tooltip};
    `}
`;
```

jorgemoya (commented):
Looking good. Just waiting on a nice, gif preview of the buttons in action.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
golcinho: I got PillTabs for the reference

jorgemoya:
@golcinho Oops, I thought you had made the changes.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
rtalvarez: Chance and I had a very elaborate discussion around this, and in this case it is very tricky to write a good test without data-testids.. in this case its a hacky nightmare, basically because of the `offsetWidths`, and ended up keeping the data-testids... these are the only reliable way we could think of for targeting them and setting the offsetWidths so that we could test the collapsing functionality (jsdom does not actually render things, so all elements have an offsetWidth of 0 in the testing environment)

valterfatia:
Visually it looks great 👌 – assuming it's the same as in the closed PR…

Reading through the code, I assume the intention is to hide/show the actions based on the space available but that opens a few questions for me, which I hope someone else in @bigcommerce/product-design can answer:

1. How do we ensure a consistent display of the icons (only icon, only text, text and icon)?
2. Will an action like delete be styled in red in both the dropdown and the list of buttons?
3. Would we allow/force certain actions to only be displayed in the dropdown?

eugene-polev:
@valterfatia you can check the guidelines here https://www.figma.com/file/duwtpbP2tnLLQQmAkwtrNC/BigDesign-Lib?node-id=5837%3A16851

valterfatia:
> @valterfatia you can check the guidelines here https://www.figma.com/file/duwtpbP2tnLLQQmAkwtrNC/BigDesign-Lib?node-id=5837%3A16851

@eugene-polev thanks! I think the guidelines don't answer my comments then. Looking at the guidelines we didn't consider that actions could be shown/hidden based on the available space and there are 2 assumptions that aren't reflected in the code, namely the ability to set an action to only show in the dropdown (destructive actions) and ensuring a consistent look and feel for all actions that are visible.

golcinho:
will re-open for clearing comments

PR #544, opened 2021-05-28
feat(component): implement keyboard navigation for Worksheet

jorgemoya:
## What

Add keyboard navigation to Worksheet component

1. Arrow keys navigation
2. Tab & Shift + Tab
3. Enter to modify cells

## Why

Primary effort to have working keyboard commands on the component, with the ultimate goal of adding copy & paste, plus cell multiselect.

## Screenshots

https://user-images.githubusercontent.com/196129/120032328-a4d4cb80-bfbf-11eb-85ca-38dc969194a8.mp4

## Tests
Some unit tests and manually

### DIFF
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -93,10 +93,23 @@ const InternalCell = <T extends WorksheetItem>({
     setSelectedCells([cell]);
   }, [cell, rowIndex, setSelectedCells, setSelectedRows]);
 
-  const renderedCell = useMemo(() => {
-    const cellValue =
-      value || value === 0 ? (formatting ? formatting(value) : `${value}`) : value !== 'undefined' ? `${value}` : '';
+  const renderedValue = useMemo(() => {
+    if (value !== 'undefined' && value !== '' && !Number.isNaN(value)) {
+      if (typeof formatting === 'function') {
+        return formatting(value);
+      }
 
+      return `${value}`;
+    }
+
+    if (Number.isNaN(value)) {
+      return `${value}`;
+    }
+
+    return '';
+  }, [formatting, value]);
+
+  const renderedCell = useMemo(() => {
     switch (type) {
       case 'select':
         return (
@@ -117,12 +130,12 @@ const InternalCell = <T extends WorksheetItem>({
         return isEditing ? (
           <TextEditor cell={cell} isEdited={isEdited} onBlur={handleBlur} onKeyDown={handleKeyDown} />
         ) : (
-          <Small color="secondary70" ellipsis title={cellValue}>
-            {`${cellValue}`}
+          <Small color="secondary70" ellipsis title={renderedValue}>
+            {renderedValue}
           </Small>
         );
     }
-  }, [cell, formatting, handleBlur, handleChange, handleKeyDown, isEdited, isEditing, options, type, value]);
+  }, [cell, formatting, handleBlur, handleChange, handleKeyDown, isEdited, isEditing, options, renderedValue, type]);
 
   return (
     <StyledCell
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -94,7 +94,8 @@ const InternalCell = <T extends WorksheetItem>({
   }, [cell, rowIndex, setSelectedCells, setSelectedRows]);
 
   const renderedCell = useMemo(() => {
-    const cellValue = value ? (formatting ? formatting(value) : `${value}`) : value !== 'undefined' ? `${value}` : '';
+    const cellValue =
+      value || value === 0 ? (formatting ? formatting(value) : `${value}`) : value !== 'undefined' ? `${value}` : '';
 
     switch (type) {
       case 'select':
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -96,11 +96,20 @@ const InternalCell = <T extends WorksheetItem>({
   const renderedCell = useMemo(() => {
     switch (type) {
       case 'select':
-        return <SelectEditor cell={cell} isEdited={isEdited} onChange={handleChange} options={options} />;
+        return (
+          <SelectEditor
+            cell={cell}
+            isEdited={isEdited}
+            isEditing={isEditing}
+            onBlur={handleBlur}
+            onChange={handleChange}
+            options={options}
+          />
+        );
       case 'checkbox':
-        return <CheckboxEditor cell={cell} onChange={handleChange} />;
+        return <CheckboxEditor cell={cell} isEditing={isEditing} onBlur={handleBlur} onChange={handleChange} />;
       case 'modal':
-        return <ModalEditor cell={cell} formatting={formatting} />;
+        return <ModalEditor cell={cell} formatting={formatting} isEditing={isEditing} />;
       default:
         return isEditing ? (
           <TextEditor cell={cell} isEdited={isEdited} onBlur={handleBlur} onKeyDown={handleKeyDown} />
@@ -108,7 +117,8 @@ const InternalCell = <T extends WorksheetItem>({
         value ? (
           <Small color="secondary70">{formatting ? formatting(value) : `${value}`}</Small>
         ) : (
-          <Small color="secondary70">{''}</Small>
+          // In case of NaN casting to string
+          <Small color="secondary70">{value !== 'undefined' ? `${value}` : ''}</Small>
         );
     }
   }, [cell, formatting, handleBlur, handleChange, handleKeyDown, isEdited, isEditing, options, type, value]);
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -94,6 +94,8 @@ const InternalCell = <T extends WorksheetItem>({
   }, [cell, rowIndex, setSelectedCells, setSelectedRows]);
 
   const renderedCell = useMemo(() => {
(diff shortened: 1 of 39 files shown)


jorgemoya:
The way this is implemented, we keep focus on the table and listen to keyboard events to handle the selected cell (contrary to focusing on each cell). This is because we eventually plan to allow multi cell selection so multiple focused elements won't come into play. Pointing this out because I don't know how this approach falls in line with accessibility.

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
```diff
@@ -94,21 +94,32 @@ const InternalCell = <T extends WorksheetItem>({
   }, [cell, rowIndex, setSelectedCells, setSelectedRows]);
 
   const renderedCell = useMemo(() => {
+    const cellValue =
```
rtalvarez, line 97: 🍹 I find this line quite long, not exactly easy to read through IMO

packages/big-design/src/components/Worksheet/Modal/Modal.tsx
```diff
@@ -48,7 +57,11 @@ const InternalWorksheetModal = <T extends WorksheetItem>({ column }: WorksheetMo
   return (
     <Modal
       actions={[
-        { text: 'Cancel', variant: 'subtle', onClick: () => setOpenModal(null) },
+        {
+          text: 'Cancel',
```
rtalvarez, line 61: Hmmm I wonder if we should provide a way for these action texts to be overriden, this text won't be able to be localized as it is

packages/big-design/src/components/Worksheet/Worksheet.tsx
```diff
@@ -17,21 +17,35 @@ const InternalWorksheet = <T extends WorksheetItem>({
...
   // Create a new reference since state mutates rows to prevent unecessary rerendering
   useEffect(() => setRows([...items]), [items, setRows]);
 
+  // Save columns to use in navigation
+  useEffect(() => setColumns(columns), [columns, setColumns]);
+
+  // Save ref to focus table when necessary
+  useEffect(() => setTableRef(tableRef.current), [setTableRef, tableRef]);
+
+  // Call onChange when a cell is edited
```
rtalvarez, line 41: I'm not sure if this comment (and below) are adding any meaningful information for the developers working in this file

packages/big-design/src/components/Worksheet/hooks/useEditableCell/spec.tsx
```diff
@@ -0,0 +1,166 @@
...
+  test('escape', () => {
+    const { result } = renderHook(() => useEditableCell(cell));
+
+    act(() => {
+      result.current.handleDoubleClick();
+      result.current.handleKeyDown(
+        ({
+          key: 'Escape',
+          preventDefault,
+        } as unknown) as React.KeyboardEvent<HTMLInputElement>,
```
rtalvarez, line 148: 🍹 hmm, i wonder if we should invest in getting rid of all of these `as` type casts, but not sure how strict yall are here in big-design

packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -463,6 +463,50 @@ describe('formatting', () => {
...
 
+describe('keyboard navigation', () => {
+  test('navigates with arrow keys', async () => {
+    const { getByText } = render(<Worksheet columns={columns} items={items} onChange={handleChange} />);
+
+    let cell = getByText('Shoes Name Three');
+
+    fireEvent.click(cell);
+
+    expect(cell.parentElement).toHaveStyle('border-color: #3C64F4');
```
rtalvarez, line 474: would it make sense to extract these values from the theme instead?

packages/big-design/src/components/Worksheet/editors/CheckboxEditor/CheckboxEditor.tsx
```diff
@@ -8,20 +8,35 @@ import { CheckboxWrapper } from './styled';
...
+  onChange,
+}: CheckboxEditorProps<T>) => {
+  useEffect(() => {
+    if (isEditing) {
+      // For checkbox we want to change the value when editing is called
+      onChange(!cell.value);
+    }
+  }, [cell.value, isEditing, onChange]);
+
   const handleChange = () => {
```
jordan-massingill, line 29: Just curious: for the checkbox, what's the difference in `isEditing` and `onChange` - are they effectively the same? Editing and "changing" the value?

packages/big-design/src/components/Worksheet/Modal/Modal.tsx
jorgemoya: Should I provide a prop in the `config` object to ask for custom text and default to english?

packages/big-design/src/components/Worksheet/hooks/useEditableCell/spec.tsx
jorgemoya: I wouldn't know how to use this without ts breaking. Maybe override the rule? Different fix, same problem?

packages/big-design/src/components/Worksheet/editors/CheckboxEditor/CheckboxEditor.tsx
jorgemoya: They are the same, except `onChange` is called when the checkbox is mouse clicked and the `isEditing` state changes when a user presses enter or space when the cell is selected. I will add some comments and change things to make it more clear.

packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
```diff
@@ -1,32 +1,46 @@
...
-  };
+  const buttonRef = createRef<HTMLButtonElement>();
+
+  useEffect(() => {
+    if (isEditing) {
+      setOpenModal(hash);
+    }
+  }, [hash, isEditing, setOpenModal]);
+
+  const handleClick = useCallback(() => {
```
jordan-massingill, line 30: 🍹 We could give this a more descriptive name (`handleEditCell` or something), but it's up to you 👍

packages/big-design/src/components/Worksheet/hooks/useKeyEvents/useKeyEvents.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        event.preventDefault();
+      }
+    },
+    [editSelectedCell, isEditing, navigate, selectedCell],
+  );
+
+  return useMemo(() => ({ handleKeyDown }), [handleKeyDown]);
+};
```
jordan-massingill, line 72: This file looks really good 👍

packages/big-design/src/components/Worksheet/Modal/Modal.tsx
rtalvarez: yeah that sounds good to me

packages/big-design/src/components/Worksheet/hooks/useKeyEvents/spec.tsx
```diff
@@ -0,0 +1,218 @@
...
+    event = new KeyboardEvent('keydown', { key: 'ArrowLeft' });
+    result.current.handleKeyDown(event);
+
+    expect(mockNavigate).toHaveBeenCalledWith({ rowIndex: 0, columnIndex: -1 });
+  });
+
+  test('navigates with tab', () => {
+    const { result } = renderHook(() => useKeyEvents());
+
+    let event = new KeyboardEvent('keydown', { key: 'Tab' });
```
chanceaclark, line 55: 🍹 We probably should replace these with [`@testing-library/user-events` tab fn](https://testing-library.com/docs/ecosystem-user-event/#tabshift-focustrap)

jordan-massingill (approved):
Looks good to me 👍 I think you're good to go after you address any outstanding comments. Nice work!

packages/big-design/src/components/Worksheet/hooks/useKeyEvents/spec.tsx
jorgemoya: `tab` seems to work best when we are using native tab navigation, but not so when we are listening for key events and handling "tabbing" in our own way.

packages/big-design/src/components/Worksheet/Cell/styled.tsx
```diff
@@ -44,6 +44,10 @@ export const StyledCell = styled.td<StyledCellProps<WorksheetItem>>`
     css`
       background-color: ${({ theme }) => theme.colors.warning10};
     `}
+
+  & p {
+    display: block;
+  }
```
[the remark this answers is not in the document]
jorgemoya: I'm 100% sure changing line-height doesn't work as desired. Am I 100% sure something else is making my line-height not to work? No. I tried a bunch of things and finally `display` made it work.

packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
jorgemoya: Still prefer `handleClick` since it handles the click of the element.

packages/big-design/src/components/Worksheet/hooks/useKeyEvents/spec.tsx
chanceaclark: I believe the user-events tab fn should work even when using our own key events. But this is definitely a 🍹 and most likely for another PR.

packages/big-design/src/components/Worksheet/Cell/styled.tsx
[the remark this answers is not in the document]
chanceaclark: Ah yeah, after looking, it seems like it's a little bit of a problem. Definitely not for this PR, but a few solutions popped into my mind:
1. Add the `display` prop for typography components. That way we can do `<Small display="block" />`.
2. Allow `as="div"` for a `<Small />` or `<Text />`. I don't like this approach since it breaks a11y in many cases.

I'm good with this though 👍

chanceaclark (approved):
Looks good to me! Great job man!

PR #549, opened 2021-06-10
feat(component): add onOpen and onClose to Selects

jorgemoya:
## What?

Add `onOpen` and `onClose` props to Selects.

## Why?

For some interactions on the Worksheet component (and other uses), I need to know when the dropdown has been opened or closed.

## Testing/Proof

Unit tests and locally

### DIFF
--- a/packages/big-design/src/components/MultiSelect/spec.tsx
+++ b/packages/big-design/src/components/MultiSelect/spec.tsx
@@ -10,6 +10,8 @@ import { MultiSelect } from './';
 
 const onChange = jest.fn();
 const onActionClick = jest.fn();
+const onOpen = jest.fn();
+const onClose = jest.fn();
 
 const mockOptions = [
   { value: 'us', content: 'United States' },
@@ -54,6 +56,8 @@ const MultiSelectMock = (
     data-testid="multi-select"
     error="Required"
     label="Countries"
+    onClose={onClose}
+    onOpen={onOpen}
     onOptionsChange={onChange}
     options={mockOptions}
     placeholder="Choose country"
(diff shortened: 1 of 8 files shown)


packages/big-design/src/components/MultiSelect/spec.tsx
```diff
@@ -369,6 +373,29 @@ test('clicking on disabled select options should not trigger onItemClick', async
...
+test('closing the MultiSelect triggers onClose', async () => {
+  const { getAllByRole, queryByRole } = render(MultiSelectMock);
+  const button = getAllByRole('button')[2];
+
+  fireEvent.click(button);
+  fireEvent.click(button);
+
+  expect(onClose).toHaveBeenCalled();
+
+  await waitForElement(() => queryByRole('listbox'));
```
icatalina, line 396: `queryBy*` should only be used for querying when the element is not present: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence


in this case we should use `findByRole` which does the wait for you:

```tsx
const listbox = await findByRole('listbox');

expect(listbox).toBeInTheDocument();
```

packages/big-design/src/components/MultiSelect/spec.tsx
```diff
@@ -369,6 +373,29 @@ test('clicking on disabled select options should not trigger onItemClick', async
...
+
+  fireEvent.click(button);
+
+  expect(onOpen).toHaveBeenCalled();
+
+  await waitForElement(() => queryByRole('listbox'));
+});
+
+test('closing the MultiSelect triggers onClose', async () => {
+  const { getAllByRole, queryByRole } = render(MultiSelectMock);
```
icatalina, line 388: It is recommended to use `screen` instead of using the output of render.

```ts
render(MultiSelectMock);

const button = screen.getAllByRole('button')[2];
```

packages/big-design/src/components/MultiSelect/spec.tsx
```diff
@@ -369,6 +373,29 @@ test('clicking on disabled select options should not trigger onItemClick', async
...
+  const button = getAllByRole('button')[2];
```
icatalina, line 389: It is better to target the actual button as opposed to use `[2]`. Not sure if it is feasible in this case, but it is better to use `getByRole('button', { name: /text of the button/i })`

PR #552, opened 2021-06-22
chore(component): add tooltip for table column header

bc-apostoliuk:
## What?

Adds 'tooltip' prop in the Table component, namely for the column header.

## Why?

To be able to use the Tooltip component for explaining the user some unobvious information about the column by hovering on the icon that will be displayed next to the column name.

image
image

## Testing/Proof

Unit test. Locally.


eugene-polev:
I think we should use 16px icon instead of 24px or 20px in order to reduce the visual weight of the icon comparing to the label
Screen Shot 2021-06-22 at 7 20 44 PM

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -73,6 +85,7 @@ const InternalHeaderCell = <T extends TableItem>({
       <StyledFlex alignItems="center" flexDirection="row" hide={hide} align={align}>
         {children}
         {!hide && renderSortIcon()}
+        {tooltip && renderTooltip()}
```
jordan-massingill, line 88: I think it be valuable to check not only that the tooltip is present but that it's also of type `string` before rendering it as-is (perhaps in `renderTooltip`, you can render `null` if not a string)

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -53,6 +54,17 @@ const InternalHeaderCell = <T extends TableItem>({
...
   };
 
+  const renderTooltip = () => {
+    return (
+      <>
+        &nbsp;
+        <Tooltip trigger={<BaselineHelpIcon data-testid="help-icon" />} placement="right-end">
+          {tooltip}
+        </Tooltip>
+      </>
```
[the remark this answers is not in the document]
chanceaclark: Looked at the designs and seem like we need to modify this a bit. Notice the `Box` and the new `placement` value.
```suggestion
      <Tooltip
        trigger={
          <Box as="span" marginLeft="xxSmall">
            <BaselineHelpIcon size="medium" data-testid="help-icon" />
          </Box>
        }
        placement="right"
      >
```

packages/big-design/src/components/Table/spec.tsx
```diff
@@ -116,6 +116,19 @@ test('renders column with custom component', () => {
...
+  const { getByTestId } = render(
+    getSimpleTable({
+      columns: [
+        { header: 'Sku', render: ({ sku }: any) => sku },
+        { header: 'Name', tooltip: 'Some text', render: ({ name }: any) => name },
+      ],
+    }),
+  );
+
+  expect(getByTestId('help-icon')).toBeTruthy();
```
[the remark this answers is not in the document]
chanceaclark: Can we avoid an `data-testid` here? Maybe you can query by `th > svg`?

chanceaclark (commented):
Hey @bc-apostoliuk! Thanks for submitting your first PR to BigDesign! Here's a few comments that should be addressed. Reach out if you need to talk through any of these!

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -53,6 +54,17 @@ const InternalHeaderCell = <T extends TableItem>({
     );
   };
 
+  const renderTooltip = () => {
+    return (
+      <>
+        &nbsp;
+        <Tooltip trigger={<BaselineHelpIcon data-testid="help-icon" />} placement="right-end">
```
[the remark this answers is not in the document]
bc-apostoliuk: @chanceaclark We discussed it with @eugene-polev, and currently we do not have the cases when to be needed the different icon for tooltip.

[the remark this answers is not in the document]
bc-apostoliuk: @catherinevalek Yep, it should be 'right-end', but only the cases when there is not enough space on the right it should be shown below, and so on. I guess this behavior should be defined tooltip component by itself.

packages/big-design/src/components/Table/spec.tsx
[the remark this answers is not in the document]
bc-apostoliuk: @chanceaclark  We can, but how about the case when there is one more `svg` within `th` ?

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
bc-apostoliuk: @jordan-massingill  Thanks, done)

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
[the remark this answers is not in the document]
bc-apostoliuk: Done))

packages/big-design/src/components/Table/spec.tsx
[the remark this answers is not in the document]
bc-apostoliuk: @chanceaclark I have removed `data-testid`, and used `title`.

packages/docs/PropTables/TablePropTable.tsx
```diff
@@ -153,6 +153,11 @@ const tableColumnsProps: Prop[] = [
     defaultValue: 'top',
     description: 'Sets vertical alignment for column (td only).',
   },
+  {
+    name: 'tooltip',
+    types: 'string',
+    description: 'Tooltip for the table column header',
+  },
```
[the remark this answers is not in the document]
bc-apostoliuk: I have added the example

packages/docs/PropTables/TablePropTable.tsx
```diff
@@ -153,6 +153,11 @@ const tableColumnsProps: Prop[] = [
     defaultValue: 'top',
     description: 'Sets vertical alignment for column (td only).',
   },
+  {
+    name: 'tooltip',
+    types: 'string',
+    description: 'Tooltip for the table column header',
```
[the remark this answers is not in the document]
bc-apostoliuk: done.

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -1,4 +1,5 @@
-import { ArrowDownwardIcon, ArrowUpwardIcon } from '@bigcommerce/big-design-icons';
+import { Tooltip } from '@bigcommerce/big-design';
```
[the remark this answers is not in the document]
bc-apostoliuk: done

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -53,6 +55,25 @@ const InternalHeaderCell = <T extends TableItem>({
     );
   };
 
+  const renderTooltip = () => {
+    if (tooltip && typeof tooltip === 'string') {
+      return (
+        <Tooltip
+          trigger={
+            <Box as="span" marginLeft="xxSmall">
+              <BaselineHelpIcon size="medium" title="Tooltip icon" />
```
chanceaclark, line 64: 🤔 So I've been reading [this article (specifically this section)](https://inclusive-components.design/tooltips-toggletips/#inclusivetooltips) about accessibility within tooltips. From what I read, there are a couple of changes we could make:

1. Add an `id` to the `Tooltip` element, and an `aria-labelledby` to the icon element. If we do this approach, we'd need to use a UUID, luckily we have helpers for that.
2. **(Recommended)** Same sort of approach as `1.` but add `aria-describedby` to the icon element instead. You'd use this approach in conjunction with the `title` attribute, but we'd need to make the `title` attribute more descriptive than pointing out it's a "Tooltip icon".

I recommend **2.** but we either need to make the `title` configurable by the user, or have a generic phrase for what the text should read for screen readers.

packages/docs/pages/Table/TablePage.tsx
```diff
@@ -182,6 +182,25 @@ const TablePage = () => {
...
+          ]}
+          items={[
+            { sku: 'SM13', name: '[Sample] Smith Journal 13', stock: 25 },
+            { sku: 'DPB', name: '[Sample] Dustpan & Brush', stock: 34 },
+          ]}
+          stickyHeader
+        />
+        {/* jsx-to-string:end */}
+      </CodePreview>
+
```
chanceaclark, line 203: We could just add `tooltip` to the first example on the page instead of creating a new one. Seems pretty excessive. 🤔

packages/big-design/src/components/Table/spec.tsx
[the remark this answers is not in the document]
chanceaclark: Left a comment about accessibility with the `title` attribute below. Let's continue the discussion in that comment.

packages/docs/pages/Table/TablePage.tsx
bc-apostoliuk: Done)

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
bc-apostoliuk: @chanceaclark  Thanks, it looks good, but our Tooltip component does not accept id as a prop to provide it for an element in DOM. If I understood correctly `id` needs to be added to the block of `Tooltip` component. Please correct me if something is wrong.)

bc-apostoliuk: Here is an example of approach I understood.
`<Tooltip
          id={tooltipId}
          trigger={
            <Box as="span" marginLeft="xxSmall">
              <BaselineHelpIcon
                aria-describedby={tooltipId}
                size="medium"
                title={tooltip}
              />
            </Box>
          }
          placement="right"
        >
          {tooltip}
</Tooltip>`

chanceaclark: Opened a PR for the `id` stuff here:
https://github.com/bigcommerce/big-design/pull/554

We'll get that merged in today so you can use it.

bc-apostoliuk: Thats cool, thank you, I am waiting for the merging.

bc-apostoliuk: @chanceaclark done

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -53,6 +56,30 @@ const InternalHeaderCell = <T extends TableItem>({
...
+    if (typeof tooltip === 'string' && tooltip.length > 0) {
+      return (
+        <Tooltip
+          id={tooltipId}
+          trigger={
+            <Box as="span" marginLeft="xxSmall">
+              <BaselineHelpIcon
+                aria-describedby={tooltipId}
+                size="medium"
+                title={`Tooltip icon for the ${columnLabel}`}
```
chanceaclark, line 69:
```suggestion
  const renderTooltip = () => {
    if (typeof tooltip === 'string' && tooltip.length > 0) {
      return (
        <Tooltip
          id={tooltipId}
          trigger={
            <Box as="span" marginLeft="xxSmall">
              <BaselineHelpIcon
                aria-describedby={tooltipId}
                size="medium"
                title="Hover or focus for additional context."
```

Chatted with @jordan-massingill about this, and this gives the most context for someone using a screen reader. Ideally, this would be passed in via the consumer.

packages/big-design/src/components/Table/spec.tsx
```diff
@@ -116,6 +116,36 @@ test('renders column with custom component', () => {
...
+      columns: [
+        { header: 'Sku', render: ({ sku }: any) => sku },
+        { header: 'Name', tooltip: 'Some text', render: ({ name }: any) => name },
+      ],
+    }),
+  );
+
+  fireEvent.mouseOver(getByTitle('Hover or focus for additional context.'));
+
+  const result = await waitForElement(() => screen.getByText('Some text'));
```
chanceaclark, line 144: 🍹 This can be simplified using `findBy*`:
```suggestion
  const result = await screen.findByText('Some text');
```

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
bc-apostoliuk: @chanceaclark I have updated the title. Check it please.

chanceaclark (approved):
Great job!! Thanks for sticking through the reviews!

PR #554, opened 2021-06-25
feat(component): pass id to Tooltip wrapper

chanceaclark:
## What?

Passes the `id` prop to the Tooltip wrapper.

## Why?

Missing functionality that should have been there.

Screen Shot 2021-06-25 at 09 02 36

## Testing/Proof

Added unit tests and testing manually in the browser.

### DIFF
--- a/packages/big-design/src/components/Tooltip/Tooltip.tsx
+++ b/packages/big-design/src/components/Tooltip/Tooltip.tsx
@@ -14,7 +14,7 @@ export interface TooltipProps extends HTMLAttributes<HTMLDivElement> {
   inline?: boolean;
 }
 
-export const Tooltip: React.FC<TooltipProps> = memo(({ children, inline = true, modifiers, trigger, ...props }) => {
+export const Tooltip: React.FC<TooltipProps> = memo(({ children, inline = true, modifiers, trigger, id, ...props }) => {
   const [isVisible, setIsVisible] = useState(false);
   const [tooltipContainer, setTooltipContainer] = useState<HTMLDivElement | null>(null);
   const tooltipModifiers = useMemo(() => {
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Tooltip/spec.tsx
```diff
@@ -0,0 +1,20 @@
+import { fireEvent, render } from '@testing-library/react';
+import React from 'react';
+
+import { Tooltip } from './Tooltip';
+
+test('passes id to tooltip wrapper', async () => {
+  const { getByText, findByText } = render(
```
rtalvarez, line 7: I think we should start using `screen` instead of reading stuff from here. afaik that is the recommended way nowadays

chanceaclark: I thought about that, but for now I'm just using the existing pattern. There's some other things, like `user-event` we'd want to use too. But we should group that in a different PR.

packages/big-design/src/components/Tooltip/Tooltip.tsx
```diff
@@ -82,7 +82,7 @@ export const Tooltip: React.FC<TooltipProps> = memo(({ children, inline = true,
             <Popper placement={props.placement || 'top'} modifiers={tooltipModifiers}>
               {({ placement, ref, style }) =>
                 isVisible && (
-                  <StyledTooltip ref={ref} style={style} data-placement={placement}>
+                  <StyledTooltip id={props.id} ref={ref} style={style} data-placement={placement}>
```
jordan-massingill, line 85: 🍹 Why not destructure this, too?

jordan-massingill (approved):

packages/big-design/src/components/Tooltip/spec.tsx
rtalvarez: I'd disagree with that, since doing it incrementally where we can will lower the burden of changing stuff further down the road. i.e. this spec file is new but it already needs a refactor. Not blocking tho, up to you

rtalvarez (approved):

bc-apostoliuk (approved):

packages/big-design/src/components/Tooltip/spec.tsx
chanceaclark: Good thought, updated 👍

PR #555, opened 2021-06-30
fix(component): handle required attribute in MultiSelects

### LINKED ISSUE #524: MultiSelect component with 'required' prop never passes validation
**Describe the bug**
When using the `MultiSelect` component with `required` attribute, the HTML validation always triggers to say there's no selection even when there is. 

**To Reproduce**

Create MultiSelect component with required prop inside a form

```
const App = () => {
  const [value, setValue] = useState(['1']);

  return (
    <Form>
      <FormGroup>
        <MultiSelect
            label="Select Multiple"
            filterable
            options={[
              { value: "1", content: "Option 1" },
              { value: "2", content: "Option 2" },
              { value: "3", content: "Option 3" }
            ]}
            value={value}
            required={true}
            onOptionsChange={(val) => setValue(val)}
          />
      </FormGroup>
      <Button>Save</Button>
    </Form>
 );
}

```

View codesandbox example: https://codesandbox.io/s/priceless-gould-focz8?file=/src/index.tsx

**Expected behavior**

When a MultiSelect with the required prop has one or more options selected, it should pass validation. 

**Screenshots**

Screenshot 2021-04-22 at 10 01 32

**Potential Diagnosis**

It _seems_ like this is because there is always an empty HTML input within the multi-select which always has the required flag without taking into account whether there have already been some selections.  

Screenshot 2021-04-22 at 10 06 41

From digging into the code, I'm not sure of the best way to go about solving this. Perhaps it would be to force set the required prop to false on the `input` if there are chips?  

https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Input/Input.tsx#L164

**Desktop:**
 - OS: Mac OS 11.2.3
 - Browser: Chrome 89

jorgemoya:
## What?

Handle the `required` attribute we add to the input in MultiSelect, by removing it when a selection has been made.

## Why?

Fixes #524.

## Testing/Proof

Added unit tests. In addition, I locally tested this with `react-final-form` and natively to make sure validation is still being handled correctly.

### DIFF
--- a/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
@@ -369,7 +369,7 @@ export const MultiSelect = typedMemo(
             }))}
             iconRight={renderToggle}
             readOnly={!filterable}
-            required={required}
+            required={required && selectedOptions.length === 0 ? true : false}
           />
         </StyledInputContainer>
       );
--- a/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
@@ -369,7 +369,7 @@ export const MultiSelect = typedMemo(
             }))}
             iconRight={renderToggle}
             readOnly={!filterable}
-            required={required && selectedOptions.length === 0 ? true : false}
+            required={required && selectedOptions.length === 0}
           />
         </StyledInputContainer>
       );
--- a/packages/big-design/src/components/MultiSelect/spec.tsx
+++ b/packages/big-design/src/components/MultiSelect/spec.tsx
@@ -453,11 +453,26 @@ test('select should render an error if one is provided', async () => {
 });
 
 test('select should have a required attr if set as required', async () => {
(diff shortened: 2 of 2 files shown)


packages/big-design/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -369,7 +369,7 @@ export const MultiSelect = typedMemo(
             }))}
             iconRight={renderToggle}
             readOnly={!filterable}
-            required={required}
+            required={required && selectedOptions.length === 0 ? true : false}
```
jordan-massingill, line 372: 🍹 
```suggestion
            required={required && selectedOptions.length === 0}
```

jorgemoya: Doh, good idea! Initially I had a different logic so I kept the conditional 🤦

chanceaclark (approved):
Looks good to me.

@bobbyshaw does this look good?

bobbyshaw:
@chanceaclark Yep! Thanks @jorgemoya.

PR #556, opened 2021-07-05

golcinho:
## What?

## Examples
Screenshot 2021-07-05 at 14 29 15
Screenshot 2021-07-05 at 14 29 23
Screenshot 2021-07-05 at 14 29 43
Screenshot 2021-07-05 at 14 30 03
Screenshot 2021-07-05 at 14 30 20

## Docs
Screenshot 2021-07-05 at 14 23 34
Screenshot 2021-07-05 at 14 23 46

## Testing/Proof
Unit tests.

ping @chanceaclark @jorgemoya


packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const nextVisibleActions = nextState.filter(({ isVisible }) => isVisible);
+
+    if (visibleActions.length !== nextVisibleActions.length) {
+      setActionsState(nextState);
+    }
+  }, [actionsState, dropdownRef, parentRef]);
+
+  const renderedDropdown = useMemo(
+    () => (
+      <StyledFlexItem data-testid="button-group-dropdown" isVisible={isMenuVisible} ref={dropdownRef} role="listitem">
```
golcinho, line 76: These are the only reliable way we could think of for targeting them and setting the offsetWidths so that we could test the collapsing functionality (jsdom does not actually render things, so all elements have an offsetWidth of 0 in the testing environment)

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,57 @@
...
+      ? css`
+          border-radius: ${theme.borderRadius.normal};
+        `
+      : css`
+          border-radius: ${theme.borderRadius.none};
+        `}
+
+  &:focus {
+    z-index: 1;
+  }
```
golcinho, line 27: I use `z-index: 1` for correct displaying of `box-shadow` (pls see the example below). I decide to not use the value from `theme` cause it will be not semantic for my case and I think creating a new one will be redundant.


**Without `z-index`**
Screenshot 2021-04-22 at 10 29 58

**With `z-index`**
Screenshot 2021-04-22 at 10 29 46

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,153 @@
...
+      </StyledFlexItem>
+    ),
+    [actionsState, dropdownRef, isMenuVisible],
+  );
+
+  const renderedActions = useMemo(
+    () =>
+      [...actionsState]
+        .reverse()
+        .sort(({ isVisible }) => (isVisible ? -1 : 1))
```
golcinho, line 109: Here I'm sorting elements that visible elements will be first in the row. I do this for the correct working of pseudo-class `:first-of-type` which controls `border-radius` for the left side of the first element.

it's the possible case when the first element will be hidden under ellispsis when this element will have `actionType="destructive"`.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { ButtonProps } from '../Button';
+import { Dropdown } from '../Dropdown';
+import { Flex } from '../Flex';
+
+import { StyledButton, StyledFlexItem } from './styled';
+
+export interface ButtonGroupAction extends Omit<ButtonProps, 'children' | 'iconOnly' | 'iconRight' | 'iconLeft'> {
+  text: string;
+  icon?: React.ReactElement;
+}
```
golcinho, line 15: Initially, `Button` has `'iconOnly' | 'iconRight' | 'iconLeft'` properties but, according to the guidelines I can show the icon only when action will be hidden under ellipsis. So, I decide to exclude `'iconOnly' | 'iconRight' | 'iconLeft'` and just add the `icon` property.

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+export const StyledFlexItem = styled(FlexItem)<StyledFlexItemProps>`
+  margin-right: -1px;
```
golcinho, line 31: Here I'm resolving a problem with the double border. I use pixels cause we have a `1px` border in the theme package.

packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -0,0 +1,57 @@
...
+  &:last-of-type {
+    margin-right: 0;
+  }
+
+  &:first-of-type ${StyledButton} {
+    border-bottom-left-radius: ${({ theme }) => theme.borderRadius.normal};
+    border-top-left-radius: ${({ theme }) => theme.borderRadius.normal};
+  }
+
+  &:nth-last-of-type(-n + 2) ${StyledButton} {
```
golcinho, line 42: I use `&:nth-last-of-type(-n + 2)` instead of `:last-of-type` according to case when dropdown visible when some of actions has `actionType="destructive"` and all action are fit... so, it sets border radius to `dropdown toggle` and the last element of action.

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,153 @@
...
+export interface ButtonGroupProps extends HTMLAttributes<HTMLDivElement>, MarginProps {
+  actions: ButtonGroupAction[];
+}
+
+const excludeIconProps = <T extends any>({
+  iconOnly,
+  iconRight,
+  iconLeft,
+  ...actionProps
+}: T): Pick<T, Exclude<keyof T, 'iconOnly' | 'iconRight' | 'iconLeft'>> => actionProps;
```
chanceaclark, line 26: Let's not complicate this with generics:
```suggestion
const excludeIconProps = ({
  iconOnly,
  iconRight,
  iconLeft,
  ...actionProps
}: ButtonProps & { text: string }): ButtonGroupAction => actionProps;
```

I don't really like the ` & { text: string }` part 😅 , but I guess it's okay since it's an internal implementation 🤷

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
chanceaclark: Yeah, that was the same issue we ran into with the PillsTabs.

🍹 I'm :gucci: with this, but could we align it with a similar name to the `PillsTabs` one? Definitely a nitpick:
```suggestion
      <StyledFlexItem data-testid="buttongroup-dropdown-toggle" isVisible={isMenuVisible} ref={dropdownRef} role="listitem">
```

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+export const ButtonGroup: React.FC<ButtonGroupProps> = memo(({ actions, ...wrapperProps }) => {
+  const parentRef = createRef<HTMLDivElement>();
+  const dropdownRef = createRef<HTMLDivElement>();
+  const [isMenuVisible, setIsMenuVisible] = useState(false);
+  const [actionsState, setActionsState] = useState(
+    actions.map((action) => ({ isVisible: true, action: excludeIconProps(action), ref: createRef<HTMLDivElement>() })),
+  );
+
+  const changeActionsVisibility = useCallback(() => {
```
chanceaclark, line 36: 🍹 nitpick, but could we align this to a similar name to the `PillsTabs` one?
```suggestion
  const hideOverflowedActions = useCallback(() => {
```

packages/big-design/src/components/ButtonGroup/styled.tsx
chanceaclark: I'd rather not use negative margins here, could we use something like this instead or does that cause an issue?
```tsx
  &:not(:first-of-type) ${StyledButton} {
    border-left: 0;
  }
```

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,153 @@
...
+              hash: action.text.toLowerCase(),
+              icon: action.icon,
+            }))}
+          toggle={
+            <StyledButton
+              borderRadius={actionsState.every(({ isVisible }) => !isVisible)}
+              iconOnly={<MoreHorizIcon title="more" />}
+              variant="secondary"
+            />
+          }
```
chanceaclark, line 98: Should we add `placement="bottom-end"` to the dropdown?

chanceaclark (commented):
Few comments but looks really good!

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
golcinho: Thank you for your comment! Sure, I agree with you!

Done ✅

packages/big-design/src/components/ButtonGroup/styled.tsx
golcinho: Good catch! I tried to follow this way but, yeah, with this case I get the problem when buttons jumping horizontally on focus cause I need to restore the border on the focus state that buttons will look normal.

So, I think I could use `border-left: 0` with `border-color: transparent` but, we have a transition for `border-color`...  I research the same case around other libs and looks like negative margin is normal 🤷

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
golcinho: Thank you! I lil bit changed your suggestion to: `{ text: string }` => `Pick<ButtonGroupAction, 'text'>`

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
golcinho: Sure, thank you, done ✅

packages/big-design/src/components/ButtonGroup/ButtonGroup.tsx
golcinho: Good catch, thank you 🙇 Added 👍

packages/big-design/src/components/ButtonGroup/styled.tsx
chanceaclark: Ahh, good point. I'm good with this approach then 👌

packages/docs/pages/ButtonGroup/ButtonGroupPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+      {/* jsx-to-string:end */}
+    </CodePreview>
+
+    <H1>API</H1>
+    <ButtonGroupPropTable />
+
+    <H1>Examples</H1>
+    <H2>Action type destructive</H2>
+    <Text>
+      By default action with <Code>actionsType: 'destructive'</Code> hides under ellipsis.
```
chanceaclark, line 39:
```suggestion
      By default action with <Code>actionsType: 'destructive'</Code> hides under the ellipsis.
```

packages/docs/pages/ButtonGroup/ButtonGroupPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <CodePreview>
+      {/* jsx-to-string:start */}
+      <ButtonGroup
+        actions={[{ actionType: 'destructive', text: 'Button 1' }, { text: 'Button 2' }, { text: 'Button 3' }]}
+      />
+      {/* jsx-to-string:end */}
+    </CodePreview>
+
+    <H2>Icon property</H2>
+    <Text>Icon is available only for actions which is hidden under ellipsis.</Text>
```
chanceaclark, line 51:
```suggestion
    <Text>Icon is available only for actions which is hidden under the ellipsis.</Text>
```

chanceaclark (approved):
Just some grammatical cleanup but ✅

packages/docs/pages/ButtonGroup/ButtonGroupPage.tsx
golcinho: Thank you! done! ✅

packages/docs/pages/ButtonGroup/ButtonGroupPage.tsx
golcinho: Thank you! done! ✅

PR #557, opened 2021-07-06
feat(component): move Search into its own exposed component

chanceaclark:
## What?

Moves the `Search` component into its own exposed component and removes it from the `Table` component.

## Why?

TL'DR: The `search` prop wasn't actually being used when we exposed it in the alpha version.

We found while implementing search patterns, that it's easier to compose the UI with compositional components instead of prebuilt UIs. Examposing it in the alpha version gave us some insight in this matter. Also, extracting it out of the `Table` component will allow us to extend this component in the future.

screencapture-localhost-3000-search-2021-07-06-16_41_33

## Testing/Proof

- I migrated the existing tests over from the `Table` tests.
- Also, tested performance on the `StatefulTable` and it's still good.

### DIFF
--- a/packages/big-design/jest.config.js
+++ b/packages/big-design/jest.config.js
@@ -9,10 +9,10 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/setupTests.ts'],
   coverageThreshold: {
     global: {
-      statements: 95.4,
-      branches: 86.72,
-      functions: 96.73,
-      lines: 95.44,
+      statements: 95,
+      branches: 87,
+      functions: 97,
+      lines: 95,
     },
   },
 };
--- a/packages/big-design/src/components/Search/Search.tsx
+++ b/packages/big-design/src/components/Search/Search.tsx
@@ -1,13 +1,14 @@
 import { SearchIcon } from '@bigcommerce/big-design-icons';
 import React, { FormEvent } from 'react';
 
-import { Button } from '../../Button';
-import { Flex, FlexItem } from '../../Flex';
-import { Form } from '../../Form';
-import { Input } from '../../Input';
-import { TableSearch } from '../types';
+import { Button } from '../Button';
+import { Flex, FlexItem } from '../Flex';
+import { Form } from '../Form';
+import { Input } from '../Input';
 
-export const Search: React.FC<TableSearch> = ({ value, onChange, onSubmit }) => {
+import { SearchProps } from './types';
+
+export const Search: React.FC<SearchProps> = ({ value, onChange, onSubmit }) => {
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
     event.stopPropagation();
@@ -22,6 +23,7 @@ export const Search: React.FC<TableSearch> = ({ value, onChange, onSubmit }) => 
       <Flex alignItems="center" backgroundColor="white" flexDirection="row" paddingBottom="xxSmall">
         <FlexItem flexGrow={1} marginRight="small">
           <Input
+            aria-label="Search"
             placeholder="Search"
             type="search"
             value={value}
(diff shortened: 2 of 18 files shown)


packages/big-design/jest.config.js
```diff
@@ -9,10 +9,10 @@ module.exports = {
...
   coverageThreshold: {
     global: {
-      statements: 95.4,
-      branches: 86.72,
-      functions: 96.73,
-      lines: 95.44,
+      statements: 95,
+      branches: 87,
+      functions: 97,
+      lines: 95,
```
chanceaclark, line 15: Rounding them to the closest here.

packages/big-design/src/components/Search/Search.tsx
```diff
@@ -22,6 +23,7 @@ export const Search: React.FC<TableSearch> = ({ value, onChange, onSubmit }) =>
       <Flex alignItems="center" backgroundColor="white" flexDirection="row" paddingBottom="xxSmall">
         <FlexItem flexGrow={1} marginRight="small">
           <Input
+            aria-label="Search"
```
chanceaclark, line 26: Added the `aria-label` for accessibility. `placeholder` attributes will not be read by screen readers.

golcinho (approved):
Good job 👍 Great changes 👍 

**Warning**, offtopic 😄 

I think `Search` is a pretty simple composition and would propose that this component could be useless at all, cause the count of usage cases could so small, at least I saw only one case when we need to use `Input` + `Button`... I often see composition of `Select(channel picker for example` + `Input(search)` + `Button(submit changes)` and other cases where we have more elements then `Input` + `Button`.

I want to say that maybe we need to create a `Toolbar` component which will contain areas for components and compose received elements between each other according to guidelines with the right indents etc. With the `Toolbar` component we will easily create need composition of inputs, selects, and buttons and not worry about the correct spacing between these elements and the correct display of these elements on mobile devices. 

Also, the toolbar component could be useful for many parts of UI and on only tables. 

So, it just info for thinking and I'm sorry if I confuse you 😅  

These changes look great to me 👍  thank you 🙇

jorgemoya (approved):
lgtm

chanceaclark:
Yeah @golcinho I totally agree with you. I can totally either see us moving in the direction of a `Toolbar` component or completely deprecating the `Search` component. We'll see with usage how it goes, but ultimately I'd like to see another package for compound components like a `Toolbar`. Time will tell...

golcinho:
> Yeah @golcinho I totally agree with you. I can totally either see us moving in the direction of a `Toolbar` component or completely deprecating the `Search` component. We'll see with usage how it goes, but ultimately I'd like to see another package for compound components like a `Toolbar`. Time will tell...

Thank you for your answer 🙇

packages/big-design/src/components/Search/spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+import React from 'react';
+
+import { Search } from './Search';
+
+test('renders the search component', () => {
+  render(<Search value="Product" onChange={jest.fn()} onSubmit={jest.fn()} />);
+
+  const input = screen.getByLabelText('Search') as HTMLInputElement;
+
+  expect(screen.getByText('Search')).toBeInTheDocument();
```
rtalvarez, line 11: Is this the button you're trying to get? If so `screen.getByRole('button', { name: /search/i })` is a bit better

packages/big-design/src/components/Search/spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  expect(onChange).toHaveBeenCalled();
+});
+
+test('call onSubmit when user click to the Search button', () => {
+  const onSubmit = jest.fn();
+
+  render(<Search value="submit" onChange={jest.fn()} onSubmit={onSubmit} />);
+
+  fireEvent.click(screen.getByText('Search'));
```
rtalvarez, line 32: Same here

packages/big-design/src/components/Search/spec.tsx
chanceaclark: Copied those over, but definitely a good catch! Pushed the changes.

rtalvarez (approved):

PR #558, opened 2021-07-07
chore(component): remove filters prop from Table component

chanceaclark:
## What?

Removes the `filters` prop from the `Table` component.

## Why?

Similar to #557; it's better as a compositional component rather than apart of the `Table` component. Still apart of `StatefulTable`.

N/A

## Testing/Proof

Removed unit tests and ensure they worked correctly for the `StatefulTable`.


golcinho (approved):
LGTM 👍 

Only one question: should we move `filters` tests from `Table/spec.tsx` to `StatefulTable/spec.tsx`?

chanceaclark:
> Only one question: should we move `filters` tests from `Table/spec.tsx` to `StatefulTable/spec.tsx`?

@golcinho we already have similar tests for the filters in the `StatefulTable` tests.

jordan-massingill (approved):

jorgemoya (approved):

PR #561, opened 2021-07-13
fix: allow multiple instances of Worksheet

jorgemoya:
## What?

Wrap `Worksheet` in a context to keep the `store` instanced per Worksheet.

## Why?

The way it is currently implemented, with multiple worksheets in the same page, the store would break since they all would share the same info.

## Testing/Proof

Removed a bunch of tests and added a more functional way of testing the component, to prevent broken tests even if we change the implementation. Tested locally with multiple Worksheets.


rtalvarez (approved):
lgtm, but 🍹 id encourage using the methods inside `screen` as opposed to the current way; i realize the worksheet spec file isn't new tho 🤷

packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -287,6 +283,34 @@ describe('edition', () => {
     await waitForElement(() => screen.getAllByRole('combobox'));
   });
 
+  test('regains focus when it stops editing', async () => {
+    const { getByText, getByDisplayValue, getByTestId } = render(
+      <Worksheet columns={columns} items={items} onChange={handleChange} />,
+    );
+
+    const table = getByTestId('worksheet');
```
chanceaclark, line 291: Alongside @rtalvarez comments around using `screen`, could we not use `data-testid` here? I think something like `screen.getByRole('table')` should:tm: work.

chanceaclark (approved):

PR #562, opened 2021-07-15
feat: add worksheet docs

jorgemoya:
## What?

Adds Worksheet documentation

worksheet

### DIFF
--- a/packages/docs/PropTables/WorksheetPropTable.tsx
+++ b/packages/docs/PropTables/WorksheetPropTable.tsx
@@ -1,0 +1,289 @@
+import React from 'react';
+
+import { Code, NextLink, Prop, PropTable, PropTableWrapper } from '../components';
+
+const worksheetProps: Prop[] = [
+  {
+    name: 'columns',
+    types: ['TextColumn', 'NumberColumn', 'CheckboxColumn', 'SelectableColumn', 'ModalColumn'],
+    description: (
+      <>
+        See <NextLink href="#worksheet-text-column-prop-table">below</NextLink> for usage.
+      </>
+    ),
+    required: true,
+  },
+  {
+    name: 'items',
+    types: 'any[]',
+    description: 'The array of items to display.',
+    required: true,
+  },
+  {
+    name: 'onChange',
+    types: '(items: any[]) => void',
+    description: 'Returns the items that have been updated.',
+    required: true,
+  },
+  {
+    name: 'onErrors',
+    types: '(items: WorksheetError[]) => void',
+    description: (
+      <>
+        Returns an array of <NextLink href="#worksheet-error-prop-table">Error</NextLink> when an error is present.
+      </>
+    ),
+  },
+];
+
+const worksheetTextColumnProps: Prop[] = [
+  {
+    name: 'hash',
+    types: 'string',
+    required: true,
+    description: 'Unique identifier for column.',
+  },
+  {
+    name: 'header',
+    types: 'string',
+    required: true,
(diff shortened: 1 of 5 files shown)


packages/docs/PropTables/WorksheetPropTable.tsx
```diff
@@ -0,0 +1,289 @@
...
+    ),
+  },
+];
+
+const worksheetTextColumnProps: Prop[] = [
+  {
+    name: 'hash',
+    types: 'string',
+    required: true,
+    description: 'Unique identifier for column.',
```
chanceaclark, line 44: Could we describe this prop a bit more? What are it's usages et. al.?

packages/docs/PropTables/WorksheetPropTable.tsx
```diff
@@ -0,0 +1,289 @@
...
+  },
+  {
+    name: 'header',
+    types: 'string',
+    required: true,
+    description: 'Header title.',
```
chanceaclark, line 50: 🍹 Maybe make this a little bit more descriptive? Maybe you can think of something better?
```suggestion
    description: 'Header for the column.',
```

packages/docs/PropTables/WorksheetPropTable.tsx
```diff
@@ -0,0 +1,289 @@
...
+  },
+  {
+    name: 'type',
+    types: ['text'],
```
chanceaclark, line 54: I would check some of these `types` key/values to see if you can just have them as a `string` instead of an `array`.
```suggestion
    types: 'text',
```

chanceaclark (commented):
Small comments, but looks great!

packages/docs/PropTables/WorksheetPropTable.tsx
jorgemoya: It is only used internally, do we need to tell them it is used to manipulate the cells? I copied this from Table btw.

jorgemoya: I guess it is also used on the rows, so that there is some structure present?

chanceaclark: My main concern is that it's a `required` prop, without really explaining why the consumer needs to use it. I'm trying to approach this from a perspective of a consumer that doesn't really know the internal implementation here; they just need it to work, but understand a little why.

jorgemoya: I don't know if I agree with this, we rarely (if ever) indicate why an `id` is required. It can be inferred the component uses it internally.

jorgemoya: Updated for some more info, let me know what you think.

chanceaclark: Well for instance, we have the `Dropdown` component which it's items have a `hash` prop. The components implementation doesn't use `hash` at all (if you search the repo) and the description of the prop is `Stored hash of the item.`. I know that is specific to the `Dropdown`, but as a consumer, it doesn't give me any knowledge of what it's used for. For the worksheet I want to avoid stuff like this, in case we drop support for `hash` or etc.. It should be clear what the intended purpose is.

chanceaclark: What about this?
```
Unique identifier for the column value of each row. Used internally to identify and manage state values.
```

chanceaclark (approved):

PR #563, opened 2021-07-22
feat(component): add the option to disable columns in Worksheet

jorgemoya:
## What?

Add the `disabled` prop for all column types in Worksheet.

## Why?

For the scenarios when a column is only used to display a static value.

screencapture-localhost-3000-worksheet-2021-07-22-16_30_05

## Testing/Proof

Locally. Unit tests.

### DIFF
--- a/packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
+++ b/packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
@@ -40,11 +40,9 @@ const InternalModalEditor = <T extends WorksheetItem>({ cell, formatting, isEdit
           {renderedValue}
         </Small>
       </StyledFlexItem>
-      {!cell.disabled && (
-        <StyledButton onClick={handleClick} ref={buttonRef} variant="subtle">
-          Edit
-        </StyledButton>
-      )}
+      <StyledButton disabled={cell.disabled} onClick={handleClick} ref={buttonRef} variant="subtle">
+        Edit
+      </StyledButton>
     </Flex>
   );
 };
--- a/packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
+++ b/packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
@@ -36,13 +36,15 @@ const InternalModalEditor = <T extends WorksheetItem>({ cell, formatting, isEdit
   return (
     <Flex justifyContent="space-between" alignItems="center" flexWrap="wrap">
       <StyledFlexItem paddingRight="small" flexShrink={1}>
-        <Small color="secondary70" ellipsis title={renderedValue}>
+        <Small color={cell.disabled ? 'secondary50' : 'secondary70'} ellipsis title={renderedValue}>
           {renderedValue}
         </Small>
       </StyledFlexItem>
-      <StyledButton onClick={handleClick} ref={buttonRef} variant="subtle">
-        Edit
-      </StyledButton>
+      {!cell.disabled && (
+        <StyledButton onClick={handleClick} ref={buttonRef} variant="subtle">
+          Edit
+        </StyledButton>
+      )}
     </Flex>
   );
 };
--- a/packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
+++ b/packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
@@ -1,5 +1,5 @@
 import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
-import styled, { css } from 'styled-components';
+import styled from 'styled-components';
 
 export const SelectWrapper = styled.div<{ isEdited: boolean }>`
   span {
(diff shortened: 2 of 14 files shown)


packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
```diff
@@ -10,17 +10,21 @@ export const SelectWrapper = styled.div<{ isEdited: boolean }>`
...
 
   input {
     font-size: ${({ theme }) => theme.typography.fontSize.small};
     font-weight: ${({ theme }) => theme.typography.fontWeight.regular};
 
-    ${({ isEdited }) =>
-      isEdited &&
-      css`
-        background-color: ${({ theme }) => theme.colors.warning10};
-      `}
```
chanceaclark, line 23: Curious about this deletion 🤔

chanceaclark (approved):
LGTM!!! Great job!

packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
rtalvarez: +1 looks unrelated?

packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -866,4 +923,91 @@ describe('ModalEditor', () => {
 
     await waitForElement(() => screen.getAllByRole('combobox'));
   });
+
+  test('renders in disabled state', async () => {
+    render(<Worksheet columns={disabledColumns} items={items} onChange={handleChange} />);
+
+    const buttons = screen.queryAllByText('Edit');
```
rtalvarez, line 930: judging by the `buttons` name, i imagine this is a button? if so `getByRole('button', ...)` is likely a better choice here

packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -866,4 +923,91 @@ describe('ModalEditor', () => {
...
+
+    await waitForElement(() => screen.getAllByRole('combobox'));
+  });
+});
+
+describe('disable', () => {
+  test('navigation works on disabled cells', async () => {
+    render(<Worksheet columns={disabledColumns} items={items} onChange={handleChange} />);
+
+    let cell = screen.getByText('Shoes Name Three');
```
rtalvarez, line 942: If all these are cell values you can grab them by `getByRole('cell', ...)`

Same with the ones below

packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
jorgemoya: Good catch, I think this was deleted by accident, let me 👀

packages/big-design/src/components/Worksheet/spec.tsx
jorgemoya: This one is a little misleading because it's actually fetching `p` elements. `getByRole` wouldn't work.

packages/big-design/src/components/Worksheet/editors/SelectEditor/styled.ts
jorgemoya: Turns out it is no longer needed since I changed the parent `span` to `transparent` and the input `inherits` the background. I will be, however, updating the `spam` to inherit too so that the `span` and `input` always have the same color of the `td`, in case we add extra color types in the future.

packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
```diff
@@ -36,13 +36,15 @@ const InternalModalEditor = <T extends WorksheetItem>({ cell, formatting, isEdit
...
       <StyledFlexItem paddingRight="small" flexShrink={1}>
-        <Small color="secondary70" ellipsis title={renderedValue}>
+        <Small color={cell.disabled ? 'secondary50' : 'secondary70'} ellipsis title={renderedValue}>
           {renderedValue}
         </Small>
       </StyledFlexItem>
-      <StyledButton onClick={handleClick} ref={buttonRef} variant="subtle">
-        Edit
-      </StyledButton>
+      {!cell.disabled && (
```
jordan-massingill, line 43: 🍹 Should we completely not render the edit button when cell is disabled, or render it as disabled as well as the cell?

jorgemoya: Good question. This was definitely something I decided, since to me it doesn't make sense to display the `Edit` button if a cell is disabled and it can't be edited. Do you think it would be a better idea to show it?

jordan-massingill (approved):

packages/big-design/src/components/Worksheet/editors/ModalEditor/ModalEditor.tsx
jordan-massingill: Idk, I'm torn. I think it's fine without for right now, just might be good to have design input at some point.

jorgemoya: I decided to add it in and we can remove later if needed.

PR #566, opened 2021-07-27
chore(deps): upgrade to styled-components v5

chanceaclark:
## What?

Upgrade to styled-components v5.

### Note:
```
BREAKING CHANGE:
You will need to update to styled-components v5. In addition,
you will need to import the base fonts in your <head> element.
See the "Getting Started" page or README.md for an example.
```

## Why?

It's been on the backburner for quite some time. We were hoping the `styled-components` team would fix the [SSR Memory Leak issue](https://github.com/styled-components/styled-components/issues/2913) using `@import` in `createGlobalStyle`, but no luck. It also seems on their v6 roadmap they don't have intentions of fixing it either. We might as well upgrade and get it over with.

ping @bigcommerce/frontend

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,21 +14,31 @@ You can find documentation and examples on our [docs page](https://bigcommerce.g
 
 ### Quick start guide
 
-Add BigDesign and styled-components@4 to your project using `npm`:
+Add BigDesign and styled-components@5 to your project using `npm`:
 
 ```
-npm install @bigcommerce/big-design styled-components@4
+npm install @bigcommerce/big-design styled-components@5
 ```
 
 or with `yarn`:
 
 ```
-yarn add @bigcommerce/big-design styled-components@4
+yarn add @bigcommerce/big-design styled-components@5
 ```
 
-Import the `GlobalStyles` component and use it once in your app. This will set a few styles globally,
-including a base font family, [Source Sans Pro](https://fonts.google.com/specimen/Source+Sans+Pro) and [normalize.css](https://github.com/necolas/normalize.css/). We recommend placing it close to your root component.
-Then import any component, such as `Button`, to use it anywhere in your app.
+Add the font as a `<link>` in your `index.html`/`<head>` element.
+
+```html
+<head>
+  <link rel="preconnect" href="https://fonts.googleapis.com" />
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="" />
+  <link href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@200;300;400;600&display=swap" rel="stylesheet" />
+</head>
+```
+
+Import the `GlobalStyles` component and use it once in your app. This will set a few styles globally
+and add [normalize.css](https://github.com/necolas/normalize.css/). We recommend placing it close to
+your root component. Then import any component, such as `Button`, to use it anywhere in your app.
 
 ```jsx
 import { Button, GlobalStyles } from '@bigcommerce/big-design';
--- a/packages/big-design/README.md
+++ b/packages/big-design/README.md
@@ -8,21 +8,31 @@ You can find documentation and examples on our [docs page](https://bigcommerce.g
 
 ### Quick start guide
 
-Add BigDesign and styled-components@4 to your project using `npm`:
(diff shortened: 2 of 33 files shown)


packages/big-design/src/components/Modal/__snapshots__/spec.tsx.snap
```diff
@@ -314,7 +314,31 @@ exports[`render open modal without backdrop 1`] = `
 <body
   style="overflow-y: hidden;"
 >
-  <div />
+  @media (min-width:720px) {
+
+}
+
+@media (min-width:720px) {
```
deini, line 321: These look a bit odd 🤔

chanceaclark: I noticed that too... My thought is that we are including media queries based on prop values inside `css`:
```tsx
  ${({ theme, variant }) =>
    variant === 'modal' &&
    css`
      ${theme.breakpoints.tablet} {
        /* ... */
      }

      ${theme.breakpoints.desktop} {
        /* ... */
      }
    `}
```

Let me do some more digging on this though

packages/big-design/README.md
```diff
@@ -8,21 +8,29 @@ You can find documentation and examples on our [docs page](https://bigcommerce.g
...
 ```
 
-Import the `GlobalStyles` component and use it once in your app. This will set a few styles globally,
-including a base font family, [Source Sans Pro](https://fonts.google.com/specimen/Source+Sans+Pro) and [normalize.css](https://github.com/necolas/normalize.css/). We recommend placing it close to your root component.
-Then import any component, such as `Button`, to use it anywhere in your app.
+Add the font as an `@import` in your `index.html`/`<head>` element within a `<style>` tag.
+
+```html
+<style>
+  @import url('https://fonts.googleapis.com/css?family=Source+Sans+Pro:200,300,400,600');
```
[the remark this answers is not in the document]
chanceaclark: Good callout, I'll flip it over to that. Definitely more beneficial to use a `link`.

packages/big-design/src/components/Modal/__snapshots__/spec.tsx.snap
chanceaclark: So I dove a bit deeper into why this was happening, and it seems to be a [known issue](https://github.com/styled-components/jest-styled-components/issues/276) in `jest-styled-components`. It's a pretty gnarly bug on their end and after looking at their code it seems like each test is overwriting a reference somewhere. When I run this test with `.only` the snapshots look fine.

There are two courses of action for this:
1. See if we can figure out the issue in the lib and submit a fix.
2. Ignore this for now, and monitor to see if they fix it... though it's been 2 years since the issue was first submitted.

After chatting with @deini I think we'll ignore this and monitor for a fix.

deini (approved):

README.md
```diff
@@ -14,21 +14,31 @@ You can find documentation and examples on our [docs page](https://bigcommerce.g
...
-Then import any component, such as `Button`, to use it anywhere in your app.
+Add the font as a `<link>` in your `index.html`/`<head>` element.
+
+```html
+<head>
+  <link rel="preconnect" href="https://fonts.googleapis.com" />
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="" />
+  <link href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@200;300;400;600&display=swap" rel="stylesheet" />
+</head>
+```
```
jorgemoya, line 37: This sucks but nothing to do

jorgemoya (approved):
Thanks for doing this!

packages/big-design/src/components/ProgressCircle/styled.tsx
```diff
@@ -22,7 +22,6 @@ export const StyledCircle = styled.circle.attrs(({ size, theme }: any) => ({
   stroke: ${({ theme }) => theme.colors.secondary20};
 `;
 
-// @ts-expect-error TODO: Remove after upgrading to styled-components v5
```
rtalvarez, line 25: this feels goood hahah

rtalvarez (approved):

PR #567, opened 2021-07-28
test(component): update MultiSelect tests to have less timeouts

chanceaclark:
## What/why?

Update the `MultiSelect` tests to perform a bit faster. We were getting a bunch of timeout issues on the `MultiSelect` tests.

## Testing/Proof

**Before:**
```bash
@bigcommerce/big-design: PASS src/components/MultiSelect/spec.tsx (47.642s)
```

**After:**
```bash
@bigcommerce/big-design: PASS src/components/MultiSelect/spec.tsx (19.414s)
```

PR #568, opened 2021-07-29
test(component): update Select tests to have less timeouts

jorgemoya:
## What?

Refactor tests just like #567 

## Why?

Tests were too slow and causing timeouts.

## Testing/Proof

Before:
`PASS  src/components/Select/spec.tsx (57.462s)`

After:
`PASS  src/components/Select/spec.tsx (26.159s)`


rtalvarez (approved):

chanceaclark (approved):
Thanks for the help with this!

PR #569, opened 2021-07-30
feat(component): add margin props to Form component

chanceaclark:
## What?

Adds `margin` props to the `Form` component.

## Why?

I noticed a ton of cases where people need to add margin to a `Form`, for example in a `Modal`. Most cases looked like this:
```tsx
<Box marginBottom="medium">
  <Form>
    {...}
  </Form>
</Box>
```

This aims to make the DX better, so we can do this:
```tsx
<Form marginBottom="medium" />
```

Screen Shot 2021-07-30 at 10 08 08


rtalvarez (approved):
lgtm

jorgemoya (approved):

PR #570, opened 2021-07-30
feat(theme): add listReset helper

chanceaclark:
## What?

Provides a `listReset` helper on the theme.

## Why?

There have been a few use-cases where we want to build a list of components. In order to do that semantically you need a `ul/ol` with reset styles. This utility should help prevent writing the following repeatedly:
```tsx
const StyledUl = styled.ul`
  list-style-type: none;
  margin: 0;
  padding: 0;
`;
```

Now you can writing it like so:
```tsx
const StyledUl = styled.ul`
  $({ theme } => theme.helpers.listReset);
`;
```

Screen Shot 2021-07-30 at 10 38 34

### DIFF
--- a/packages/big-design-theme/src/helpers/helpers.ts
+++ b/packages/big-design-theme/src/helpers/helpers.ts
@@ -1,4 +1,5 @@
 import { em, math, rem, transparentize } from 'polished';
+import { css, FlattenSimpleInterpolation } from 'styled-components';
 
 import { themeOptions } from '../options';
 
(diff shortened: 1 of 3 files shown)


packages/big-design-theme/src/helpers/helpers.ts
```diff
@@ -7,6 +8,7 @@ export interface Helpers {
   createRGBA(color: string, alpha: number): string;
   remCalc(value: string | number): string;
   emCalc(value: string | number): string;
+  listReset: FlattenSimpleInterpolation;
```
jorgemoya, line 11: Is this the `type` returned when it simply returns the styles via `css`?

jorgemoya (approved):
lgtm

packages/big-design-theme/src/helpers/helpers.ts
chanceaclark: Yeah, `css` returns a `ReadonlyArray<...>` of values. In this case it's a `ReadonlyArray<string>`. styled-components will be able to take care of that output.

PR #578, opened 2021-08-18
fix(component): alert autoDismiss firing when not visible

chanceaclark:
## What?

Fixes the alert manager auto-dismissing hidden alerts.

## Why?

The `setTimeout` was being invoked as soon as the alert was created, which meant that the alert was already counting down to 0 before you even saw it.

https://user-images.githubusercontent.com/10539418/129981438-ece33b9a-a859-4ea5-8346-4bf1bfd95bfb.mov

## Testing/Proof

Added additional test for it.

shoutout to @jordan-massingill for catching this

### DIFF
--- a/packages/big-design/src/managers/alerts/manager.ts
+++ b/packages/big-design/src/managers/alerts/manager.ts
@@ -2,18 +2,22 @@ import { AlertProps } from '../../components/Alert';
 
 import { Subscriber, TypeMap } from './types';
 
-interface PrivateAlert extends AlertProps {
-  onClose(): void;
-}
-
 interface AddAlertConfig extends AlertProps {
   autoDismiss?: boolean;
+}
+
+interface PrivateAlert extends AddAlertConfig {
+  onClose(): void;
 }
 
 class AlertsManager {
   private alerts: PrivateAlert[] = [];
   private counter = 0;
   private subscribers: Subscriber[] = [];
+  private timeout: {
+    id?: number;
+    key?: PrivateAlert['key'];
+  } = {};
   private readonly typeMap: TypeMap = {
     error: 0,
     warning: 1,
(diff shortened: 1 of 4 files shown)


packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -2,18 +2,19 @@ import { AlertProps } from '../../components/Alert';
...
 
-interface PrivateAlert extends AlertProps {
-  onClose(): void;
-}
-
 interface AddAlertConfig extends AlertProps {
   autoDismiss?: boolean;
 }
 
+interface PrivateAlert extends AddAlertConfig {
```
chanceaclark, line 9: Extended from the `AddAlertConfig` to ensure we got the `autoDismiss`.

packages/big-design/src/utils/messagingHelpers.tsx
```diff
@@ -17,8 +17,6 @@ export interface SharedMessagingProps extends HTMLAttributes<HTMLDivElement> {
 
 export type MessagingType = 'success' | 'error' | 'warning' | 'info';
 
-export type MessagingVariant = 'alert' | 'message' | 'inline';
```
chanceaclark, line 20: Saw this when looking through types. This is unused and was residual from previous alerts/messages implementation.

jorgemoya (approved):

packages/big-design/src/managers/alerts/manager.ts
```diff
@@ -84,8 +81,27 @@ class AlertsManager {
...
 
+  private afterEvent() {
+    this.manageTimeout();
+    this.notifySubscribers();
+  }
+
+  private manageTimeout() {
+    const alert = this.alerts[0] ?? null;
+
+    if (this.timeout) {
```
deini, line 92: Won't this reset the timer for an ongoing `autoDismiss`?

Eg: `Alert 1` is displayed with `autoDismiss`, 4 seconds go by and a new alert is is added. `Alert 1` will remain for 9 seconds now. Or am I missing something?

chanceaclark: Good gatch, I assumed that newly added alert would replace the current alert. Let me think about how to alleviate this.

deini (approved):

jordan-massingill (approved):
🙌

PR #579, opened 2021-08-20
feat(component): add expandable rows

jorgemoya:
## What?

Add expandable rows functionality to Worksheet.

## Why?

Per the feature request. 

I did, however, move the buttons to their own column to handle keyboard navigation properly. I created a `ToggleEditor` component which isn't really an "Editor", but keeps it together with all the other components that render items inside the cells. This enabled me to listen to keyboard events succesfully.

Kapture 2021-08-20 at 11 23 26

## Testing/Proof

Unit tests and locally.


eugene-polev:
I'd like to propose several changes from the design perspective:
1. Change + and - icons to "chevron right" and "chevron down" accordingly
2. Make the first column narrower to hug the action icon for collapse/expand
3. Remove the border between the first column with collapse/expand action and the column with the name
4. Use white background for the header row (it creates not a very smooth transition between table header and page background - see in Figma)
5. Use a larger indent for expanded rows instead of background color (the same issue as with the header + this pattern is not scalable for cases with more than one hierarchical level)

[Figma file](https://www.figma.com/file/duwtpbP2tnLLQQmAkwtrNC/BigDesign-Lib?node-id=11359%3A32358)

image

jorgemoya:
Great feedback @eugene-polev. To maintain scope of this PR, I have updated the icons and the column size, but we should move the rest of the planned changes to a separate PR.

jennifer-lindeman:
Thanks @eugene-polev and @jorgemoya. I'm going to pass this PR since you're going to address his other feedback in another one.

jennifer-lindeman:
Currently requesting access to product-design assignment.

packages/big-design/src/components/Worksheet/Row/styled.ts
```diff
@@ -0,0 +1,9 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+export const StyledTableRow = styled.tr<{ isChild: boolean; isExpanded: boolean }>`
+  display: ${({ isExpanded }) => (isExpanded ? 'visible' : 'none')};
```
chanceaclark, line 5: `visible` isn't a valid `display` attribute. You might want to change it to `table-row`;

chanceaclark (commented):
Few comments but looks really good! Great job on this.

PR #584, opened 2021-09-10
fix(icons): add aria-hidden=true if an icon doesn't have a title 

icatalina:
## What?
Update to `polished@4`. 3.7 shows a lot of warnings.
Add `aria-hidden=true` if an icon doesn't have a title

## Why?

Icons without a title are likely just decoration and should always have
aria-hidden="true"

<table>
<tr>
	<td>Without Title
	<td>With Title
<tr>
	<td>image
	<td>image
</table>

## Testing/Proof
Tested this in Chrome/Firefox

### DIFF
--- a/packages/big-design-icons/src/flags/components/BZFlagIcon.tsx
+++ b/packages/big-design-icons/src/flags/components/BZFlagIcon.tsx
@@ -12,7 +12,13 @@ const FlagIcon: React.FC<FlagIconProps & PrivateIconProps> = ({ svgRef, title = 
   const titleId = title ? props.titleId || uniqueTitleId : undefined;
 
   return (
-    <svg viewBox="0 0 640 480" ref={svgRef} aria-labelledby={titleId} {...props}>
+    <svg
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      viewBox="0 0 640 480"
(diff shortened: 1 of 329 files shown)


icatalina:
It is a bit ridiculous that I had to update 500 snapshots for this change 😅

packages/big-design-icons/src/flags/components/BZFlagIcon.tsx
```diff
@@ -12,7 +12,13 @@ const FlagIcon: React.FC<FlagIconProps & PrivateIconProps> = ({ svgRef, title =
   const titleId = title ? props.titleId || uniqueTitleId : undefined;
 
   return (
-    <svg viewBox="0 0 640 480" ref={svgRef} aria-labelledby={titleId} {...props}>
+    <svg
+      xmlnsXlink="http://www.w3.org/1999/xlink"
```
icatalina, line 16: This icon does have some `xlink:href` which I believe require the `xmlnsXlink` attribute to be present. It seems like the new version of `svgo` takes care of this

deini (approved):

jorgemoya (approved):

chanceaclark (approved):

PR #589, opened 2021-09-13
fix(docs): dependency is an expression

deini:
## What & Why?

Fixes the logs happening on the `docs` package.
```
Critical dependency: the request of a dependency is an expression
```

image

This PR makes webpack not parse `@babel/standalone`.

### DIFF
--- a/packages/docs/next.config.js
+++ b/packages/docs/next.config.js
@@ -14,12 +14,20 @@ module.exports = {
     BD_VERSION: bdPkg.version,
   },
   webpack: (config) => {
-    config.module.rules.push({
-      test: /\.(ts|tsx)$/,
-      use: [require.resolve('jsx-to-string-loader')],
-    });
-
-    return config;
+    return {
+      ...config,
+      module: {
+        ...config.module,
+        rules: [
+          ...config.module.rules,
+          {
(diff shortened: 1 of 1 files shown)


packages/docs/next.config.js
```diff
@@ -14,12 +14,20 @@ module.exports = {
...
+      module: {
+        ...config.module,
+        rules: [
+          ...config.module.rules,
+          {
+            test: /\.(ts|tsx)$/,
+            use: [require.resolve('jsx-to-string-loader')],
+          },
+        ],
+        noParse: /@babel\/standalone/,
```
chanceaclark, line 28: IIRC this issue was getting fixed in the next version of prettier, after `v2.3.2`. So prettier `v2.4.0` should do the trick.

PR that fixed it: https://github.com/prettier/prettier/pull/11130

deini: The issue we are having is with `@babel/standalone`

Screen 2021-09-13 at 16 52 06

chanceaclark: Ah, guess I didn't dig deep enough when I looked at that 🤦

chanceaclark (approved):

PR #590, opened 2021-09-14
fix(icons): Ensure aira-hidden is correctly set in flags

icatalina:
## What?
Ensure aria-hidden is set correctly in icons.

## Why?
Because of the way we added the aria-hidden attribute, flags with default
title will still be set as `aria-hidden="true"`. This change ensures any
icon with a title is not set as hidden.

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>image
	<td>image
<tr>
  <td>n/a
	<td>image
<tr>
  <td>n/a
	<td>image
</table>

## Testing/Proof

Manually tested it in the documentation preview.

### DIFF
--- a/packages/big-design-icons/scripts/svgr-flags.config.js
+++ b/packages/big-design-icons/scripts/svgr-flags.config.js
@@ -5,6 +5,9 @@ module.exports = {
   ref: true,
   ext: 'tsx',
   plugins: ['@svgr/plugin-svgo', '@svgr/plugin-jsx', '@svgr/plugin-prettier'],
+  svgProps: {
+    'aria-hidden': '{ariaHidden}',
+  },
   template({ template }, _, { componentName, jsx }) {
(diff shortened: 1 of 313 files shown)


packages/big-design-icons/scripts/svgr-flags.config.js
```diff
@@ -31,7 +35,7 @@ module.exports = {
     };
 
     BREAK
-    const FlagIconWithForwardedRef = forwardRef<SVGSVGElement, FlagIconProps>((iconProps, ref) => <FlagIcon aria-hidden={iconProps.title ? undefined : true} {...iconProps} svgRef={ref} />);
```
icatalina, line 34: At this point the "default" flag title wasn't available, hence `aria-hidden` was set to true even though flags have always a title.

chanceaclark (approved):

deini (approved):

PR #594, opened 2021-09-16
feat(component): prevent unecessary rerender of icons in Dropdowns

jorgemoya:
## What?

Prevent unnecessary rerender of icons and content in Dropdowns and Selects.

## Why?

Performance

| Before  | After |
| ------------- | ------------- |
| Kapture 2021-09-16 at 11 02 08  | Kapture 2021-09-16 at 11 06 00  |

## Testing/Proof

Tested locally


packages/big-design/src/components/List/Item/useContent.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const iconColor = useMemo(() => {
+    if (item.disabled) {
+      return 'secondary40';
+    }
+
+    if (!isHighlighted || !('onActionClick' in item)) {
+      return 'secondary60';
+    }
+
+    return 'actionType' in item ? (item.actionType === 'destructive' ? 'danger50' : 'primary') : 'primary';
```
golcinho, line 26: 🍹 it could be simpler 👍 
```suggestion
    return 'actionType' in item && item.actionType === 'destructive' ? 'danger50' : 'primary';
```

packages/big-design/src/components/List/Item/useContent.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }, [isHighlighted, item]);
+
+  const renderIcon = useMemo(() => {
+    return (
```
golcinho, line 30: 🍹 With simple expression return is not needed 👍 

If you decide to resolve this comment, could you please also check all the same cases across the hook? if not, just close this comment, 💯 optional 👍

golcinho (approved):

chanceaclark (approved):
I'm good with this, just a little weird we are returning JSX out of a hook 🤔

jorgemoya:
> I'm good with this, just a little weird we are returning JSX out of a hook 🤔 

Good point. Let me change it.

PR #598, opened 2021-09-23
fix: fix StatefulTree prop mutation

Meztimuro:
## What?

Prevent StatefulTree prop and state mutation by replacing `.splice()` with `.slice()`.

## Why?

`.splice()` function mutates the array which in case of a prop may lead to unexpected data changes higher in the component tree.

Component behaviour should not change
image

## Testing/Proof

Ran existing tests. Functionality should not change since `.slice()` should return same data as `.splice()` without altering the original array.


chanceaclark (approved):
Thank you 🙇

golcinho (approved):

PR #601, opened 2021-09-30
chore(docs): remove routes from exportPathMap

chanceaclark:
## What?

"Removed" the `exportPathMap` definitions. In order to do this I `git mv`'d the pages to their base pathname. So `src/pages/Alert/AlertPage.tsx` became `src/pages/alert.tsx` for a flatter folder structure.

## Why?

`exportPathMap` technically wasn't needed for all the pages. The bigger issue is that when using `useRouter` `push({ query: { ... }})` was updating the route to the base name (`Alert/AlertPage`) instead of the `asPath`. This just simplifies how we do routing, and it's one less place we need to update when adding a documentation page.

N/A

## Testing/Proof

Ran the `build-static` and ran `next start` on them them. The exported files didn't include the `/dev` page, which is what we want.


packages/docs/PropTables/index.ts
```diff
@@ -17,6 +17,7 @@ export * from './InlineMessagePropTable';
 export * from './InputPropTable';
 export * from './LinkPropTable';
 export * from './MarginPropTable';
+export * from './MessagePropTable';
```
chanceaclark, line 20: This was added to clean up imports on the `Message` documentation page.

deini (approved):

jorgemoya (approved):
Basically we had to set these up for the way we structured our files and because we no longer want to do this we just flattened everything up. Is this the better approach/standard? Do we lose something by going this route?

chanceaclark:
>Basically we had to set these up for the way we structured our files and because we no longer want to do this we just flattened everything up. Is this the better approach/standard? Do we lose something by going this route?

I was chatting with @deini offline about this and the initial reason we had the pages in folders was because we were going to make a folder like structure for the side nav. I guess that ended up not happening 🤷 The best approach from what I found is not use the `exportPathMap` until you absolutely need it. And no, we didn't lose anything by this.

PR #604, opened 2021-10-05
chore: improve dx by making dev env faster

deini:
**tl;dr: This PR makes our dev environment faster.**

## What
Speed up dev environment by using HMR for changes in big-design / big-design-theme.

## Why?
Currently, we use our documentation page when developing / updating components. For every change we make in big-design / big-design-theme, babel's watcher kicks in and transpiles the whole library. Then our doc's page (next.js) detects the change and re-build the active page and is forced to do a full page reload.

This process was working fine for a while, but as the library grew, we started feeling it in our DX. It now takes some 5-10s plus a full page reload.

## How?
**tl;dr: Instead of having a separate babel watcher for big-design and big-design-theme, we now let next.js transpile them**

We solved this by using untranspiled internal typescript packages. I've seen this pattern in [the wild](https://github.com/yarnpkg/berry/blob/3e01b8e11dd90e651d4f8cacdfb73908c34e5248/packages/yarnpkg-core/package.json#L66:L69) in a couple of places. I'm not sure if this pattern has a name 🤔.

The idea is to set internal modules' `main` entrypoint in package.json to the direct `.ts` file instead of the transpiled `/dist/index.js` file. This way we can skip our babel watchers and let next.js handle it in a single step.

However, this approach comes with a few issues:

### Issue 1
Next.js won't transpile `node_modules` which we need, since we are pointing directly to the `.ts` files now. Fortunately for us, there is plugin for that [next-transpile-modules](https://www.npmjs.com/package/next-transpile-modules). This plugin allows us to define which node_modules packages to transpile.

### Issue 2
If we have a `main` entrypoint pointing to the direct `.ts` file we can't publish it to npm. Well ... we can but that will break our consumers apps. 

Yarn v2+ and pnpm package managers solve this with [publishConfig overrides](https://yarnpkg.com/configuration/manifest#publishConfig). Npm and yarn v1 do accept some overrides but they don't allow to override main / module, etc.

We could use one of those package managers, however, we currently rely on lerna for our releases and it doesn't play well with them. Even if we used them, lerna rolls their own publish and pack commands instead of directing them to the package managers. So we wouldn't get the overrides to work for free either way.

That's why we introduced a `@bigcommerce/pack` private package that runs in pre/post pack lifecycles that will manually do the overrides. This way we can use the `.ts` files directly in development and when we publish our packages, they will be changed to the normal `/dist/index.js` entry points.

### Before:

https://user-images.githubusercontent.com/2752665/136081809-6be1cbbb-30eb-4aef-84c6-176c632dae4d.mp4

### After:

https://user-images.githubusercontent.com/2752665/136080344-fa23ec30-233c-425a-b4fe-0d5fde2c84f2.mp4

## Testing/Proof

Published to a local npm registry and verify main/module were being replaced as expected.


chanceaclark (approved):

jorgemoya (commented):
🪄

jorgemoya (approved):

PR #620, opened 2021-10-08
feat(docs): update panel page to use new layout

jorgemoya:
## What?

Update Panel page to use new layout.

panel

## Testing/Proof

N/A

### DIFF
--- a/packages/docs/pages/panel.tsx
+++ b/packages/docs/pages/panel.tsx
@@ -1,54 +1,55 @@
 import { H1, Panel, Text } from '@bigcommerce/big-design';
 import React from 'react';
 
-import { CodePreview, PageNavigation } from '../components';
+import { CodePreview, List } from '../components';
 import { MarginPropTable, PanelPropTable } from '../PropTables';
 
 const PanelPage = () => {
-  const items = [
-    {
-      id: 'examples',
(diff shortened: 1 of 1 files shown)


packages/docs/PropTables/PanelPropTable.tsx
```diff
@@ -15,7 +15,7 @@ const panelProps: Prop[] = [
   },
   {
     name: 'action',
-    types: 'ButtonProps &amp; { text: string }',
+    types: 'ButtonProps & { text: string }',
```
jorgemoya, line 18: I'm assuming this is what we wanted.

packages/docs/pages/panel.tsx
```diff
@@ -1,54 +1,55 @@
...
+          contained in a panel.
+        </Text>
+        <Text bold>When to use:</Text>
+        <List>
+          <List.Item>Use seperate panels to group related content or tasks on a page.</List.Item>
+        </List>
+      </Panel>
+
+      <Panel header="Implementation" headerId="implementation">
+        <Text>The panel component is used to contain content in a structured format.</Text>
```
chanceaclark, line 24: I think this might be a little redundant based on the above description 🤷 What do you think?

PR #631, opened 2021-10-12
feat(docs): update tabs page to use new layout

jorgemoya:
## What?

Update Tabs page to use new layout.

tabs-page

## Testing/Proof

N/A

### DIFF
--- a/packages/docs/pages/tabs.tsx
+++ b/packages/docs/pages/tabs.tsx
@@ -1,68 +1,71 @@
 import { Box, H1, Panel, Tabs, Text } from '@bigcommerce/big-design';
 import React, { useState } from 'react';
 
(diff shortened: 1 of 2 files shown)


packages/docs/pages/tabs.tsx
```diff
@@ -1,68 +1,71 @@
...
-  ];
-
   return (
     <>
       <H1>Tabs</H1>
 
-      <PageNavigation items={items} />
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          <Code primary>Tabs</Code> are navigation elements at the top of the page, used to switch between different
```
chanceaclark, line 14:
```suggestion
          <Code primary>Tab</Code>'s are navigation elements at the top of the page, used to switch between different
```

PR #640, opened 2021-10-19
fix(component): change select all margin inside the table

golcinho:
## What?
Change margin for the `total selected items` in the `SelectAll` component. 

## Why?
According to guidelines.

Before | After
--- | ---
Screenshot 2021-10-19 at 10 52 15 | Screenshot 2021-10-19 at 10 51 41

## Testing/Proof
Locally + Update snapshots

ping @valterfatia


eugene-polev:
I don't think we should keep alignment between content inside the table and checkbox label in a bulk operations row. Inside the row, we have a checkbox as a separate column with its own padding and margins. It is a stand-alone control. And bulk selection checkbox has a label and should follow guidelines for checkboxes with labels.

PR #643, opened 2021-10-20
fix(component): fix checkbox alignment to label

golcinho:
## What?
Add `margin-top` & `margin-bottom` for `StyledCheckbox` to resolve problem with alignment between `checkbox` and `label`.

## Why?
I faced a problem when the checkbox is not alignments correctly to label.

**Why it happens:**
Reason: the difference between next 2 values.
- Checkbox label has `line-height` which is equal to `1.5rem(24px)`;
- `StyledCheckbox` has `height` and `width` which equal to `1.25rem(22px)`.

**Why I added `margin-bottom` property:**
- We have a case when the checkbox uses without a label so, if I will use only `margin-top`, the `checkbox` will have the wrong view (slight downward displacement), for example in the `Worksheet`, `Table` components and other same cases.

Before | After
--- | ---
Screenshot 2021-10-20 at 11 40 20 | Screenshot 2021-10-20 at 11 40 02
Screenshot 2021-10-20 at 11 38 08 | Screenshot 2021-10-20 at 11 38 25

## Testing/Proof
Locally + Snapshots

### DIFF
--- a/packages/big-design/src/components/Checkbox/__snapshots__/spec.tsx.snap
+++ b/packages/big-design/src/components/Checkbox/__snapshots__/spec.tsx.snap
@@ -75,6 +75,8 @@ exports[`render Checkbox checked 1`] = `
   -webkit-justify-content: center;
   -ms-flex-pack: center;
   justify-content: center;
+  margin-bottom: 0.0625rem;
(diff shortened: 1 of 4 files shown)


eugene-polev:
looks great!

PR #649, opened 2021-10-28
[DRAFT]fix(component): change position of selection icon on Select component

yurytut1993:
## What?

This PR updates position of check icon in select list. ([BIGDESIGN-207](https://jira.bigcommerce.com/browse/BIGDESIGN-207))

## Why?

According to design check icon should be located 8px from list item label. Also width of the list should be static and not changed after any item selection.

BEFORE
init-icon2
init-icon
AFTER
updated-icon
updated-icon2

### DIFF
--- a/packages/big-design/src/components/List/Item/Content.tsx
+++ b/packages/big-design/src/components/List/Item/Content.tsx
@@ -6,14 +6,15 @@ import { SelectAction, SelectOption } from '../../Select';
 import { Tooltip } from '../../Tooltip';
 import { Small } from '../../Typography';
 
-import { StyledLink } from './styled';
+import { StyledLabel, StyledLink } from './styled';
 
 interface ContentProps {
   item: DropdownItem | DropdownLinkItem | SelectOption<any> | SelectAction;
   isHighlighted: boolean;
+  isSelected: boolean;
 }
 
-export const Content = memo(({ item, isHighlighted }: ContentProps) => {
+export const Content = memo(({ item, isHighlighted, isSelected }: ContentProps) => {
   const iconColor = useMemo(() => {
     if (item.disabled) {
       return 'secondary40';
@@ -84,7 +85,9 @@ export const Content = memo(({ item, isHighlighted }: ContentProps) => {
             <Small color={descriptionColor(disabled)}>{description}</Small>
           </FlexItem>
         ) : (
-          content
+          <StyledLabel content={content} isSelected={isSelected}>
+            {content}
+          </StyledLabel>
         )}
       </Flex>
     );
@@ -93,7 +96,7 @@ export const Content = memo(({ item, isHighlighted }: ContentProps) => {
       'type' in item && item.type === 'link' && !disabled ? wrapInLink(item, baseContent) : baseContent;
 
     return disabled && 'tooltip' in item && item.tooltip ? wrapInTooltip(item.tooltip, finalContent) : finalContent;
-  }, [descriptionColor, item, renderIcon, wrapInLink, wrapInTooltip]);
+  }, [descriptionColor, item, renderIcon, wrapInLink, wrapInTooltip, isSelected]);
 
   return getContent;
 });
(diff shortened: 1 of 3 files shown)


andriiVitvitskyi1990 (approved):

packages/big-design/src/components/List/Item/styled.tsx
```diff
@@ -69,6 +69,19 @@ export const StyledListItem = styled.li<ListItemProps<unknown>>`
   }
 `;
 
+export const StyledLabel = styled.span<any>`
```
jorgemoya, line 72: Is this `any` necessary?

packages/big-design/src/components/List/Item/styled.tsx
```diff
@@ -69,6 +69,19 @@ export const StyledListItem = styled.li<ListItemProps<unknown>>`
...
+  margin-right: ${({ theme, isSelected }) => (isSelected ? theme.helpers.remCalc(8) : theme.helpers.remCalc(28))};
+
+  &:after {
+    display: block;
+    content: attr(content);
+    height: 0;
+    overflow: hidden;
+    visibility: hidden;
+    font-weight: 700;
+  }
```
jorgemoya, line 82: Can you help me understand why we need this?

packages/big-design/src/components/List/Item/Content.tsx
```diff
@@ -84,7 +85,9 @@ export const Content = memo(({ item, isHighlighted }: ContentProps) => {
             <Small color={descriptionColor(disabled)}>{description}</Small>
           </FlexItem>
         ) : (
-          content
+          <StyledLabel content={content} isSelected={isSelected}>
+            {content}
+          </StyledLabel>
```
jorgemoya, line 90: I can see why you're repeating `content` but seems weird to me.

jorgemoya (commented):
Thanks for looking into this! I just have some quick questions.

packages/big-design/src/components/List/Item/styled.tsx
yurytut1993: List with items should have static width according to the ticket. List width equals to the width of the biggest item. When some item activates text becomes bold and wider, so we should calculate bold text to define item width. Styles above add pseudo element with bold text. I has no height but its width help us to reserve space needed for bold text when item is active.

packages/big-design/src/components/List/Item/Content.tsx
yurytut1993: I was trying to find better way but for now this only one that works

packages/big-design/src/components/List/Item/styled.tsx
yurytut1993: No. Types will be updated. But for now I want to validate the approach. Because in case approach will be changed this code might be removed at all

packages/big-design/src/components/List/Item/styled.tsx
```diff
@@ -69,6 +69,19 @@ export const StyledListItem = styled.li<ListItemProps<unknown>>`
   }
 `;
 
+export const StyledLabel = styled.span<any>`
+  margin-right: ${({ theme, isSelected }) => (isSelected ? theme.helpers.remCalc(8) : theme.helpers.remCalc(28))};
```
yurytut1993, line 73: @jorgemoya also I have some doubts about this values. They related to check icon width but now it is not obvious. And if someday icon width will be changed I believe nobody know that this values should be also updated. I will appreciated for some ideas how to improve it.

packages/big-design/src/components/List/Item/styled.tsx
jorgemoya: Ok, thanks for the explanation!

Do you think we can use something like `content: attr(textContent)` to remove the redundant `content` prop?

jorgemoya:
Should we close this? @yurytut1993

bc-apostoliuk (approved):

golcinho:
Outdated - close. Agreed with @yurytut1993

Issue #635, opened 2021-10-13
Use best practices in our test suites


chanceaclark:
**Is your feature request related to a problem? Please describe.**
There's a few best practices we have been shirking on. First of all we should be using the `screen.*By` queries rather than using the `const { findBy* } = render()` to get elements and whatnot. We also shouldn't be using `container` to search for elements*. We should go through our test suites and clean up some of these inefficiencies.

*some conditions may apply

**Describe the solution, feature, or improvement you'd like**
Pulling some common mistakes we initially made from [Kent C. Dodds article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library).

```tsx
// From this
const { container, getByText } = render(<Table {...} />);

const rows = container.querySelectorAll('tbody > tr');
const nextPageButton = getByText('Next page');
// ...

// To this
render(<Table {...} />);

const rows = await screen.findAllByRole('row');
const nextPageButton = screen.getByRole('button', { name: /next page/i });
// ....
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #650, opened 2021-11-12
test(component): refactor  Alert component tests

bc-alexsaiannyi:
## What?

This PR is aimed to refactor Alert component's tests according to best practices. We take into account recommendations from  Kent C. Dodds [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements). For instance, using `container` for  `query`-ing elements has been removed. Instead,  we replace it with `screen`. 

## Why?

## Testing/Proof


packages/big-design/src/components/Alert/spec.tsx
```diff
@@ -35,26 +36,24 @@ test('render info Alert', () => {
 });
 
 test('renders with link', () => {
-  const { queryByRole, container } = render(
-    <Alert messages={[{ text: 'Success', link: { text: 'Link', href: '#' } }]} />,
-  );
+  const { container } = render(<Alert messages={[{ text: 'Success', link: { text: 'Link', href: '#' } }]} />);
```
chanceaclark, line 39: Let's refactor these tests to not use `container` too. For example,
```tsx
  render(<Alert messages={[{ text: 'Success', link: { text: 'Link', href: '#' } }]} />);
  
  const alert = await screen.findByRole('alert');
  
  expect(alert).toMatchSnapshot();
  // ...
```

packages/big-design/src/components/Alert/spec.tsx
```diff
@@ -35,26 +36,24 @@ test('render info Alert', () => {
...
 
   expect(container.firstChild).toMatchSnapshot();
 
-  const link = queryByRole('link') as HTMLAnchorElement;
+  const link = screen.getByRole('link') as HTMLAnchorElement;
```
chanceaclark, line 43: Let's remove these casts and use `await screen.findBy*`. You'll have to convert the test callback to be `async`.
```suggestion
  const link = await screen.findByRole<HTMLAnchorElement>('link');
```

chanceaclark (commented):
Looking good so far, but some comments to address.

packages/big-design/src/components/Alert/spec.tsx
```diff
@@ -1,97 +1,109 @@
...
-  );
+test('does not forward styles', async () => {
+  render(<Alert messages={[{ text: 'Success' }]} className="test" style={{ background: 'red' }} />);
+
+  const alert = await screen.findByRole('alert');
+  const elem = screen.queryByText('Success');
 
-  expect(container.getElementsByClassName('test').length).toBe(0);
-  expect(container.firstChild).not.toHaveStyle('background: red');
+  expect(elem).not.toHaveClass('test');
```
chanceaclark, line 107: Why are we expecting the text `Success` to not have a class here, instead of the alert role?

Even so, I believe this is changing the behavior of the test. The diff from before is testing to see if any elements in the alert role does not contain a class `test`. While this is now testing if just the text `Success` doesn't contain the class `test`. For this test, this expect is a valid use-case for using `container`, based on the intentions of the test. (The `expect(alert).not.toHaveStyle...` is still good to keep.)

I believe this is what we might want 🤔 
```suggestion
  expect(container.getElementsByclassName('test')).toHaveLength(0);
```

packages/big-design/src/components/Alert/spec.tsx
```diff
@@ -1,97 +1,109 @@
...
+  expect(screen.getByRole('button')).not.toBeUndefined();
 });
 
 test('trigger onClose', () => {
   const fn = jest.fn();
-  const { queryByRole } = render(<Alert onClose={fn} messages={[{ text: 'Success' }]} />);
+  render(<Alert onClose={fn} messages={[{ text: 'Success' }]} />);
 
-  const button = queryByRole('button') as HTMLButtonElement;
+  const button = screen.getByRole('button') as HTMLButtonElement;
```
chanceaclark, line 94:
```suggestion
  const button = await screen.findByRole('button');
```

packages/big-design/src/components/Alert/spec.tsx
```diff
@@ -1,97 +1,109 @@
...
-  const { queryByText, container } = render(<Alert header="Header" messages={[{ text: 'Success' }]} />);
+test('renders header', async () => {
+  render(<Alert header="Header" messages={[{ text: 'Success' }]} />);
+
+  const alert = await screen.findByRole('alert');
 
-  expect(container.firstChild).toMatchSnapshot();
-  expect(queryByText('Header')).not.toBeUndefined();
+  expect(alert).toMatchSnapshot();
+  expect(screen.getByRole('heading', { name: /header/i })).not.toBeUndefined();
```
chanceaclark, line 78: Let's assign these to a variable before we expect on them, along with using the `toBeDefined` method.
```tsx
  const alert = await screen.findByRole('alert');
  const heading = screen.getByRole('heading', { name: /header/i });

  expect(alert).toMatchSnapshot();
  expect(heading).toBeDefined();
```

packages/big-design/src/components/Alert/spec.tsx
```diff
@@ -1,97 +1,109 @@
...
-  const { queryByRole, container } = render(<Alert onClose={() => null} messages={[{ text: 'Success' }]} />);
+test('renders close button', async () => {
+  render(<Alert onClose={() => null} messages={[{ text: 'Success' }]} />);
+
+  const alert = await screen.findByRole('alert');
 
-  expect(container.firstChild).toMatchSnapshot();
-  expect(queryByRole('button')).not.toBeUndefined();
+  expect(alert).toMatchSnapshot();
+  expect(screen.getByRole('button')).not.toBeUndefined();
```
chanceaclark, line 87: Same here

packages/big-design/src/components/Alert/spec.tsx
bc-alexsaiannyi: @chanceaclark,  I wonder when it’s acceptable to use container at all. I’ve gone through Kent's article again and it looks for me as he doesn’t recommend to use container in general.. Maybe you can advise some article or examples? Thanks.

chanceaclark: @bc-alexsaiannyi One of the primary goals of tests is to gain confidence in our application/page/component. RTL targets that goal  in a way to write tests that mimic how normal users would interact with it, which in turn, helps us gain confidence that our application/page/component works. With that in mind, Kent recommends avoiding using `container` because using it is how a developer would interact, not a normal user. However, for this test case we are trying to test how a developer interacts with it and gain confidence that our component doesn't pass in a `className` and `style` props. So I would say this is a perfectly acceptable usage of `container`. The bottom line is that we should ask ourselves, _is using `container` helping us gain confidence in our tests_? If yes, and we know what we are doing, go right ahead.

I would say the **TL'DR** is to avoid using `container` unless you are trying to test something around how a developer would interact. And it's really only in niche cases, like this `className` stuff since we don't want developers applying custom styles to some of our components.

bc-alexsaiannyi: Thanks for explaining.

packages/big-design/src/components/Alert/spec.tsx
```diff
@@ -1,97 +1,111 @@
...
+test('renders with external link', async () => {
+  render(
     <Alert messages={[{ text: 'Success', link: { text: 'Link', href: '#', external: true, target: '_blank' } }]} />,
   );
 
-  expect(container.firstChild).toMatchSnapshot();
-
-  const link = queryByRole('link') as HTMLAnchorElement;
+  const alert = await screen.findByRole('alert');
+  const link = screen.getByRole('link') as HTMLAnchorElement;
```
chanceaclark, line 63: Let's remove this cast too.
```suggestion
  const link = await screen.fihdByRole<HTMLAnchorElement>('link');
```

chanceaclark (approved):
One last comment, otherwise looks good to me

jorgemoya (approved):

packages/big-design/src/components/Alert/spec.tsx
bc-alexsaiannyi: Totally agree with removing cast here. But It looks like we don’t have async code here so what’re your thoughts on using a `getBy*` instead?
 
`const link = screen.getByRole<HTMLAnchorElement>('link')`

chanceaclark: I guess that's alright 👍

PR #652, opened 2021-11-22
test(component): refactor Box component tests

yurytut1993:
## What?

This PR is aimed to refactor Box component's tests according to best practices. We take into account recommendations from  Kent C. Dodds [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements). For instance, using `container` for  `query`-ing elements has been removed. Instead,  we replace it with `screen`. 

## Why?

## Testing/Proof

### DIFF
--- a/packages/big-design/src/components/Box/spec.tsx
+++ b/packages/big-design/src/components/Box/spec.tsx
@@ -1,4 +1,5 @@
 import { theme } from '@bigcommerce/big-design-theme';
+import { screen } from '@testing-library/react';
 import React, { createRef } from 'react';
 import 'jest-styled-components';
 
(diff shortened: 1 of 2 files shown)


packages/big-design/src/components/Box/spec.tsx
```diff
@@ -7,91 +8,89 @@ import { render } from '@test/utils';
 import { Box } from './index';
 
 test('has margin props', () => {
-  const { container, rerender } = render(<Box />);
+  const { rerender } = render(<Box data-testid="box" />);
```
chanceaclark, line 11: Hmmm... 🤔 we should probably avoid using `data-testid` unless it's absolutely needed.

Maybe we can add some text and query that?
```tsx
  const { rerender } = render(<Box>test</Box>); // screen.getByText('test');
```

chanceaclark (commented):
Just the `data-testid` comment, but otherwise looks good to me.

packages/big-design/src/components/Box/spec.tsx
yurytut1993: I think its a good way to use the most appropriate queries. Changed all over the component and updated snapshots

chanceaclark (approved):

PR #654, opened 2021-11-24
test(component): refactor collapse component tests

bc-alexsaiannyi:
## What?

This PR is aimed to refactor Collapse component's tests according to best practices. We take into account recommendations from  Kent C. Dodds [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements). For instance, using `container` for  `query`-ing elements has been removed. Instead,  we replace it with `screen`. 

**NOTE**: userEvent package has been added for switching over fireEvent.

## Why?

## Testing/Proof

### DIFF
--- a/packages/big-design/src/components/Collapse/spec.tsx
+++ b/packages/big-design/src/components/Collapse/spec.tsx
@@ -1,7 +1,9 @@
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
 import 'jest-styled-components';
 
-import { fireEvent, render } from '@test/utils';
+import { render } from '@test/utils';
 
 import { Text } from '../Typography';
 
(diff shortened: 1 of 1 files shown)


packages/big-design/src/components/Collapse/spec.tsx
```diff
@@ -21,150 +23,159 @@ const CollapseWithVisiblePanelMock = (
...
+  expect(button.getAttribute('aria-controls')).toBe(panelId);
 });
 
 test('title button has icon', () => {
-  const { getByRole } = render(CollapseWithStaticTitleMock);
-  const trigger = getByRole('button');
-  const icon = trigger.querySelector('svg');
+  render(CollapseWithStaticTitleMock);
+
+  const icon = screen.getByRole('button').querySelector('svg');
```
chanceaclark, line 61: 🍹 Using `querySelector` to get the SVG is a little concerning. It might mean that we need to add a role to the image for accessibility, potentially along with any additional aria tags 🤔 

No action needed for this, just pointing out that if we are using `querySelector` something might be wrong.

chanceaclark (approved):

bc-apostoliuk (approved):

eugene-polev (approved):

PR #655, opened 2021-12-03
test(component): refactor Checkbox component tests

yurytut1993:
## What?

This PR is aimed to refactor Checkbox component's tests according to best practices. We take into account recommendations from  Kent C. Dodds [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements). For instance, using `container` for  `query`-ing elements has been removed. Instead,  we replace it with `screen`. 

## Why?

## Testing/Proof

### DIFF
--- a/packages/big-design/src/components/Checkbox/spec.tsx
+++ b/packages/big-design/src/components/Checkbox/spec.tsx
@@ -1,7 +1,9 @@
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React, { createRef } from 'react';
 import 'jest-styled-components';
 
-import { fireEvent, render } from '@test/utils';
+import { render } from '@test/utils';
 
 import { warning } from '../../utils';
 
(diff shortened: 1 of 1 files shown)


packages/big-design/src/components/Checkbox/spec.tsx
```diff
@@ -54,115 +56,102 @@ describe('render Checkbox', () => {
...
 });
 
 test('triggers onChange when clicking styled and text label', () => {
   const onChange = jest.fn();
-  const { container } = render(<Checkbox label="Checked" checked={true} onChange={onChange} data-testid="checkbox" />);
 
-  const labels = container.querySelectorAll('label');
+  render(<Checkbox label="Checked" checked={true} onChange={onChange} />);
+
+  const labels = screen.getByRole<HTMLInputElement>('checkbox').parentElement!.querySelectorAll('label');
```
chanceaclark, line 120: Could this be `screen.getAllByRole('label')`?

chanceaclark (commented):
Looks good, however one minor comment.

packages/big-design/src/components/Checkbox/spec.tsx
yurytut1993: I was also looking on similar approach but looks like such role does not exist

icatalina: what about `getByText(/checked/)`? might still be better than all this... 🤷

yurytut1993: In this case we will not catch `label` with `svg` inside

chanceaclark: Totally forgot there is the `screen.getByLabelText('')` query 🤦

yurytut1993: Query above may return an input associated with the label, but we need to collect all labels and then iterate through them. Can you provide some code according to our component? Maybe I have wrong understanding of your idea

golcinho:
Outdated 👍

PR #658, opened 2021-12-08
test(component): refactor Counter component tests

bc-alexsaiannyi:
This PR is aimed to refactor Counter component's tests according to best practices. We take into account recommendations from  Kent C. Dodds [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-container-to-query-for-elements). For instance, using `container` for  `query`-ing elements has been removed. Instead,  we replace it with `screen`. 

## Why?

## Testing/Proof

### DIFF
--- a/packages/big-design/src/components/Counter/spec.tsx
+++ b/packages/big-design/src/components/Counter/spec.tsx
@@ -1,320 +1,330 @@
-import React, { createRef, Ref } from 'react';
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React, { ComponentProps, createRef, useState } from 'react';
 import 'jest-styled-components';
 
-import { fireEvent, render } from '@test/utils';
+import { render } from '@test/utils';
 
 import { FormControlDescription, FormControlError, FormControlLabel, FormGroup } from '../Form';
 
-import { Counter, CounterProps } from './index';
+import { Counter } from './index';
 
-interface MockCounterProps extends CounterProps {
-  dataTestId?: string;
-  ref?: Ref<HTMLInputElement>;
-}
+const CounterMock: React.FC<Partial<ComponentProps<typeof Counter>>> = ({ value = 5, ...props }) => {
+  const [counterValue, setCounterValue] = useState(value);
+  const handleChange = (value: number) => setCounterValue(value);
 
-const val = 5;
-
-const handleChange = jest.fn((num) => num);
-
-const requiredAttributes = {
-  value: val,
-  onCountChange: handleChange,
+  return (
+    <Counter
+      label="Label"
+      error=""
+      description=""
+      value={counterValue}
+      min={0}
+      max={100}
+      onCountChange={props.onCountChange ?? handleChange}
+      {...props}
+    />
+  );
 };
-
-const counterMock = ({
-  ref,
-  label = 'Label',
-  labelId = '',
-  id = '',
-  error = '',
-  description = '',
-  value,
-  onCountChange,
-  min = 0,
-  max = 10,
-  dataTestId = '',
-}: MockCounterProps) => (
-  <Counter
-    ref={ref}
-    label={label}
-    labelId={labelId}
-    error={error}
-    id={id}
-    description={description}
-    value={value}
-    data-testid={dataTestId}
-    min={min}
-    max={max}
-    onCountChange={onCountChange}
-  />
-);
 
 test('forwards ref', () => {
   const ref = createRef<HTMLInputElement>();
-  const { container } = render(counterMock({ ref, ...requiredAttributes }));
-  const counter = container.querySelector('input');
 
-  expect(ref.current).toBe(counter);
+  render(<Counter value={5} onCountChange={jest.fn()} ref={ref} />);
+
+  const input = screen.getByRole('textbox');
+
+  expect(ref.current).toBe(input);
 });
 
 test('renders a counter as an input tag', () => {
-  const { container } = render(counterMock(requiredAttributes));
+  render(<CounterMock value={5} />);
 
-  expect(container.querySelector('input')).toBeInTheDocument();
+  expect(screen.getByLabelText(/Label/i)).toBeInTheDocument();
 });
 
 test('renders a counter with matched label', () => {
-  const { queryByLabelText } = render(counterMock({ label: 'Test Label', ...requiredAttributes }));
+  render(<CounterMock value={5} label="Test Label" />);
 
   // This one checks for matching id and htmlFor
-  expect(queryByLabelText('Test Label')).toBeInTheDocument();
+  expect(screen.getByLabelText(/test label/i)).toBeInTheDocument();
 });
 
 test('create unique ids if not provided', () => {
-  const { queryByTestId } = render(
+  render(
     <>
-      {counterMock({ dataTestId: 'item1', label: 'Test Label', ...requiredAttributes })}
-      {counterMock({ dataTestId: 'item2', label: 'Test Label', ...requiredAttributes })}
+      <CounterMock value={5} label="Label First" />
+      <CounterMock value={5} label="Label Second" />
     </>,
   );
 
-  const item1 = queryByTestId('item1') as HTMLInputElement;
-  const item2 = queryByTestId('item2') as HTMLInputElement;
+  const item1 = screen.getByLabelText(/label first/i);
+  const item2 = screen.getByLabelText(/label second/i);
 
-  expect(item1).toBeDefined();
-  expect(item2).toBeDefined();
   expect(item1.id).not.toBe(item2.id);
 });
 
 test('respects provided id', () => {
-  const { container } = render(counterMock({ id: 'test', label: 'Test Label', ...requiredAttributes }));
-  const counter = container.querySelector('#test') as HTMLInputElement;
+  const providedIdValue = 'test';
 
-  expect(counter.id).toBe('test');
+  render(<CounterMock value={5} id={providedIdValue} label="Test Label" />);
+
+  const counter = screen.getByLabelText(/test label/i);
+
+  expect(counter).toHaveAttribute('id', providedIdValue);
 });
 
 test('matches label htmlFor with id provided', () => {
-  const { container } = render(counterMock({ id: 'test', label: 'Test Label', ...requiredAttributes }));
-  const label = container.querySelector('label') as HTMLLabelElement;
+  const providedValue = 'test';
 
-  expect(label.htmlFor).toBe('test');
+  render(<CounterMock value={5} id={providedValue} label="Test Label" />);
+
+  const label = screen.getByText<HTMLLabelElement>(/test label/i);
+
+  expect(label).toHaveAttribute('for', providedValue);
 });
 
 test('respects provided labelId', () => {
-  const { container } = render(counterMock({ label: 'Test Label', labelId: 'test', ...requiredAttributes }));
-  const label = container.querySelector('#test') as HTMLLabelElement;
+  const providedValue = 'test';
+  render(<CounterMock value={5} labelId={providedValue} label="Test Label" />);
 
-  expect(label.id).toBe('test');
+  const label = screen.getByText<HTMLLabelElement>(/test label/i);
+
+  expect(label).toHaveAttribute('id', providedValue);
 });
 
 test('renders a description', () => {
-  const descriptionText = 'This is a description';
-  const { queryByText } = render(counterMock({ description: descriptionText, ...requiredAttributes }));
+  const description = 'This is a description';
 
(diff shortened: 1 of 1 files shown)


packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -346,15 +363,16 @@ describe('error does not show when invalid type', () => {
       </div>,
     ];
 
-    const { queryByTestId } = render(<FormGroup>{counterMock({ error: errors, ...requiredAttributes })}</FormGroup>);
+    render(<FormGroup>{<CounterMock value={5} error={errors} />}</FormGroup>);
```
icatalina, line 366: No need for the `{}`

```suggestion
    render(<FormGroup><CounterMock value={5} error={errors} /></FormGroup>);
```

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
+  const { rerender } = render(<FormGroup>{<CounterMock value={5} error="" />}</FormGroup>);
 
-  expect(container.querySelector('[class*="StyledError"]')).not.toBeInTheDocument();
+  expect(screen.queryByText('Error')).not.toBeInTheDocument();
 
-  rerender(<FormGroup>{counterMock({ error, ...requiredAttributes })}</FormGroup>);
+  rerender(<FormGroup>{<CounterMock value={5} error={error} />}</FormGroup>);
 
-  expect(container.querySelector('[class*="StyledError"]')).toBeInTheDocument();
+  expect(screen.getByText(error).parentElement).toBeInTheDocument();
```
icatalina, line 324: If `error` is in the document, `parentElement` should also be in the document, is it really necessary to go up one level?

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
 
 test('value increases when increase or decrease icons are clicked', () => {
-  const { getByDisplayValue, container } = render(counterMock(requiredAttributes));
+  const handleChange = jest.fn((num: number) => num);
+  render(<CounterMock value={5} onCountChange={handleChange} />);
 
-  const counter = getByDisplayValue('5') as HTMLInputElement;
-
-  const icons = container.getElementsByTagName('svg');
+  const counterValue = screen.getByRole<HTMLInputElement>('textbox').value;
```
icatalina, line 257: Instead of specifying `<HTMLInputElement>`, grab the whole element and expect with `toHaveValue()`

```tsx
const counter = screen.getByRole<HTMLInputElement>('textbox');

expect(counter).toHaveValue('5');
```

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
+  const icons = Array.from(screen.getAllByTitle(/[a-z]{2}crease count/i)).map(
+    (title: HTMLElement): HTMLElement | null => title.parentElement,
+  );
 
-  expect(counter.value).toEqual('5');
-  fireEvent.click(icons[1]);
+  expect(counterValue).toEqual('5');
+  icons[1] && userEvent.click(icons[1]);
```
icatalina, line 263: `if`s are not recommended, in other repos we've created a helper that throws when `undefined` or `null`.

```ts
export function assertDefined<T>(a: T | undefined | null): asserts a is T {
  if (a === undefined || a === null) {
    throw new Error('Element not found');
  }
}
```

and then in your tests, you can just use them:

```ts

const [decrease, increase] = Array.from(screen.getAllByTitle(/[a-z]{2}crease count/i)).map(
  (title) => title.parentElement,
);

assertDefined(increase);
assertDefined(decrease);

userEvent.click(increase);

expect(handleChange).toHaveBeenCalledWith(6);

userEvent.click(decrease);
expect(handleChange).toHaveBeenCalledWith(4);
```

A much better alternative will be to just get each button independently:

```ts
const increaseButton = screen.getByRole('button', { name: /increase count/i });
const decreaseButton = screen.getByRole('button', { name: /decrease count/i });
```

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
   expect(handleChange).toHaveBeenCalledWith(4);
 });
 
 test('value increases and decreases with arrow keypresses', () => {
-  const { getByDisplayValue } = render(counterMock(requiredAttributes));
+  const handleChange = jest.fn((num: number) => num);
+  render(<CounterMock value={5} onCountChange={handleChange} />);
 
-  const counter = getByDisplayValue('5') as HTMLInputElement;
+  const counter = screen.getByRole<HTMLInputElement>('textbox');
```
icatalina, line 274: I don't think the type (`HTMLInputElement`) is necessary anywhere in the file, consider removing it.

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
 });
 
 test('create unique ids if not provided', () => {
-  const { queryByTestId } = render(
+  render(
     <>
-      {counterMock({ dataTestId: 'item1', label: 'Test Label', ...requiredAttributes })}
-      {counterMock({ dataTestId: 'item2', label: 'Test Label', ...requiredAttributes })}
+      {<CounterMock value={5} dataTestId="item1" label="Test Label" />}
+      {<CounterMock value={5} dataTestId="item2" label="Test Label" />}
```
icatalina, line 87: `{}` are only required when calling a method, in this case we are just rendering components:


```suggestion
      <CounterMock value={5} dataTestId="item1" label="Test Label" />
      <CounterMock value={5} dataTestId="item2" label="Test Label" />
```

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
+      id={id}
+      description={description}
+      value={counterValue}
+      data-testid={dataTestId}
+      min={min}
+      max={max}
+      onCountChange={onCountChange ?? handleChange}
+    />
+  );
+};
```
icatalina, line 58: I wonder if this is better written this way:

```tsx
const CounterMock: React.FC<Partial<ComponentProps<typeof Counter>>> = ({ value = 5, ...props }) => {
  const [counterValue, setCounterValue] = useState(value);
  const handleChange = (value: number) => setCounterValue(value);

  return (
    <Counter
      label="Label"
      error=""
      description=""
      value={counterValue}
      min={0}
      max={100}
      onCountChange={handleChange}
      {...props}
    />
  );
};
```

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
     </>,
   );
 
-  const item1 = queryByTestId('item1') as HTMLInputElement;
-  const item2 = queryByTestId('item2') as HTMLInputElement;
+  const item1 = screen.getByTestId<HTMLInputElement>('item1');
+  const item2 = screen.getByTestId<HTMLInputElement>('item2');
 
   expect(item1).toBeDefined();
   expect(item2).toBeDefined();
```
icatalina, line 95: When using `getBy*` they must be defined, no need to double check.

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
-      {counterMock({ dataTestId: 'item2', label: 'Test Label', ...requiredAttributes })}
+      {<CounterMock value={5} dataTestId="item1" label="Test Label" />}
+      {<CounterMock value={5} dataTestId="item2" label="Test Label" />}
     </>,
   );
 
-  const item1 = queryByTestId('item1') as HTMLInputElement;
-  const item2 = queryByTestId('item2') as HTMLInputElement;
+  const item1 = screen.getByTestId<HTMLInputElement>('item1');
+  const item2 = screen.getByTestId<HTMLInputElement>('item2');
```
icatalina, line 92: No need to cast them to `HTMLInputElement`, all `HTMLElements` have `id`s

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
-  const counter = container.querySelector('#test') as HTMLInputElement;
+  const providedIdValue = 'test';
+
+  render(<CounterMock value={5} id={providedIdValue} label="Test Label" />);
 
-  expect(counter.id).toBe('test');
+  const counterId = screen.getByLabelText<HTMLInputElement>(/test label/i).id;
+
+  expect(counterId).toBe(providedIdValue);
 });
```
icatalina, line 107: Prefer 

```ts
const counter = screen.getByLabelText(/test label/i);

expect(counter).toHaveAttribute('id', providedIdValue);
```

If there's an error, this will show a better error message.

image

vs

image

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,335 @@
...
 
 test('matches label htmlFor with id provided', () => {
-  const { container } = render(counterMock({ id: 'test', label: 'Test Label', ...requiredAttributes }));
-  const label = container.querySelector('label') as HTMLLabelElement;
+  render(<CounterMock value={5} id="test" label="Test Label" />);
+
+  const labelForAttr = screen.getByText<HTMLLabelElement>(/test label/i).htmlFor;
```
icatalina, line 112: Same as the one before, prefer `toHaveAttribute`

icatalina (commented):
Hey @bc-alexsaiannyi!

Thanks for this PR, looking good, I just added a few pointer, let me know what you think about them.

🙇

packages/big-design/src/components/Counter/spec.tsx
bc-alexsaiannyi: Noted, I haven't thought about it in this way.

packages/big-design/src/components/Counter/spec.tsx
bc-alexsaiannyi: yeah, totally agree.  it looks much cleaner.

packages/big-design/src/components/Counter/spec.tsx
bc-alexsaiannyi: Thanks about fn `assertDefine`. I will keep it in mind. but in this case your suggestion is much better. changed)

packages/big-design/src/components/Counter/spec.tsx
bc-alexsaiannyi: Yeah,  removed. sorry about this.

bc-alexsaiannyi:
Hey @icatalina. I've added changes. Thank you for going through and adding all your suggestions.

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,330 @@
...
-
 test('forwards ref', () => {
   const ref = createRef<HTMLInputElement>();
-  const { container } = render(counterMock({ ref, ...requiredAttributes }));
-  const counter = container.querySelector('input');
 
-  expect(ref.current).toBe(counter);
+  render(<Counter value={5} onCountChange={jest.fn()} ref={ref} />);
+
+  const input = screen.getByRole<HTMLInputElement>('textbox');
```
icatalina, line 35: Do we need to cast to `HTMLInputElement`?

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,330 @@
...
-  expect(buttons[1]).toHaveProperty('disabled', true);
+  expect(buttons[0]).toBeEnabled;
+  expect(buttons[1]).toBeDisabled;
 });
 
 test('value prop only accepts whole numbers', () => {
-  const { container } = render(counterMock({ max: 20, ...requiredAttributes }));
+  render(<CounterMock value={5} />);
+
+  const input = screen.getByRole<HTMLInputElement>('textbox');
```
icatalina, line 227: No need to cast to `HTMLInputElement`:

image

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -1,320 +1,330 @@
...
   const error = 'Error';
-  const { container, rerender } = render(<FormGroup>{counterMock({ error: '', ...requiredAttributes })}</FormGroup>);
+  const { rerender } = render(
+    <FormGroup>
+      <CounterMock value={5} error="" />
+    </FormGroup>,
+  );
 
-  expect(container.querySelector('[class*="StyledError"]')).not.toBeInTheDocument();
+  expect(screen.queryByText('Error')).not.toBeInTheDocument();
```
icatalina, line 307: Why do we swap between `error` and `'Error'`?

icatalina (commented):
Looking good, just a few small comments :)

chanceaclark:
I'm going to close this since I already did most of it in #1187

PR #665, opened 2022-01-13
feat(docs): update breakpoints page to use new layout

chanceaclark:
## What?
Update Breakpoints page to use new layout.

screencapture-localhost-3000-breakpoints-2022-01-14-11_38_36
Screen Shot 2022-01-14 at 11 38 50
Screen Shot 2022-01-14 at 11 39 04

## Testing/Proof

N/A

### DIFF
--- a/packages/docs/components/Code/styled.tsx
+++ b/packages/docs/components/Code/styled.tsx
@@ -10,7 +10,7 @@ export const StyledCode = styled.code<CodeProps>`
     highlight &&
     !primary &&
     css`
-      background-color: ${theme.colors.warning10};
+      background-color: ${theme.colors.warning20};
       padding: ${theme.spacing.xxSmall};
     `};
 
--- a/packages/docs/pages/breakpoints.tsx
+++ b/packages/docs/pages/breakpoints.tsx
@@ -1,63 +1,133 @@
-import { Box, H1, Panel, Text } from '@bigcommerce/big-design';
+import { BoxProps, Box as Container, H1, Message, Panel, Text } from '@bigcommerce/big-design';
 import { breakpointValues } from '@bigcommerce/big-design-theme';
 import React from 'react';
 import styled from 'styled-components';
 
-import { Code, CodePreview } from '../components';
+import { Code, CodePreview, ContentRoutingTabs, GuidelinesTable } from '../components';
+
+const Box: React.FC<BoxProps> = ({ children, ...props }) => (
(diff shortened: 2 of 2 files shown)


packages/docs/components/Code/styled.tsx
```diff
@@ -10,7 +10,7 @@ export const StyledCode = styled.code<CodeProps>`
     highlight &&
     !primary &&
     css`
-      background-color: ${theme.colors.warning10};
+      background-color: ${theme.colors.warning20};
```
chanceaclark, line 13: Updated to be just a little bit brighter

packages/docs/pages/breakpoints.tsx
```diff
@@ -3,61 +3,94 @@ import { breakpointValues } from '@bigcommerce/big-design-theme';
...
+        ]}
+      />
+    </Panel>
 
-          return <StyledBox backgroundColor="secondary20" />;
-        }}
-        {/* jsx-to-string:end */}
-      </CodePreview>
+    <Panel header="Do's and Don'ts" headerId="guidelines">
+      <GuidelinesTable recommended={['Use built in responsive props, where applicable.']} discouraged={[]} />
```
chanceaclark, line 93: Let me know if there might be additions to this.

golcinho (approved):

jorgemoya (commented):
Should we include an example of how to breakpoint inline in props of existing components, like the spacing props? 

Maybe add a Do to use inline breakpoint values when possible (instead of extending)?

chanceaclark:
> Should we include an example of how to breakpoint inline in props of existing components, like the spacing props?

So I thought about this, but we show examples in each of the spacing props on how to use it. My rationale for adding another example is that a user may not see the other example, so probably showing an example here might be beneficial 🤷‍♂️ 

> Maybe add a Do to use inline breakpoint values when possible (instead of extending)?

So I did add `Use built in responsive props, where applicable.`. To me that's pretty similar 🤔

jorgemoya:
> So I did add `Use built in responsive props, where applicable.`. To me that's pretty similar 🤔

Ah yeah, that is good enough. 

I'm conflicted because if I was a user of BigDesign and I wanted to remember the correct syntax for using built in props I would look here first. 🤔

chanceaclark:
I'll go ahead an add an example then!

jorgemoya (approved):
Perfect

chanceaclark:
Those last two force pushes were to fix some grammatical errors I found.

PR #666, opened 2022-01-14
feat(docs): update flex page to use new layout

jorgemoya:
## What?

Update to new layout.

screencapture-localhost-3000-flex-2022-01-14-11_43_09


packages/docs/pages/flex.tsx
```diff
@@ -24,153 +24,225 @@ const ExampleBox: React.FC<{ vertical?: boolean }> = ({ children, vertical }) =>
...
+          recommended={[
+            <>
+              Use <Code primary>Flex</Code> when focusing on content flow.
+            </>,
+            <>
+              Use <Code primary>FlexItem</Code> when children need specific flex properties.
+            </>,
+            <>
+              Use <Link href="https://css-tricks.com/snippets/css/a-guide-to-flexbox/">CSS-Tricks complete guide</Link>{' '}
+              to <Code primary>Flexbox</Code>.
```
chanceaclark, line 232: I don't think we need to wrap it in a `<Code>` element when it's referring to the idea of "flexbox".
```suggestion
              to flexbox.
```

packages/docs/pages/flex.tsx
```diff
@@ -24,153 +24,225 @@ const ExampleBox: React.FC<{ vertical?: boolean }> = ({ children, vertical }) =>
...
+            </>,
+          ]}
+          discouraged={[
+            <>
+              Don’t use <Code primary>Flex</Code> when focusing on content placement. Use{' '}
+              <NextLink href="grid">Grid</NextLink> instead.
```
chanceaclark, line 238: 🍹 I was reading up on the MDN and CSS Tricks documentation of flexbox. They mention some things that might be good to include in this description:
- CSS Tricks pointed out that flexbox should be used for simple layouts while grid should be used for more complex layout.
- MDN made a point to think of flexbox as 1-dimensional, while grid is 2-dimension. I think that's a great way of looking at the two.

packages/docs/pages/flex.tsx
```diff
@@ -24,153 +24,225 @@ const ExampleBox: React.FC<{ vertical?: boolean }> = ({ children, vertical }) =>
...
-            </CodePreview>
-          </Panel>
-          <Panel header="Column layout">
-            <Text>
-              One way of creating a column based layout is using a combination of the <Code primary>flexWrap</Code> prop
-              on the flex container with an additional <Code primary>flexBasis</Code> prop on the flex items.
-            </Text>
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          A <Code primary>Flex</Code> container used for customizable layouts.
```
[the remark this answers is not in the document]
jorgemoya: This sounds good to me

packages/docs/pages/flex.tsx
jorgemoya: I was reading up about this as there is disagreement between the dimensional distinction, since both can be used for 1 or 2 dimensions, and a more agreeable distinction is to use grid for content placement (to place things exactly were you want to) while flex is more "flow-ish" in nature.

packages/docs/pages/flex.tsx
```diff
@@ -24,153 +24,225 @@ const ExampleBox: React.FC<{ vertical?: boolean }> = ({ children, vertical }) =>
...
-            <Text>
-              Using the <Code primary>flexOrder</Code> prop you can reorganize content based on the size generated by
-              setting the <Code primary>flexGrow</Code> prop. <Code primary>flexGrow</Code> allows the flex items to
-              grow relative to the flex container's width.
-            </Text>
+        <Text bold>When to use:</Text>
+        <List>
+          <List.Item>Want to lay a collection of items out in one direction or another.</List.Item>
+          <List.Item>
+            Want to control the dimensions of the items in that one dimension, or control the spacing between items.
```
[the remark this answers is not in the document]
jorgemoya: I pulled these from [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Typical_Use_Cases_of_Flexbox). I agree with it, since you can definitely control the dimension by setting how much space each element should take in an unknown container.

packages/docs/pages/flex.tsx
```diff
@@ -24,153 +24,227 @@ const ExampleBox: React.FC<{ vertical?: boolean }> = ({ children, vertical }) =>
...
-          </Panel>
-          <Panel header="Column layout">
-            <Text>
-              One way of creating a column based layout is using a combination of the <Code primary>flexWrap</Code> prop
-              on the flex container with an additional <Code primary>flexBasis</Code> prop on the flex items.
-            </Text>
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          The <Code primary>Flex</Code> component extends the <Code primary>Box</Code> utility component and provides
+          CSS flexbox container alongside flexbox properties.
```
chanceaclark, line 34:
```suggestion
          a CSS flexbox container alongside flexbox properties.
```

chanceaclark (approved):

packages/docs/pages/flex.tsx
chanceaclark: Maybe we should also link to it 🤔 
```suggestion
          <Link href="https://developer.mozilla.org/en-US/docs/Web/CSS/flex" external target="_blank">CSS flexbox</Link> container alongside flexbox properties.
```

Issue #664, opened 2022-01-13
PillTabs does not support items changing


bobbyshaw:
**Describe the bug**
I was trying to use the PillTabs component to filter a table. The items prop is populated through an external call (e.g. fetch order statuses that can then be used within the PillTabs.

When the items array changes length an array access error occurs: "Cannot read properties of undefined (reading 'ref')"

**To Reproduce**
See example at https://codesandbox.io/s/bigcommerce-react-pilltabs-djef3?file=/src/index.tsx

1. Use PillTabs component 
2. Change length of items array provided to PillsTab, e.g. with useEffect hook
4. See error at https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/PillTabs/PillTabs.tsx#L108

**Expected behavior**
Can dynamically change the items without error.

**Screenshots**
![Screenshot 2022-01-13 at 20 45 26](https://user-images.githubusercontent.com/553566/149406671-4d320a43-1515-4b9a-8675-c02406519aa6.png)

 - OS: MacOS 11.6
 - Browser Chrome
 - Version 97

**Additional information**

It seems that [`pillsState` is initiated](https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/PillTabs/PillTabs.tsx#L26) but then not updated when items change .

rtalvarez:
Thanks for the report! I'll look into it

rtalvarez:
Please try using the `0.30.3-alpha.1` version and see if it works!

bobbyshaw:
@rtalvarez That works. Thanks!

PR #667, opened 2022-01-14
fix(component): allows pill tabs to receive updated items

rtalvarez:
## What?
Accounts for a use case where users need to update the `items` prop in `PillTabs` 

Addresses https://github.com/bigcommerce/big-design/issues/664

## Why?
Filters might be computed from async sources and thus the prop needs to be able to handle updates

## How?
Adding a `useEffect` path where it checks for id mismatches between `items` and the internal state. This handles the case where pills are added, removed, or swapped

## Testing/Proof
- Verified adding/removing pills via async actions (setTimeout)
- Verified pill swapping works
- Verified scenario via unit tests

### DIFF
--- a/packages/big-design/src/components/PillTabs/PillTabs.tsx
+++ b/packages/big-design/src/components/PillTabs/PillTabs.tsx
@@ -99,27 +99,57 @@ export const PillTabs: React.FC<PillTabsProps> = ({ activePills, items, onPillCl
     );
   }, [items, pillsState, isMenuVisible, dropdownRef, activePills, onPillClick]);
 
+  useEffect(() => {
+    if (items.length !== pillsState.length) {
+      const newState = items.map((item) => {
+        const oldItem = pillsState.filter((stateItem) => stateItem.item === item)[0];
+
+        if (oldItem) {
+          return oldItem;
+        }
+
+        return {
+          // hideOverflownPills will correct this field if it needs correction
+          isVisible: true,
+          item,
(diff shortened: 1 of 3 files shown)


packages/big-design/src/components/PillTabs/PillTabs.tsx
```diff
@@ -99,27 +99,60 @@ export const PillTabs: React.FC<PillTabsProps> = ({ activePills, items, onPillCl
...
   }, [items, pillsState, isMenuVisible, dropdownRef, activePills, onPillClick]);
 
+  useEffect(() => {
+    const itemIds = items.map((item) => item.id);
+    const stateIds = pillsState.map((stateItem) => stateItem.item.id);
+
+    // The item ids and their order must match exactly with the internal state, if not, the state needs to be synced up
+    if (itemIds.join() !== stateIds.join()) {
+      const newState = items.map((item) => {
+        const oldItem = pillsState.filter((stateItem) => stateItem.item === item)[0];
```
chanceaclark, line 109: Can we use `.find` here instead?

chanceaclark (commented):
One comment, but overall looks great.

packages/big-design/src/components/PillTabs/PillTabs.tsx
rtalvarez: Ahh yeah.. that makes way more sense

chanceaclark (approved):

PR #685, opened 2022-01-20
feat(component): table row with expandable details

bc-rmalyavc:
## What?
Table row with expandable details

## Why?
To add a possibility to pass additional expandable content to specific table rows

image
image

## Testing/Proof
Manually tested on chrome 95.0.4638.69

### DIFF
--- a/packages/big-design/src/components/Table/Row/Row.tsx
+++ b/packages/big-design/src/components/Table/Row/Row.tsx
@@ -1,14 +1,16 @@
-import { DragIndicatorIcon } from '@bigcommerce/big-design-icons';
-import React, { forwardRef, TableHTMLAttributes } from 'react';
+import { ChevronRightIcon, DragIndicatorIcon, ExpandMoreIcon } from '@bigcommerce/big-design-icons';
+import React, { forwardRef, TableHTMLAttributes, useEffect, useState } from 'react';
 
 import { typedMemo } from '../../../utils';
+import { Button } from '../../Button';
 import { Checkbox } from '../../Checkbox';
 import { DataCell } from '../DataCell';
 import { TableColumn, TableItem } from '../types';
 
-import { StyledTableRow } from './styled';
+import { StyledDataCell, StyledTableRow } from './styled';
 
 export interface RowProps<T> extends TableHTMLAttributes<HTMLTableRowElement> {
+  isExpandable?: boolean;
   isDragging?: boolean;
   isSelected?: boolean;
   isSelectable?: boolean;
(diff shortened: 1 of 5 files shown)


solofeed (approved):

packages/big-design/src/components/Table/types.ts
```diff
@@ -21,6 +21,7 @@ export interface TableSortable<T> {
 export interface TableItem {
   id?: string | number;
   [key: string]: any;
+  expandableContent?: ReactNode;
```
chanceaclark, line 24: We shouldn't be adding additional properties to the `TableItem` prop. Instead we should think about making the table expandable from the `TableColumn` perspective. I don't see the need of individually adding it per item.

Think about something like this:
```tsx
const columns = [
  { header: 'ID', hash: 'id', expandedContent: ({ id }) => <ComponentThatFetchesById id={id} /> },
]
```

packages/big-design/src/components/Table/Row/Row.tsx
```diff
@@ -40,35 +43,68 @@ const InternalRow = <T extends TableItem>({
...
-        </DataCell>
+    <>
+      <StyledTableRow isSelected={isSelected} ref={forwardedRef} isDragging={isDragging} {...rest}>
+        {isExpandable && (
+          <DataCell withPadding={false}>
+            {itemHasChild && (
+              <Button
+                variant="subtle"
+                onClick={toggleExpanded}
+                iconOnly={isExpanded ? <ExpandMoreIcon color="gray" /> : <ChevronRightIcon color="gray" />}
```
chanceaclark, line 69: I see no accessibility features added to this addition. This [codepen](https://codepen.io/aardrian/pen/wvKWVBW) has a good example, which we'll have to pick and choose from.

packages/big-design/src/components/Table/Row/Row.tsx
```diff
@@ -40,35 +43,68 @@ const InternalRow = <T extends TableItem>({
   };
 
   const label = isSelected ? `Selected` : `Unselected`;
+  const [isExpanded, setIsExpanded] = useState(false);
```
chanceaclark, line 46: The `Table` component should be stateless by design so we should not be adding in new state to the component. See the `selectable` prop to how we achieve this.

🍹 Also, new changes to the `Table` component should be added to the `StatefulTable` component. This can happen in another PR.

chanceaclark (changes requested):
Here's my first pass, but two items need to be addressed before reviewed any further:
1. Add documentation
2. Add tests

We should be talking about the API design before we add any new code. We have bi-weekly syncs that you can attend for new additions like this.

eugene-polev (approved):

dhc86:
Hello team, any updates on this PR?

solofeed:
@dhc86 Unfornutelly we did not finish all the work, this PR requires more attention to architecture aspects of the big design standards. 
Most likely this feature will be implemented later by the big design team itself, but I don't have ETA for you.

PR #688, opened 2022-01-20
feat(docs): update popover page to use new layout

jorgemoya:
## What?

Update Popup page to use new layout.

screencapture-localhost-3000-popover-2022-01-20-14_23_19

### DIFF
--- a/packages/docs/pages/popover.tsx
+++ b/packages/docs/pages/popover.tsx
@@ -1,20 +1,27 @@
 import { Button, H1, Panel, Popover, Text } from '@bigcommerce/big-design';
 import React, { useState } from 'react';
 
-import { CodePreview, PageNavigation } from '../components';
+import { Code, CodePreview, GuidelinesTable, List } from '../components';
 import { BoxPropTable, PaddingPropTable, PopoverPropTable } from '../PropTables';
 
 const PopoverPage = () => {
-  const items = [
-    {
-      id: 'examples',
-      title: 'Examples',
-      render: () => (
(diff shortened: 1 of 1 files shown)


packages/docs/pages/popover.tsx
```diff
@@ -39,30 +46,27 @@ const PopoverPage = () => {
...
 
-  return (
-    <>
-      <H1>Popover</H1>
-
-      <PageNavigation items={items} />
+      <Panel header="Do's and Don'ts" headerId="guidelines">
+        <GuidelinesTable
+          recommended={['Tie popup to click or hover events on elements.', 'Use short texts.']}
+          discouraged={['Always display popup.', 'Extended texts that disrupt the screen.']}
```
chanceaclark, line 67: What does `'Use short texts.'` and `'Extended texts that disrupt the screen.'` mean? Very confused by this 🤔

chanceaclark (approved):

packages/docs/pages/popover.tsx
jorgemoya: I'm trying to represent that popups aren't supposed to be used for longer texts/paragraphs. Not sure how to word it correctly, or if it's relevant.

chanceaclark: Hmm... maybe like `Use concise textual phrases within.`? Not sure how to rephrase the last one though 🤔

jorgemoya: I'm just gonna remove it.

PR #711, opened 2022-02-08
feat(docs): update button page to use new layout

MariaJose:
## What?
Update button documentation page to use new layout. 

## Screenshots
Screen Shot 2022-02-08 at 2 10 40 PM


packages/docs/components/CodePreview/CodePreview.tsx
```diff
@@ -89,7 +89,3 @@ export const CodePreview: React.FC<CodePreviewProps> = (props) => {
     </BigDesign.Box>
   );
 };
-
-CodePreview.defaultProps = {
-  scope: defaultScope,
-};
```
MariaJose, line 95: Noticed that `defaultScope` was directly used in the following code (shown below), so I think `lines 92 to 95` might not be necessary? Unless I'm missing something 😅 :

```
const [scope, setScope] = useState({ ...defaultScope, ...props.scope });

useEffect(() => {
    setScope({ ...defaultScope, ...props.scope });
  }, [props.scope, setScope]);
```

packages/docs/pages/button.tsx
```diff
@@ -1,145 +1,221 @@
...
-      <PageNavigation items={items} />
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          <Code primary>Buttons</Code> trigger an immediate action, which can continue within the current page, a new
+          page or modal.
+        </Text>
+        <Text bold>When to use:</Text>
+        <List>
+          <List.Item>
+            Use <Code primary>Primary Buttons</Code> for the most important action on a page (e.g. creating an object,
```
chanceaclark, line 21: I like where your head is at here, but we probably don't want this all to be in a `<Code>` element. The general rule of thumb is:
- `<Code primary />` for elements/prop names.
- `<Code />` for prop values.
(we have been/are inconsistent)

I'd change this to the following since "buttons" not entirely referring to the component name. Same applies for the rest of the page.
```suggestion
            Use <Code>primary</Code> buttons for the most important action on a page (e.g. creating an object,
```

packages/docs/pages/button.tsx
```diff
@@ -1,145 +1,221 @@
...
+                    {/* jsx-to-string:end */}
+                  </CodePreview>
+                </>
+              ),
+            },
+            {
+              id: 'variants',
+              title: 'Variants',
+              render: () => (
+                <Fragment key="variants">
```
chanceaclark, line 63: You can actually just leave these as fragments without keys `<></>`

chanceaclark (commented):
Looking good, just a few comments...

packages/docs/pages/button.tsx
MariaJose: I tried to use the `<></>` but when using the fragments like that the `LiveEditor` wasn't working as expected. It was rendering the same `code` when switching between tabs.

jorgemoya: If that is the case we need to fix these in a bunch of pages, or maybe figure out why `LiveEditor` doesn't like that,

MariaJose: sounds good, I'll check which pages are having the same issue

chanceaclark: Ah yeah good catch. I now remember having to do this in the `Form` documentation.

PR #719, opened 2022-02-15
feat(docs): Update multi select page to use new layout

MariaJose:
## What?
Update MultiSelect documentation page to use the new layout

Screen Shot 2022-02-15 at 2 27 25 PM

### DIFF
--- a/packages/docs/pages/multi-select.tsx
+++ b/packages/docs/pages/multi-select.tsx
@@ -461,10 +461,6 @@ const MultiSelectPage = () => {
             },
           ]}
         />
-
-        {/* <MultiSelectPropTable />
-        <SelectOptionPropTable />
-        <SelectActionPropTable /> */}
       </Panel>
 
       <Panel header="Do's and Don'ts" headerId="guidelines">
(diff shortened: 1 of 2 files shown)


packages/docs/pages/multi-select.tsx
```diff
@@ -1,397 +1,483 @@
...
+              ),
+            },
+            {
+              id: 'position',
+              title: 'Position',
+              render: () => (
+                <Fragment key="position">
+                  <Text>
+                    <Code primary>MultiSelect</Code> can be anchored in different directions with the{' '}
+                    <Code primary>placement</Code> property. It will automatically find a position if there's not enough
```
MariaJose, line 95: I updated ` <Code primary>position</Code>` to `  <Code primary>placement</Code>` since I think this implementation is talking about different ways of using the placement property. Example:
` placement="top"`
`placement="bottom-start"`
...and so on.

packages/docs/pages/multi-select.tsx
```diff
@@ -1,397 +1,483 @@
...
+                </Fragment>
+              ),
+            },
+            {
+              id: 'error',
+              title: 'Error',
+              render: () => (
+                <Fragment key="error">
+                  <Text>
+                    An <Code primary>error</Code> prop receives a <Code>string</Code> to display.
```
MariaJose, line 314: I updated the ```<Code>error</Code>``` to ```<Code primary>error</Code>``` since this is referring to the error prop.

chanceaclark (approved):

Issue #661, opened 2021-12-10
'wrap-reversed' type in flex-wrap is misspelled, should be 'wrap-reverse'


clopezh:
**Describe the bug**
flex-wrap property has a type https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Flex/types.ts#L30 **wrap-reversed** but the correct value is **wrap-reverse**

**To Reproduce**
1. Try to use flexWrap in a Flex component
2. See the unexpected behavior in browser and if you check the inspector you will see the warning:
![image](https://user-images.githubusercontent.com/35146660/145614268-089ec7cb-b37d-4335-80a8-a6d300e9c839.png)

**Expected behavior**
**wrap-reverse** is the correct property according to flex rules.

chanceaclark:
Good catch @clopezh! Definite something to fix!

bc-prestonhuth:
Was this issue ever addressed?

PR #721, opened 2022-02-16
fix(component): change flex-reversed to flex-reverse

bc-juanvasquez:
## What?

Fixes [https://github.com/bigcommerce/big-design/issues/661](url)

## Why?

Flex component has type `wrap-reversed` but the correct value is `wrap-reverse`

Screen Shot 2022-02-16 at 4 11 46 PM

## Testing/Proof

Inspected Flex component example and verified correct property is being applied.


jorgemoya (approved):
lgtm! 👍

chanceaclark (approved):
Looks great!

golcinho (approved):

eugene-polev (approved):

PR #722, opened 2022-02-17
feat(docs): update Pill tabs to use new layout

MariaJose:
## What?
Update pill tabs to use new layout.

## Screenshots
Screen Shot 2022-02-17 at 4 18 28 PM

### DIFF
--- a/packages/docs/PropTables/PillTabsPropTable.tsx
+++ b/packages/docs/PropTables/PillTabsPropTable.tsx
@@ -11,7 +11,10 @@ const pillTabsPropTable: Prop[] = [
   {
     description: (
       <>
-        See <NextLink href={{ hash: 'pill-tabs-items-prop-table', query: { props: 'pill-tab-item' } }}>here</NextLink>{' '}
+        See{' '}
+        <NextLink href={{ hash: 'pill-tabs-items-prop-table', query: { props: 'pill-tab-item' } }}>
+          PillTabItem
+        </NextLink>{' '}
         for usage.
       </>
     ),
--- a/packages/docs/PropTables/PillTabsPropTable.tsx
+++ b/packages/docs/PropTables/PillTabsPropTable.tsx
@@ -11,12 +11,12 @@ const pillTabsPropTable: Prop[] = [
   {
     description: (
       <>
-        See <NextLink href="#pill-tabs-items-prop-table">below</NextLink> for usage.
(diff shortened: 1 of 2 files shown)


packages/docs/PropTables/PillTabsPropTable.tsx
```diff
@@ -11,12 +11,12 @@ const pillTabsPropTable: Prop[] = [
   {
     description: (
       <>
-        See <NextLink href="#pill-tabs-items-prop-table">below</NextLink> for usage.
+        See <NextLink href="/pill-tabs?props=pill-tab-item#props">here</NextLink> for usage.
```
chanceaclark, line 14: We did something a little differently for the other pages. See the [TablePropTable](https://github.com/bigcommerce/big-design/blob/master/packages/docs/PropTables/TablePropTable.tsx#L11) as an example.

chanceaclark (commented):
Looks great, beside one little comment!

packages/docs/PropTables/PillTabsPropTable.tsx
MariaJose: sounds good, I'll take a look. Ty!

packages/docs/pages/pill-tabs.tsx
```diff
@@ -1,110 +1,134 @@
...
+        />
+      </Panel>
+
+      <Panel header="Do's and Don'ts" headerId="guidelines">
+        <GuidelinesTable
+          recommended={[
+            <>Use on pages that have a large amount of content.</>,
+            <>Be concise on the navigation labels, ideally one or two words rather than a phrase.</>,
+            <>
+              Default page view should have no <Code primary>PillTabs</Code> selected.
```
MariaJose, line 120: Not sure if we should update this sentence (line 120) like this:
`When filtering a collection of items, the default page view should have no <Code primary>PillTabs</Code> selected`

MariaJose:
♻️

jorgemoya (approved):

packages/docs/pages/pill-tabs.tsx
jorgemoya: To me this change reads as if we have already clicked a pill tab (thus filtered).

MariaJose: makes sense, Initially I thought that pill tabs were used as tabs, and not only filters. Maybe we could leave it just as it is.

chanceaclark (approved):

PR #735, opened 2022-02-22
feat(docs): update radio page to use new layout

MariaJose:
## What?
Update radio page to use new layout.

## Why?

## Screenshots

Screen Shot 2022-02-22 at 1 57 40 PM

### DIFF
--- a/packages/docs/PropTables/shared.tsx
+++ b/packages/docs/PropTables/shared.tsx
@@ -67,7 +67,7 @@ export const messagingLinkItemProps: Prop[] = [
     types: 'boolean',
     description: (
       <>
-        Shows an external icon when the <Code primary>external</Code> flag is set and target="_blank".
+        Shows an external icon when the <Code primary>external</Code> prop is set and target="_blank".
       </>
     ),
   },
--- a/packages/docs/PropTables/shared.tsx
+++ b/packages/docs/PropTables/shared.tsx
@@ -67,7 +67,7 @@ export const messagingLinkItemProps: Prop[] = [
     types: 'boolean',
     description: (
       <>
-        Shows an external icons when the <Code primary>external</Code> flag is set and target="_blank".
+        Shows an external icon when the <Code primary>external</Code> flag is set and target="_blank".
       </>
     ),
   },
--- a/packages/docs/pages/radio.tsx
+++ b/packages/docs/pages/radio.tsx
@@ -10,7 +10,7 @@ const RadioPage = () => {
       <H1>Radio</H1>
 
       <Panel header="Overview" headerId="overview">
-        <Text>Radio buttons let users select an option from a list of two or more.</Text>
+        <Text>Radio buttons let users select an option from a list of two or more items.</Text>
         <Text bold>When to use:</Text>
         <List>
           <List.Item>
@@ -29,7 +29,7 @@ const RadioPage = () => {
               render: () => (
                 <Fragment key="basic">
                   <Text>
-                    <Code primary>Radios</Code> are group of items which a single option can be selected.
+                    A <Code primary>Radio</Code> is a group of items from which a single option can be selected.
                   </Text>
                   <CodePreview>
                     {/* jsx-to-string:start */}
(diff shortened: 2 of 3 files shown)


packages/docs/PropTables/shared.tsx
```diff
@@ -67,7 +67,7 @@ export const messagingLinkItemProps: Prop[] = [
     types: 'boolean',
     description: (
       <>
-        Shows an external icons when the <Code primary>external</Code> flag is set and target="_blank".
+        Shows an external icon when the <Code primary>external</Code> flag is set and target="_blank".
```
MariaJose, line 70: Should we update `external flag` to `external prop`?

packages/docs/pages/radio.tsx
```diff
@@ -1,151 +1,219 @@
...
+                </Fragment>
+              ),
+            },
+            {
+              id: 'grouping',
+              title: 'Grouping',
+              render: () => (
+                <Fragment key="grouping">
+                  <Text>
+                    In order to group radio controls, use the <Code>Fieldset</Code> component to separate the controls.
```
MariaJose, line 68: Tried to find the `Fieldset` page, but I didn't find it, Is that expected?
Should  we remove the `<Code/>` tag?

packages/docs/PropTables/shared.tsx
jorgemoya: Sounds like a good idea to me

packages/docs/pages/radio.tsx
jorgemoya: I guess we're referencing the html element. I would remove the `Code` tag.

chanceaclark: We have a component for [`Fieldset`](https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Fieldset/Fieldset.tsx) but we don't have a dedicated page for it. It's located on the `Forms` page props.

jorgemoya: We group `Fieldset` in the `Form` component. I think it's fine to leave as is.

chanceaclark (approved):
Besides the `flag` -> `prop` change, lgtm

jorgemoya (approved):

packages/docs/pages/radio.tsx
```diff
@@ -1,151 +1,219 @@
...
-              {/* jsx-to-string:start */}
-              {function Example() {
-                const [selected, setSelected] = useState('');
+  return (
+    <>
+      <H1>Radio</H1>
 
-                const handleChange = (event) => setSelected(event.target.value);
+      <Panel header="Overview" headerId="overview">
+        <Text>Radio buttons let users select an option from a list of two or more.</Text>
```
bc-andreadao, line 13:
```suggestion
        <Text>Radio buttons let users select an option from a list of two or more items.</Text>
```

packages/docs/pages/radio.tsx
```diff
@@ -1,151 +1,219 @@
...
+        <ContentRoutingTabs
+          id="implementation"
+          routes={[
+            {
+              id: 'basic',
+              title: 'Basic',
+              render: () => (
+                <Fragment key="basic">
+                  <Text>
+                    <Code primary>Radios</Code> are group of items which a single option can be selected.
```
bc-andreadao, line 32:
```suggestion
                    A <Code primary>Radio</Code> is a group of items from which a single option can be selected.
```

packages/docs/pages/radio.tsx
```diff
@@ -1,151 +1,219 @@
...
+          ]}
+        />
+      </Panel>
 
-      <PageNavigation items={items} />
+      <Panel header="Do's and Don'ts" headerId="guidelines">
+        <GuidelinesTable
+          recommended={[
+            <>Group related radio buttons under input headings.</>,
+            <>Radio buttons should always have a default selected option.</>,
```
bc-andreadao, line 206:
```suggestion
            <>Include a default selected option with the radio buttons.</>,
```

packages/docs/pages/radio.tsx
```diff
@@ -1,151 +1,219 @@
...
+            <>Radio buttons should always be laid out vertically.</>,
```
bc-andreadao, line 207:
```suggestion
            <>Lay radio buttons vertically.</>,
```

packages/docs/pages/radio.tsx
```diff
@@ -1,151 +1,219 @@
...
+          ]}
+          discouraged={[
+            <>Don’t use radio buttons for long lists of short items - use a select input instead.</>,
```
bc-andreadao, line 210:
```suggestion
            <>Don’t use radio buttons for long lists of short items. Use a select input instead.</>,
```

packages/docs/pages/radio.tsx
```diff
@@ -1,151 +1,219 @@
...
+            <>
+              A set of radio buttons should not have a state of being “unselected” - there must always be a selection.
```
bc-andreadao, line 212:
```suggestion
              A set of radio buttons should not have a state of being “unselected.” There must always be a selection.
```

MariaJose:
♻️

jorgemoya (approved):

bc-andreadao (approved):

Issue #676, opened 2022-01-17
ButtonGroup hovers the first destructive button even if disabled


samufort:
**Describe the bug**
Hi,

I have a button group with two destructive buttons hidden under the ellipsis. In some use cases, these buttons are both disabled. When this happens and the user opens the ellipsis, the first button hovers even if it's disabled.

**To Reproduce**
1. Create a button group with two disabled destructive buttons
2. Click on the ellipsis
3. See error

**Expected behavior**
No hover on the first item

**Screenshots**
![2022-01-17 10 53 20](https://user-images.githubusercontent.com/16794649/149802379-0bda7e19-61b6-4cf6-90e4-f2146a0b7f3b.gif)

**Additional context**

chanceaclark:
Hey @samufort, thanks for opening up this issue. This definitely seems like bug so I'll mark it as such.

PR #737, opened 2022-02-23
feat(docs): fix button group issue (676)

MariaJose:
## What?
**Issue1**: When destructive action buttons are disabled (hidden under the ellipsis), the first destructive action button is highlighted, and it should not be highlighted. For more context: (https://github.com/bigcommerce/big-design/issues/676)
**Issue2**: All buttons overlay when rendering just one action button + the ellipsis on the screen..
**Issue3**: The ellipsis appears when all the action buttons fit inside a container.

**Images:**
Issue1:
Screen Shot 2022-02-23 at 5 31 52 PM

Issue2:
Screen Shot 2022-02-23 at 5 30 33 PM

Issue3: 
Screen Shot 2022-03-08 at 12 27 25 PM

## Why?
To fix bugs.

## Screenshots
**Issue 1 and 2:** 

https://user-images.githubusercontent.com/39739180/155427754-7aaba7c5-766e-42f6-bdff-ddc768f0db41.mp4

**Issue 3:**
Screen Shot 2022-03-08 at 2 11 22 PM

## Testing/Proof
- Run ButtonGroup tests.
- Run Button tests

- Use a `ButtonGroup` component, and add some destructive action buttons. Make sure the `highlight` (color red) doesn't appear when using the `disabled` prop.

```jsx
 <ButtonGroup
      actions={[
        { actionType: 'destructive', text: 'Something destructive 1', disabled: true },
        { actionType: 'destructive', text: 'Something destructive 2', disabled: true },
        { actionType: 'destructive', text: 'Something destructive 3', disabled: true },
        { text: 'View in BigCommerce' },
      ]}
    />
```

### DIFF
--- a/packages/big-design/src/components/Button/spec.tsx
+++ b/packages/big-design/src/components/Button/spec.tsx
@@ -187,3 +187,25 @@ test('render only icon only with left and right icons button', () => {
 
   expect(screen.getAllByTestId('icon-only')).toHaveLength(1);
 });
+
+test('hides content when isLoading is true', () => {
+  const plusIcon = <AddIcon data-testid="icon-only" role="icon" aria-hidden="false" />;
+  const { rerender } = render(
+    <Button isLoading={false} iconLeft={plusIcon}>
+      Button
+    </Button>,
+  );
+  const visibleContent = screen.getByRole('icon');
+
+  expect(visibleContent).toBeInTheDocument();
+
+  rerender(
+    <Button isLoading={true}>
+      <strong>Button</strong>
+    </Button>,
+  );
+
+  const hiddenContent = screen.queryByRole('icon');
+
+  expect(hiddenContent).not.toBeInTheDocument();
+});
--- a/packages/big-design/src/components/Button/spec.tsx
+++ b/packages/big-design/src/components/Button/spec.tsx
@@ -189,7 +189,7 @@ test('render only icon only with left and right icons button', () => {
 });
 
 test('hides content when isLoading is true', () => {
-  const plusIcon = <AddIcon data-testid="icon-only" role="icon" aria-hidden="false" />;
+  const plusIcon = <AddIcon role="icon" aria-hidden="false" />;
   const { rerender } = render(
     <Button isLoading={false} iconLeft={plusIcon}>
       Button
--- a/packages/big-design/src/components/Button/styled.tsx
+++ b/packages/big-design/src/components/Button/styled.tsx
@@ -88,6 +88,7 @@ export const ContentWrapper = styled.span.attrs<Record<string, unknown>, { isLoa
   display: inline-grid;
   grid-auto-flow: column;
   grid-gap: ${({ theme }) => theme.spacing.xSmall};
+
   ${({ isLoading }) =>
     isLoading &&
     css`
--- a/packages/big-design/src/components/Button/styled.tsx
+++ b/packages/big-design/src/components/Button/styled.tsx
@@ -81,6 +81,7 @@ export const StyledButton = styled.button<ButtonProps & MarginProps>`
 
   ${(props) => getButtonStyles(props)}
 `;
+// visibility: ${({ isLoading }) => (isLoading ? 'hidden' : 'visible')};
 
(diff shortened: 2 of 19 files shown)


packages/big-design/src/components/ButtonGroup/styled.tsx
```diff
@@ -6,6 +6,7 @@ import { FlexItem } from '../Flex';
 
 interface StyledButtonProps {
   borderRadius?: boolean;
+  isVisible?: boolean;
```
chanceaclark, line 9: So I don't think we need to add an `isVisible` prop here 🤔  Since there is an absolute positioning happening [here](https://github.com/bigcommerce/big-design/blob/7ce0bbee8b686bd9b6c49333a3a19a084c6ca206/packages/big-design/src/components/ButtonGroup/styled.tsx#L53), all we should need to do is add a huge `right` value:

Screen Shot 2022-03-08 at 18 31 56

From what I gathered, we need the button to still be in the dom, but hidden away from everything.

MariaJose: Why do we need the ellipsis button to still be in the Dom?

chanceaclark: It doesn't just apply to the ellipsis, but the other buttons as well. They need to be in the DOM, but hidden from the user. If you use `display:none` the `ref` is set to `undefined` thus we can't do anything with it.

packages/big-design/src/components/Button/styled.tsx
```diff
@@ -88,7 +88,11 @@ export const ContentWrapper = styled.span.attrs<Record<string, unknown>, { isLoa
   display: inline-grid;
   grid-auto-flow: column;
   grid-gap: ${({ theme }) => theme.spacing.xSmall};
-  visibility: ${({ isLoading }) => (isLoading ? 'hidden' : 'visible')};
+  ${({ isLoading }) =>
+    isLoading &&
+    css`
+      visibility: hidden;
+    `};
```
MariaJose, line 95: Need to update this logic since this was causing that the style (visibility: hidden) was not applied for the buttons since they have by default `visibility: visible`.
s

packages/big-design/src/components/Button/spec.tsx
```diff
@@ -187,3 +187,25 @@ test('render only icon only with left and right icons button', () => {
 
   expect(screen.getAllByTestId('icon-only')).toHaveLength(1);
 });
+
+test('hides content when isLoading is true', () => {
+  const plusIcon = <AddIcon data-testid="icon-only" role="icon" aria-hidden="false" />;
```
MariaJose, line 192: Had to add the `role=icon` so I could grab it from the DOM, and be able to test if the content was visible or not since using `getByText('Button')` the content appeared in the DOM 🤔 even though `isLoading` was true.

packages/big-design/src/components/List/List.tsx
```diff
@@ -128,7 +128,7 @@ const StyleableList = typedMemo(
                 index={key}
                 isAction={isDropdown}
                 isChecked={isChecked || false}
-                isHighlighted={highlightedIndex === key}
+                isHighlighted={!item.disabled && highlightedIndex === key}
```
jorgemoya, line 131: 👍 👍

packages/big-design/src/components/Button/spec.tsx
jorgemoya: If you're using `role` do we need `data-testid`? Otherwise can you use `getByTestId('icon-only')` instead of `role`?

Hmm, maybe the `Button` label is only being visually hidden but remains available for DOM/accessibility?

MariaJose: ohh yes! Good catch, it shoulnd't have the data-testid 😅 . I copied and pasted `<AddIcon data-testid="icon-only" role="icon" aria-hidden="false" />;` from a previous test, but I forgot to delete the things that I didn't need for this test. 

Yes, I think the selectors that works for accessibility are the ones that uses `*ByRole`

chanceaclark (approved):

PR #738, opened 2022-02-24
feat(docs): update select page to use new layout

MariaJose:
## What?
Update select page to use new layout

## Why?

Screen Shot 2022-02-24 at 3 26 31 PM

### DIFF
--- a/packages/docs/pages/select.tsx
+++ b/packages/docs/pages/select.tsx
@@ -1,399 +1,499 @@
 import { Form, FormGroup, Grid, H1, Panel, Select, Text } from '@bigcommerce/big-design';
 import { DeleteIcon } from '@bigcommerce/big-design-icons';
-import React, { useState } from 'react';
+import React, { Fragment, useState } from 'react';
 
-import { Code, CodePreview, PageNavigation } from '../components';
+import { Code, CodePreview, ContentRoutingTabs, GuidelinesTable, List, NextLink } from '../components';
 import { SelectActionPropTable, SelectGroupPropTable, SelectOptionPropTable, SelectPropTable } from '../PropTables';
 
 const SelectPage = () => {
-  const items = [
-    {
-      id: 'examples',
-      title: 'Examples',
-      render: () => (
-        <>
-          <Panel>
-            <Text>Select are typeable inputs with selectable dropdown items.</Text>
-            <CodePreview>
-              {/* jsx-to-string:start */}
-              {function Example() {
-                const [value, setValue] = useState('mx');
-                const handleChange = (val) => setValue(val);
-
-                return (
-                  <Form>
-                    <FormGroup>
-                      <Select
-                        action={{
-                          actionType: 'destructive' as const,
-                          content: 'Remove Country',
-                          icon: <DeleteIcon />,
-                          onActionClick: () => null,
-                        }}
-                        filterable={true}
-                        label="Countries"
-                        maxHeight={300}
-                        onOptionChange={handleChange}
-                        options={[
-                          { value: 'us', content: 'United States' },
-                          { value: 'mx', content: 'Mexico' },
-                          { value: 'ca', content: 'Canada' },
-                          { value: 'en', content: 'England' },
-                          { value: 'fr', content: 'France' },
-                          { value: 'gr', content: 'Germany' },
-                          { value: 'ar', content: 'Argentina' },
-                          { value: 'ru', content: 'Russia', disabled: true },
-                          { value: 'ch', content: 'Chile' },
-                          { value: 'bo', content: 'Bolivia' },
-                          { value: 'jp', content: 'Japan' },
-                          { value: 'cn', content: 'China' },
-                          { value: 'sk', content: 'South Korea' },
-                          { value: 'au', content: 'Australia' },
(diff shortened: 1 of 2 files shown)


packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
+        <ContentRoutingTabs
+          id="implementation"
+          routes={[
+            {
+              id: 'basic',
+              title: 'Basic',
+              render: () => (
+                <Fragment key="basic">
+                  <Text>
+                    <Code primary>Select</Code> are typeable inputs with selectable options.
```
MariaJose, line 35: I updated `selectable dropdown items` to `selectable options` is that ok? suggestions?

packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
+                </Fragment>
+              ),
+            },
+            {
+              id: 'max-height',
+              title: 'Max height',
+              render: () => (
+                <Fragment key="max-height">
+                  <Text>
+                    Once the content is longer than the max-height, the <Code primary>Select</Code> will be scrollable.
```
MariaJose, line 166: I updated `the Dropdown will be scrollable` to `the Select will be scrollable`? Is that ok? other suggestions?

packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
-            </CodePreview>
-          </Panel>
-          <Panel header="Disabled">
-            <Text>
-              It is possible to disable the entire select component, similar to how you would disable a native HTML
-              select element, by using the <Code primary>disabled</Code> property.
-            </Text>
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          <Code primary>Select</Code> give merchants the ability to make a single selection or multiple selections from
```
rtalvarez, line 15: Are we missing an `s` in "give"? So it kinda reads "Select gives merchants the ability ... "

MariaJose: ohh yes, good catch!

packages/docs/pages/select.tsx
jorgemoya: Good to me.

packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
+        <ContentRoutingTabs
+          id="implementation"
+          routes={[
+            {
+              id: 'basic',
+              title: 'Basic',
+              render: () => (
+                <Fragment key="basic">
+                  <Text>
+                    <Code primary>Select</Code> are typeable inputs with selectable options.
```
jorgemoya, line 35:
```suggestion
                    <Code primary>Selects</Code> are typeable inputs with selectable options.
```

packages/docs/pages/select.tsx
jorgemoya: Good to me.

jorgemoya (commented):
Looking good!

packages/docs/pages/select.tsx
MariaJose: Good catch, ty!

rtalvarez (approved):

packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
+          ]}
+        />
+      </Panel>
 
-      <PageNavigation items={items} />
+      <Panel header="Do's and Don'ts" headerId="guidelines">
+        <GuidelinesTable
+          recommended={[
+            <>
+              Typically used within <NextLink href="/panel">Panels</NextLink>.
```
bc-andreadao, line 476:
```suggestion
              Typically use within <NextLink href="/panel">Panels</NextLink>.
```

packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
+            </>,
+            'All select options should be within a container.',
```
bc-andreadao, line 478:
```suggestion
            'Create all select options within a container.',
```

packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
+            <>
+              Be used for a single selection between 4 or more pre-defined <Code primary>options</Code>.
```
bc-andreadao, line 480:
```suggestion
              Use for a single selection between 4 or more pre-defined <Code primary>options</Code>.
```

packages/docs/pages/select.tsx
```diff
@@ -1,399 +1,499 @@
...
+            </>,
+            'Default a selection if possible.',
+            'Placeholder text can be added if additional context is needed.',
```
bc-andreadao, line 483:
```suggestion
            'Add placeholder text if additional context is needed.',
```

packages/docs/pages/select.tsx
chanceaclark: The suggestion is now phrase _wierd_. Could we potentially just drop `Typically`?

MariaJose: FYI, I dropped the `Typically` 👍🏼

chanceaclark (approved):

PR #740, opened 2022-02-28
feat(docs): update input page to use new layout

MariaJose:
## What?
Update input page to use new layout.

## Why?
Provide better experience to the user when reading or checking the docs.

## Screenshots
Screen Shot 2022-02-28 at 1 08 46 PM

### DIFF
--- a/packages/docs/pages/input.tsx
+++ b/packages/docs/pages/input.tsx
@@ -1,98 +1,147 @@
 import { Form, FormGroup, H1, Input, Panel, Text } from '@bigcommerce/big-design';
 import { CheckCircleIcon } from '@bigcommerce/big-design-icons';
-import React, { useState } from 'react';
+import React, { Fragment, useState } from 'react';
 
-import { Code, CodePreview, PageNavigation } from '../components';
+import { Code, CodePreview, ContentRoutingTabs, GuidelinesTable, List, NextLink } from '../components';
 import { InputPropTable } from '../PropTables';
 
 const InputPage = () => {
-  const items = [
-    {
-      id: 'examples',
-      title: 'Examples',
-      render: () => (
-        <>
-          <Panel>
-            <Text>Inputs are stylized form controls with the ability of controling validation.</Text>
-            <CodePreview>
-              {/* jsx-to-string:start */}
-              {function Example() {
-                const [value, setValue] = useState('');
-
-                const handleChange = (event) => setValue(event.target.value);
-
-                return (
-                  <Form>
-                    <FormGroup>
-                      <Input
-                        label="Label"
-                        description="Description for the input."
-                        placeholder="Placeholder"
-                        type="text"
-                        value={value}
-                        onChange={handleChange}
-                      />
-                    </FormGroup>
-                  </Form>
-                );
-              }}
-              {/* jsx-to-string:end */}
-            </CodePreview>
-          </Panel>
-          <Panel header="Error states">
-            <Text>
-              Inputs allow you to pass in an <Code primary>error</Code> message that will control the styles of an
-              input. The logic on the input can be controlled with the <Code primary>onChange</Code> prop.
-            </Text>
-
-            <CodePreview>
-              {/* jsx-to-string:start */}
-              <Form>
-                <FormGroup>
-                  <Input
-                    label="Email Address"
-                    description="Provide a valid email address."
-                    value="example@"
-                    error="Email address must contain a domain name."
-                    onChange={() => null}
-                  />
-                </FormGroup>
-              </Form>
-              {/* jsx-to-string:end */}
-            </CodePreview>
-          </Panel>
-
-          <Panel header="Icons">
-            <Text>
-              Inputs can also contain icons via the <Code>iconLeft</Code> &amp; <Code>iconRight</Code> props.
-            </Text>
-
-            <CodePreview>
-              {/* jsx-to-string:start */}
-              <Form>
-                <FormGroup>
-                  <Input label="Example" placeholder="Example" iconLeft={<CheckCircleIcon color="success" />} />
-                  <Input label="Example" placeholder="Example" iconRight={<CheckCircleIcon color="success" />} />
-                </FormGroup>
-              </Form>
-              {/* jsx-to-string:end */}
-            </CodePreview>
-          </Panel>
-        </>
-      ),
-    },
-    {
-      id: 'props',
-      title: 'Props',
-      render: () => <InputPropTable />,
-    },
-  ];
-
   return (
     <>
       <H1>Input</H1>
 
-      <PageNavigation items={items} />
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          <Code primary>Inputs</Code> are form controls that allow the entry of numeric or text data, often supported by
+          system validation.
+        </Text>
+        <Text bold>When to use:</Text>
+        <List>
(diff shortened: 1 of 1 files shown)


packages/docs/pages/input.tsx
```diff
@@ -1,98 +1,147 @@
...
+
+      <Panel header="Props" headerId="props">
+        <InputPropTable renderPanel={false} />
+      </Panel>
 
-      <PageNavigation items={items} />
+      <Panel header="Do's and Don'ts" headerId="guidelines">
+        <GuidelinesTable
+          recommended={[
+            'Provide placeholder text, examples or formats of relevent content. Placeholder text disappears after user input.',
```
MariaJose, line 128: Updated `Provider` to `Provide`

packages/docs/pages/input.tsx
```diff
@@ -1,98 +1,147 @@
...
-              {function Example() {
-                const [value, setValue] = useState('');
+  return (
+    <>
+      <H1>Input</H1>
 
-                const handleChange = (event) => setValue(event.target.value);
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          <Code primary>Inputs</Code> are form controls that allow the entry of numeric or text data, often supported by
```
rtalvarez, line 15: Hmmm, trying to find a better phrasing here .. maybe "... that allow entering numeric or text data,... " ??

🤔 maybe others will have better suggestions, otherwise lgtm

chanceaclark: I'm good with either or 🤷‍♂️

packages/docs/pages/input.tsx
```diff
@@ -1,98 +1,147 @@
...
+          system validation.
+        </Text>
+        <Text bold>When to use:</Text>
+        <List>
+          <List.Item>Use text inputs to let the user input short lines of text or numbers.</List.Item>
+          <List.Item>
+            Typically used within <NextLink href="/form">Forms</NextLink>.
```
chanceaclark, line 22: Umm... I would argue that they should _always_ be used in forms.

packages/docs/pages/input.tsx
chanceaclark, line 128: >Placeholder text disappears after user input.

This doesn't seem necessary since it's an inherit behavior of HTML. What do y'all think?

packages/docs/pages/input.tsx
```diff
@@ -1,98 +1,147 @@
...
+      <Panel header="Props" headerId="props">
+        <InputPropTable renderPanel={false} />
+      </Panel>
 
-      <PageNavigation items={items} />
+      <Panel header="Do's and Don'ts" headerId="guidelines">
+        <GuidelinesTable
+          recommended={[
+            'Provide placeholder text, examples or formats of relevent content. Placeholder text disappears after user input.',
+            'Provide inline error notification post input submission.',
```
chanceaclark, line 129: 🍹 Hmm... we've had a few conversations about this. There's two factors here:
1. We probably should use validation on blur/change, to allow users to fix their issues before they submit.
2. We should also validate post submit to show any server-side validation errors.

I'm not 100% what to do about this right now, but we might want to chat with design about this cc. @bigcommerce/product-design

packages/docs/pages/input.tsx
rtalvarez: I think its valid HTML5 to put them outside ?

packages/docs/pages/input.tsx
rtalvarez: We have been back and forth on 1 for years afaik, a ton of the Amazon angular stuff was validated on blur, and then big-design came around and afaik we collectively decided to validate on submit

packages/docs/pages/input.tsx
MariaJose: I'm also good with `allow the entry`, but I think I like better `that allow entering numeric or text data` @bigcommerce/dev-docs-team any suggestions? Thanks!

packages/docs/pages/input.tsx
chanceaclark: It's valid, but when would not put one inside a form?

MariaJose: Tried to fiend a official resource that explicity says that is ok to use it without a form, but I didn't. However, I saw some examples of using the <input/> outside a form. 
Some references:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search
https://html.spec.whatwg.org/multipage/input.html

I think would be more semantic correct if we use the input inside a form.

packages/docs/pages/input.tsx
MariaJose: I agree with it

packages/docs/pages/input.tsx
rtalvarez: > It's valid, but when would not put one inside a form?

The top channel-select for cp apps that are multi channel comes to mind. Essentially, I think design is exploring a page level header that provides the ability to quickly swap between channels. Its kind of a top level control that essentially re-renders the entire app with different data, and arguably is not technically a form

chanceaclark: I'd say that should be inside a form element still. It tells something to go fetch new data or sorts.

Also, encapsulating form controls inside forms provides encapsulation, especially if there are two sets of form controls on a page (channel swapping vs. the form and data it controls/renders)

rtalvarez: > I'd say that should be inside a form element still. It tells something to go fetch new data of sorts.

Hmm, lots of other things fetch data and are not inside forms. I'm not sure that just this by itself justifies the need for a form

> Also, wrapping form controls inside forms provides encapsulation, especially if there are two sets of form controls on a page (channel swapping vs. the form and data it controls/renders)

As long as all the form controls on the page are correctly ID-ed and labeled, and a11y accounted for, I would assume that's sufficient encapsulation

To be clear, I'm all in for forms. Kinda just playing devil's advocate here because the term _should always be used inside forms_ is quite harsh IMO

chanceaclark: >To be clear, I'm all in for forms. Kinda just playing devil's advocate here because the term should always be used inside forms is quite harsh IMO

Totally 😂 Probably shouldn't have used _always_, but "Typically" seems too light-hearted. Maybe it should be _most of the time_ 🤷‍♂️ 

There are definitely a few unique cases which probably you don't need a form, but in most cases you should use a form - you get a lot more benefit using them then not.

MariaJose: or could it be `Frequently use...` ?

rtalvarez: Yeah I would probably go with `Most of the time...`

packages/docs/pages/input.tsx
MariaJose: Note: I updated `line 15` with the following suggestion: `that allow entering numeric or text data`

packages/docs/pages/input.tsx
MariaJose: Note: I updated `line 22` with the following suggestion: `Most of the time...`

packages/docs/pages/input.tsx
MariaJose: Note: I deleted `Placeholder text disappears after user input.` sentence in `line 128`

rtalvarez (approved):
code lgtm

chanceaclark (approved):

jorgemoya (approved):

PR #746, opened 2022-03-02
feat(docs): fix code editor bug part 1

MariaJose:
## What?
Code editor wasn't rendering the corresponding code when switching between tabs inside the implementation section of a documentation page.

**Review:**
We are using a library called `react-live`, and the `<LiveEditor/>` (from react-live) component is the one that is causing the issue. Before this implementation, I tried two things:
- Use a reference (useRef) where I could assign the latest code and pass it directly to the `<LiveEditor/>` via the `code` prop or via the `onChange` prop, but the issue was still happening.
- Looked at the library, and found a fix for that. Basically the `<LiveEditor/>` wasn't updating the `code` that it was obtaining on every render since the `useEffect` was not used. (Reference: https://github.com/FormidableLabs/react-live/pull/271/files).

This fix was added in a new `react-live` version (not released yet, but you can use a tag to use the 3.0.0 version), however this version has some breaking changes that affects our library.

 Screen Shot 2022-02-28 at 5 33 26 PM

**Actual implantation:** Assign a `key` to all fragments used by  `react-live` so react can identify what to render. If someone else has another suggestion in mind, please let me know. Thanks!

Note: I'm gonna open additional PR's to make it easier to review.

## Why?
To display the corresponding code when users switch between the tabs. 

https://user-images.githubusercontent.com/39739180/156409195-fc4b8fcd-c1b1-4834-8a14-c2f8700c0f14.mp4


rtalvarez (approved):
Nice find, the `key` approach seems reasonable to me

jorgemoya (approved):

chanceaclark (approved):

PR #755, opened 2022-03-10
feat(docs): update margin and padding pages to use new layout

MariaJose:
## What?
Update Margin and Padding pages to use new layout.

## Why?

Screen Shot 2022-03-10 at 3 54 09 PM

Screen Shot 2022-03-10 at 3 54 41 PM

### DIFF
--- a/packages/docs/PropTables/MarginPropTable.tsx
+++ b/packages/docs/PropTables/MarginPropTable.tsx
@@ -5,37 +5,37 @@ import { NextLink, Prop, PropTable, PropTableWrapper } from '../components';
 const marginProps: Prop[] = [
   {
     name: 'margin',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
     description: 'Determines the margin to be applied.',
   },
   {
     name: 'marginTop',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
     description: 'Determines the top margin to be applied.',
   },
   {
     name: 'marginRight',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
     description: 'Determines the right margin to be applied.',
   },
   {
     name: 'marginBottom',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
     description: 'Determines the bottom margin to be applied.',
   },
   {
     name: 'marginLeft',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
     description: 'Determines the left margin to be applied.',
   },
   {
     name: 'marginVertical',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
     description: 'Determines the top and bottom margin to be applied.',
   },
   {
     name: 'marginHorizontal',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
     description: 'Determines the left and right margin to be applied.',
   },
 ];
--- a/packages/docs/pages/margin.tsx
+++ b/packages/docs/pages/margin.tsx
@@ -1,55 +1,67 @@
 import { Button, H1, Panel, Text } from '@bigcommerce/big-design';
-import React from 'react';
+import React, { Fragment } from 'react';
 
-import { CodePreview, NextLink, PageNavigation } from '../components';
+import { Code, CodePreview, ContentRoutingTabs, NextLink } from '../components';
 import { MarginPropTable } from '../PropTables';
 
 const MarginPage = () => {
-  const items = [
-    {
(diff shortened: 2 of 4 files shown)


packages/docs/PropTables/MarginPropTable.tsx
```diff
@@ -5,37 +5,37 @@ import { NextLink, Prop, PropTable, PropTableWrapper } from '../components';
 const marginProps: Prop[] = [
   {
     name: 'margin',
-    types: <NextLink href="/margin">Margin</NextLink>,
+    types: <NextLink href="/spacing">Spacing</NextLink>,
```
MariaJose, line 8: Updated these as `Spacing` since I think it's more descriptive. Is that ok? Thoughts?

packages/docs/pages/margin.tsx
```diff
@@ -1,55 +1,67 @@
...
+              ),
+            },
+          ]}
+        />
+      </Panel>
+
+      <Panel header="Props" headerId="props">
+        <MarginPropTable renderPanel={false} />
+      </Panel>
     </>
```
MariaJose, line 65: Note: 
I was thinking of adding the following in the `Guideline` section: 
All values must be of{' '}<NextLink href="/spacing">Spacing</NextLink> type.

However, I feel that the `<MarginPropTable/>` shows the Spacing as the type, so might not be necessary to mention it in the `Guideline` section. 

Other than that, I think there is nothing important that I can think of 🤔 ... Does someone has a suggestion of what could be added in this section?

packages/docs/pages/padding.tsx
```diff
@@ -1,59 +1,74 @@
...
+              ),
+            },
+          ]}
+        />
+      </Panel>
+
+      <Panel header="Props" headerId="props">
+        <PaddingPropTable renderPanel={false} />
+      </Panel>
     </>
```
MariaJose, line 72: Note:
I was thinking of adding the following in the Guideline section:
All values must be of{' '}Spacing type.

However, I feel that the <PaddingPropTable/> shows the Spacing as the type, so might not be necessary to mention it in the Guideline section.

Other than that, I think there is nothing important that I can think of 🤔 ... Does someone has a suggestion of what could be added in this section?

packages/docs/pages/margin.tsx
```diff
@@ -1,55 +1,67 @@
...
-  ];
-
   return (
     <>
       <H1>Margin</H1>
 
-      <PageNavigation items={items} />
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          BigDesign provides the <Code primary>margin</Code> prop which can be used in some of our components.
```
chanceaclark, line 14: Maybe we can add some sort of verbiage here to indicate that each component will show it supports this prop in the `Inherited` section of its props (for both pages)? What do you think?

packages/docs/pages/margin.tsx
chanceaclark: I think it's safe to omit that sentence. `Spacing` is perfect enough and I have some future plans to show what values you can use.

bc-andreadao (approved):

packages/docs/pages/margin.tsx
MariaJose: I like that 👍🏼 ! that will be more informative. I'll think about how to redact that

jorgemoya (approved):

packages/docs/pages/margin.tsx
```diff
@@ -1,55 +1,68 @@
...
-
   return (
     <>
       <H1>Margin</H1>
 
-      <PageNavigation items={items} />
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          BigDesign provides <Code primary>padding</Code> prop which can be used in some of our components. Check out a
+          components' prop table to see if it inherits this prop.
```
MariaJose, line 15: I updated this sentence, When having a chance could you all check if this is ok? Thanks :)! @bigcommerce/dev-docs

packages/docs/pages/margin.tsx
```diff
@@ -11,7 +11,8 @@ const MarginPage = () => {
 
       <Panel header="Overview" headerId="overview">
         <Text>
-          BigDesign provides the <Code primary>margin</Code> prop which can be used in some of our components.
+          BigDesign provides <Code primary>padding</Code> prop which can be used in some of our components. Check out a
```
bc-andreadao, line 14:
```suggestion
          BigDesign provides <Code primary>padding</Code> prop which can be used for some of our components. Check out a
```

chanceaclark (approved):
Looks great!

packages/docs/pages/margin.tsx
bc-andreadao: Added a couple of edits

packages/docs/pages/margin.tsx
```diff
@@ -1,55 +1,68 @@
...
-  ];
-
   return (
     <>
       <H1>Margin</H1>
 
-      <PageNavigation items={items} />
+      <Panel header="Overview" headerId="overview">
+        <Text>
+          BigDesign provides <Code primary>margin</Code> prop which can be used for some of our components. Check out a
```
MariaJose, line 14: I was thinking if it would be ok to add an `a` in here?: `BigDesign provides a margin prop...`
@bc-andreadao 

Thanks!

bc-andreadao: @MariaJose Yes, that would be a good idea

PR #762, opened 2022-03-15
fix(component): add missing aria roles to tabs

rtalvarez:
## What?
Add aria roles to the Tabs component

## Why?
Missing aria roles in a few places:

- Each tab must have an `aria-expanded` role that is set to `true` or `false` if it is active
- Each tab must have an `aria-controls` role that points to the content it displays (this relies on the developer adding the right IDs, so I've updated the example to reflect this)

[Aria rules Source](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role#description)

**Because we don't have access to the tab panel content, I have assumed IDs of the form `${id}-content`. Open to feedback here, not super reliable, but I don't see another option (short of adding a completely new `TabItem` which opens up its own can of worms)**

## Testing/Proof

aria tabs 2

### DIFF
--- a/packages/big-design/src/components/Tabs/Tabs.tsx
+++ b/packages/big-design/src/components/Tabs/Tabs.tsx
@@ -3,6 +3,7 @@ import React, { HTMLAttributes, memo } from 'react';
 import { StyledTab, StyledTabs } from './styled';
 
 export interface TabItem {
+  ariaControls?: string;
   id: string;
   title: string;
   disabled?: boolean;
@@ -29,11 +30,11 @@ export const Tabs: React.FC<TabsProps> = memo(
     return (
       <>
         <StyledTabs {...props} flexDirection="row" role="tablist">
-          {items.map(({ id, title, disabled }) => (
+          {items.map(({ ariaControls, id, title, disabled }) => (
             <StyledTab
               activeTab={activeTab}
               aria-expanded={id === activeTab ? 'true' : 'false'}
-              aria-controls={`${id}-content`}
+              aria-controls={ariaControls || `${id}-content`}
               id={id}
               key={id}
               onClick={handleOnTabClick}
--- a/packages/big-design/src/components/Tabs/Tabs.tsx
+++ b/packages/big-design/src/components/Tabs/Tabs.tsx
@@ -32,6 +32,8 @@ export const Tabs: React.FC<TabsProps> = memo(
           {items.map(({ id, title, disabled }) => (
             <StyledTab
               activeTab={activeTab}
+              aria-expanded={id === activeTab ? 'true' : 'false'}
+              aria-controls={`${id}-content`}
               id={id}
               key={id}
               onClick={handleOnTabClick}
--- a/packages/big-design/src/components/Tabs/Tabs.tsx
+++ b/packages/big-design/src/components/Tabs/Tabs.tsx
@@ -1,5 +1,7 @@
 import React, { HTMLAttributes, memo } from 'react';
 
+import { warning } from '../../utils';
+
 import { StyledTab, StyledTabs } from './styled';
 
 export interface TabItem {
@@ -26,6 +28,17 @@ export const Tabs: React.FC<TabsProps> = memo(
         onTabClick(tabId);
       }
     };
+
+    const activeItem = activeTab && items[Number(activeTab)];
+    const missingAriaControls = items.some((item) => !item.ariaControls);
+    const missingFallback =
+      activeItem && !document.getElementById(activeItem.ariaControls || `${activeItem.id}-content`);
+
+    if (missingAriaControls || missingFallback) {
+      warning(
+        'TabItems must have an ariaControls field, otherwise, an element with fallback ID "{tab.id}-content" must exist in the DOM',
+      );
+    }
 
     return (
       <>
--- a/packages/docs/pages/tabs.tsx
+++ b/packages/docs/pages/tabs.tsx
@@ -38,12 +38,18 @@ const TabsPage = () => {
 
             return (
               <>
-                <Tabs activeTab={activeTab} items={items} onTabClick={setActiveTab} />
+                <Tabs
+                  aria-label="Example Tab Content"
+                  activeTab={activeTab}
+                  items={items}
+                  id="tab-example"
+                  onTabClick={setActiveTab}
+                />
                 <Box marginTop="large">
-                  {activeTab === 'tab1' && <Text>Content 1</Text>}
-                  {activeTab === 'tab2' && <Text>Content 2</Text>}
-                  {activeTab === 'tab3' && <Text>Content 3</Text>}
-                  {activeTab === 'tab4' && <Text>Content 4</Text>}
+                  {activeTab === 'tab1' && <Text id="tab1-content">Content 1</Text>}
+                  {activeTab === 'tab2' && <Text id="tab2-content">Content 2</Text>}
+                  {activeTab === 'tab3' && <Text id="tab3-content">Content 3</Text>}
+                  {activeTab === 'tab4' && <Text id="tab4-content">Content 4</Text>}
                 </Box>
               </>
             );
--- a/packages/docs/pages/tabs.tsx
+++ b/packages/docs/pages/tabs.tsx
@@ -30,10 +30,10 @@ const TabsPage = () => {
             const [activeTab, setActiveTab] = useState('tab1');
 
             const items = [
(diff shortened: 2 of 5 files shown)


MariaJose (approved):
Lgtm 👍🏼 , just a minor comment

bc-andreadao:
Good on the copyedit front

packages/big-design/src/components/Tabs/Tabs.tsx
```diff
@@ -32,6 +32,8 @@ export const Tabs: React.FC<TabsProps> = memo(
           {items.map(({ id, title, disabled }) => (
             <StyledTab
               activeTab={activeTab}
+              aria-expanded={id === activeTab ? 'true' : 'false'}
+              aria-controls={`${id}-content`}
```
jorgemoya, line 36: I would default it but also accept a template.

packages/docs/pages/tabs.tsx
```diff
@@ -38,12 +38,18 @@ const TabsPage = () => {
...
                 <Box marginTop="large">
-                  {activeTab === 'tab1' && <Text>Content 1</Text>}
-                  {activeTab === 'tab2' && <Text>Content 2</Text>}
-                  {activeTab === 'tab3' && <Text>Content 3</Text>}
-                  {activeTab === 'tab4' && <Text>Content 4</Text>}
+                  {activeTab === 'tab1' && <Text id="tab1-content">Content 1</Text>}
+                  {activeTab === 'tab2' && <Text id="tab2-content">Content 2</Text>}
+                  {activeTab === 'tab3' && <Text id="tab3-content">Content 3</Text>}
+                  {activeTab === 'tab4' && <Text id="tab4-content">Content 4</Text>}
                 </Box>
```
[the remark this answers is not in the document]
jorgemoya: @MariaJose brings out a good point. 

Can we also add a warning if the `id` is not being used? I don't know how feasible this is to implement. I also don't know if the browser does this by default.

packages/big-design/src/components/Tabs/Tabs.tsx
rtalvarez: Do you think this should be a required prop? I kinda don't want to force people to use it if there's a default behavior, but if we don't make it required, some people might not match the ids correctly

rtalvarez: Thinking about adding `ariaControls?: string;` to the `TabItem` interface

jorgemoya: Yeah, definitely optional.

packages/docs/pages/tabs.tsx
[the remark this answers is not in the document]
MariaJose: Yeah, probably adding a more detailed note in this example (see example above). And a general one in the `do`s section? thoughts?

Screen Shot 2022-03-16 at 1 42 00 PM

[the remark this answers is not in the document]
jorgemoya: @rtalvarez Just to be clear, what I meant was if there is a way to notify developer (via a console warning, maybe), that the element with the `id` of `tab1-content` isn't in the dom? I don't know if the browser does this automatically.

packages/docs/pages/tabs.tsx
```diff
@@ -87,6 +93,10 @@ const TabsPage = () => {
             <>
               Use <Code primary>Tabs</Code> to switch between related content.
             </>,
+            <>
+              Use the <Code primary>ariaControls</Code> prop to match your content with the <Code primary>TabItem</Code>{' '}
+              that displays it.
+            </>,
```
MariaJose, line 99: 🙆🏼‍♀️

packages/docs/pages/tabs.tsx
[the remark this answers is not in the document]
rtalvarez: Do we do this anywhere else? It would warn on every re-render, I'm a bit hesitant to spam the console :|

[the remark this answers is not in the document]
jorgemoya: Good point. Honestly I don't think so? But I've seen libraries do it. I'm only concerned because I imagine some will not use ids haha.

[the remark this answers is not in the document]
chanceaclark: So we do have a `warning` method, like Reacts `warning`/`error` methods that only fire in development environments (so prod will be safe). We use it in our `FormGroup` component if you want to look at an example.
---
Ideally we'd add these accessibility controls for the user instead. Maybe it might be time to revisit this API as a `Tab.Item` approach? 🤔  Or potentially we ship a provider? Many options, but may be worth changing before we move out of "beta"?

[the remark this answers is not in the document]
jorgemoya: Problem with this API that I can think of, it is still up to the developer to pass in the "appropriate" component to the render prop that:

1. Accepts an `id` prop
2. Actually attaches that `id` to the DOM element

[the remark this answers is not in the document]
rtalvarez: Haven't messed around too much with API changes, but I was playing around with a warning for this. It feels kind of weird because I need to execute additional checks and DOM querying just to print a message, eg:

```javascript
    const showWarning = items.some((item) =>
      Boolean(!item.ariaControls && document.getElementById(`${item.id}-content`)),
    );

    if (showWarning) {
      warning(
        'TabItems must have an ariaControls field, otherwise, an element with fallback ID "{tab.id}-content" must exist in the DOM',
      );
    }
```

Is the overhead worth it? 🤔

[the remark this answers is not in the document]
chanceaclark: I think we should include it for now. At least there will be a warning for developers. I'd even consider making it an `error` to enforce it better 🤷

packages/big-design/src/components/Tabs/Tabs.tsx
```diff
@@ -26,12 +29,25 @@ export const Tabs: React.FC<TabsProps> = memo(
...
+    }
+
     return (
       <>
         <StyledTabs {...props} flexDirection="row" role="tablist">
-          {items.map(({ id, title, disabled }) => (
+          {items.map(({ ariaControls, id, title, disabled }) => (
             <StyledTab
               activeTab={activeTab}
+              aria-expanded={id === activeTab ? 'true' : 'false'}
```
chanceaclark, line 49: I think this should be `aria-selected` not `aria-expanded`. Was looking at the W3's example: https://codepen.io/pen and noticed they use that instead

chanceaclark (approved):
I'm good with this, minus the `aria-selected` comment.

However, I still think we are missing some accessibility stuff after reading [this article](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html), specifically ensuring that the panel has a `tabpanel` role. I still think we should look into refactoring this components API to either be, more controlled and auto-generate ids and accessibility attributes _or_ provide hooks that auto-generate stuff too but keep the current uncontrolled behavior.

packages/big-design/src/components/Tabs/Tabs.tsx
rtalvarez: Interesting, [MDN says](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role#description)

```
The tab elements not the tablist, have the [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) attribute. Set to aria-expanded="true" for the tabs associated with each visible tabpanel. The tabs associated with hidden tabpanel elements have their aria-expanded attributes set to false.
```

rtalvarez: But [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected#tab) they say aria-selected

rtalvarez: I'm going to stay with `expanded`, I did some testing with Voiceover (works with both - only notes the tab is selected/expanded), but the chrome screen reader did not work with `selected` 🤷

chanceaclark: Those docs might be wrong then with `aria-expanded` 🤔  I even saw on the [`role=tab`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#associated_roles_and_attributes) page that it should be `selected`

PR #764, opened 2022-03-21
fix(component): dont open select on focus

rtalvarez:
## What?

Don't open the select component when focus occurs

## Why?

It is not intended behavior, it's not native behavior, and nobody else does this.
It also creates some issues with modals, if the select is the first element it auto expands it causing weird UX

https://user-images.githubusercontent.com/6025510/159347918-22b5b472-5de1-4be6-95aa-c74019bb2653.mov

## Testing/Proof

Verified locally

### DIFF
--- a/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
@@ -330,8 +330,6 @@ export const MultiSelect = typedMemo(
                 !isOpen && openMenu();
               },
               onFocus: (event) => {
-                !isOpen && openMenu();
-
                 if (typeof props.onFocus === 'function') {
                   props.onFocus(event);
                 }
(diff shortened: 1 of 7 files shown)


MariaJose (commented):
Looks good 👍🏼 , just curious if we want to apply the same logic for the `multiSelect` component as well?

chanceaclark (approved):

MariaJose (approved):

jorgemoya (commented):
Can you do the same for `MultiSelect`?

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -330,8 +330,6 @@ export const MultiSelect = typedMemo(
                 !isOpen && openMenu();
               },
               onFocus: (event) => {
-                !isOpen && openMenu();
-
```
MariaJose, line 334: ❤️

rtalvarez: Thanks for the comment i completely forgot `MultiSelect` was a thing HAHA 😅 😅

PR #791, opened 2022-04-21
fix: tiny-async-pool api changes

chanceaclark:
## What?

Changes how we use `tiny-async-pool`. They updated their library to use an iterator approach so we needed to adjust that logic.

## Testing/Proof

See CI.

### DIFF
--- a/packages/big-design-icons/scripts/build-flags.js
+++ b/packages/big-design-icons/scripts/build-flags.js
@@ -42,9 +42,7 @@ async function convertToReactComponent(filePath, iconName) {
   return outputFile(destPath, code);
 }
 
-async function generateFlags() {
-  const iconFiles = await glob(SOURCE);
-
+function convertWithConcurrencyPool(iconFiles) {
   return asyncPool(cpus().length, iconFiles, (iconFilePath) => {
     const filename = basename(iconFilePath, '.svg');
     const name = `${filename.replace('-', '').toUpperCase()}FlagIcon`;
@@ -56,6 +54,13 @@ async function generateFlags() {
 
     return convertToReactComponent(iconFilePath, name);
   });
+}
+
+async function generateFlags() {
+  const iconFiles = await glob(SOURCE);
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for await (const _result of convertWithConcurrencyPool(iconFiles));
 }
 
 function cleanDestDirectory() {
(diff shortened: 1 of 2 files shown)


packages/big-design-icons/scripts/build-flags.js
```diff
@@ -55,7 +55,9 @@ async function generateFlags() {
     console.log(`Building: ${name}`);
 
     return convertToReactComponent(iconFilePath, name);
-  });
+  })) {
+    await buildPromise;
```
chanceaclark, line 59: Not sure of a better approach for looping through iterators. @deini do you have a better idea?

bc-maksym-konohorov (approved):

packages/big-design-icons/scripts/build-flags.js
deini: Since we aren't using the result, what about something like:

```ts
async function generateIcons() {
  const iconFiles = await glob(SOURCE);

  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  for await (const _result of asyncPool(cpus().length, iconFiles, (iconFilePath) => {
    const filename = basename(iconFilePath, '.svg');
    const name = `${camelcase(filename, { pascalCase: true })}Icon`;

    componentNames.add(name);

    // eslint-disable-next-line no-console
    console.log(`Building: ${name}`);

    return convertToReactComponent(iconFilePath, name);
  }));
}
```

Maybe moving the `asyncPool` call to a fn may make it easier to read too.

chanceaclark: I was trying to avoid an `eslint-disable` 😂  but good idea. Updated.

deini (approved):

packages/big-design-icons/scripts/build.js
```diff
@@ -24,10 +24,9 @@ async function convertToReactComponent(filePath, iconName) {
   return outputFile(destPath, code);
 }
 
-async function generateIcons() {
-  const iconFiles = await glob(SOURCE);
-
-  return asyncPool(cpus().length, iconFiles, (iconFilePath) => {
+async function convertWithConcurrencyPool(iconFiles) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for await (const _result of asyncPool(cpus().length, iconFiles, (iconFilePath) => {
```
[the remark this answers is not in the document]
chanceaclark: 🤦

PR #811, opened 2022-05-03
chore(configs): add eslint rule for curly braces presence

MariaJose:
## What?

Add eslint rule for curly braces presence.
Reference: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md

## Why?
Prevent using curly braces when adding a string as a value in a prop.

Screen Shot 2022-05-03 at 2 01 12 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
     '@typescript-eslint/no-use-before-define': 'off',
     'no-console': 'warn',
     'react/display-name': 'off',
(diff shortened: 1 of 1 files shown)


MariaJose:
Note: Not longer necessary. I'll close this PR soon

MariaJose:
Adding rule in here: https://github.com/bigcommerce/eslint-config/pull/171

PR #819, opened 2022-05-10
fix(docs): fix typo in DropdownPropTable

exeDog:
## What?

Fix typo in DropDown props doc page.

## Why?

'text' was printed twice on screen under Dropdown[DropdownItem].

## Testing/Proof

Before
 
image

After

image


packages/docs/PropTables/DropdownPropTable.tsx
```diff
@@ -132,8 +132,7 @@ const dropdownItemProps: Prop[] = [
   },
   {
     name: 'type',
-    types: "'text'",
-    defaultValue: "'text'",
+    types: 'text',
```
chanceaclark, line 135: I think we still want the `'` around this since it's referring to an enum value.
```suggestion
    types: "'text'",
```

chanceaclark (approved):

jorgemoya (approved):

bc-andreadao (approved):

PR #850, opened 2022-06-01
chore: bump React dependencies

jorgemoya:
## What
Bump react to v18 and related dependencies.

## Issues
- I had to recreate `yarn.lock` to make my environment work. Seemed to be a common issue with bumping react types.
- I couldn't use latest `react-datepicker` because it was breaking jest.
- I found an issue with `Datepicker` (manually entering the date won't do anything).


packages/big-design/src/components/Button/spec.tsx
```diff
@@ -189,19 +189,6 @@ describe('isLoading', () => {
...
-      <Button isLoading={true} onClick={mockClickHandler}>
-        Button
-      </Button>,
-    );
-
-    userEvent.click(screen.getByRole('button'), undefined, { skipPointerEventsCheck: true });
-
-    expect(mockClickHandler).not.toHaveBeenCalled();
-  });
-
```
jorgemoya, line 204: Removed because `skipPointerEventsCheck` is no longer a thing (?? not sure about this), so it would error out.

packages/big-design/src/components/Counter/spec.tsx
```diff
@@ -261,7 +262,7 @@ test('value increases and decreases with arrow keypresses', () => {
   const { getByDisplayValue } = render(counterMock(requiredAttributes));
 
   const counter = getByDisplayValue('5') as HTMLInputElement;
-  counter.focus();
+  act(() => counter.focus());
```
jorgemoya, line 265: Wrapping in `act` was easier fix, we could potentially update this file to use `userEvent`.

packages/big-design/src/components/Datepicker/spec.tsx
```diff
@@ -60,39 +60,33 @@ test('renders select label', () => {
...
 
-  fireEvent.input(input as HTMLInputElement, {
-    target: {
-      value: dateString,
-    },
-  });
+  await userEvent.type(input as HTMLInputElement, dateString);
 
-  expect(changeFunction).not.toBeCalled();
+  expect(changeFunction).toBeCalled();
```
jorgemoya, line 89: This is a big issue. Our `Datepicker` component doesn't update date when entered manually in my local environment, but on tests it looks like it does. Will open a github issue for this.

packages/big-design/src/components/Modal/spec.tsx
```diff
@@ -263,11 +263,11 @@ test('renders destructive action button', () => {
 
 test('unmounts appropriately', () => {
   const onClick = jest.fn();
-  const { getByTestId, rerender, unmount } = render(<Modal isOpen={true} />);
+  const { getByTestId, unmount } = render(<Modal isOpen={true} />);
 
   unmount();
 
-  rerender(
+  render(
```
jorgemoya, line 270: I was getting this error when rerendering: `Cannot update an unmounted root.`.

packages/big-design/src/components/Datepicker/spec.tsx
jorgemoya: #852

packages/big-design/src/components/Button/spec.tsx
chanceaclark: Could we just get rid of the `{ skipPointerEventsCheck: true }` object instead of the test?

packages/big-design/src/components/Counter/spec.tsx
chanceaclark: Could you add a `TODO` comment here so we know to come back around to it?

packages/big-design/src/components/Table/DataCell/DataCell.tsx
```diff
@@ -6,6 +6,7 @@ import { StyledTableDataCell, StyledTableDataCheckbox } from './styled';
 
 export interface DataCellProps extends TableHTMLAttributes<HTMLTableCellElement>, TableColumnDisplayProps {
   align?: 'left' | 'center' | 'right';
+  children: React.ReactNode;
```
chanceaclark, line 9:
```suggestion
  children?: React.ReactNode;
```

packages/docs/components/CodePreview/CodePreview.tsx
```diff
@@ -57,14 +57,20 @@ function transformCode(input: string): string {
...
 export interface CodePreviewProps {
+  children?: React.ReactNode | (() => JSX.Element);
   scope?: { [key: string]: unknown };
 }
 
 export const CodePreview: React.FC<CodePreviewProps> = (props) => {
   const { children } = props;
+
+  // We are intentionally casting the children to React.ReactNode because our code
+  // doesn't know that our custom babel loader converts () => JSX.Element to a string.
```
chanceaclark, line 68: Whoever wrote this comment must be awesome 😜

chanceaclark (commented):
Some minor comments but looks good.

packages/big-design/src/components/MultiSelect/spec.tsx
```diff
@@ -381,11 +382,9 @@ test('enter should trigger onOptionsChange', async () => {
...
   const input = screen.getByTestId('multi-select');
 
-  await act(async () => {
-    await fireEvent.click(input);
-    await fireEvent.keyDown(input, { key: 'ArrowDown' });
-    await fireEvent.keyDown(input, { key: 'Enter' });
-  });
+  await userEvent.click(input);
+
+  await userEvent.keyboard('[arrowdown][enter]');
```
MariaJose, line 387: 🙌🏼

packages/big-design/src/components/Modal/spec.tsx
MariaJose: Interesting, probably because the `button` wasn't the one that was rendered at first 🤷🏼‍♀️ 🤔

MariaJose (approved):
Lgtm 💯 !

chanceaclark (approved):
Looks like we have some peerDeps that are missing upgrades, but we might need to tackle that in the future/wait for those libraries to update.

jorgemoya:
> Looks like we have some peerDeps that are missing upgrades, but we might need to tackle that in the future/wait for those libraries to update.

How do you find these?

chanceaclark:
> > Looks like we have some peerDeps that are missing upgrades, but we might need to tackle that in the future/wait for those libraries to update.
> 
> How do you find these?

With `yarn install` any of the ones with `dfd69583-bd17-43d3-87e0-af94dd7185e6`

Screen Shot 2022-06-03 at 13 34 19

bc-andreadao (approved):

bc-apostoliuk (approved):

PR #860, opened 2022-06-08
fix(component): StatefulTable and Table has rendering hiccup when dragging & dropping

MariaJose:
## What?
When dragging and dropping a row in a StatefulTable and Table, the columns widths change their values.

## Why?
Width should stay the same when dragging and dropping.

**Before:**

https://user-images.githubusercontent.com/39739180/172660167-2e62130e-450a-4f5c-8bdc-c0b25fe862a5.mp4

**After:**
<table>
<tr>
	<td>With width
	<td>With no width
<tr>
<tr>
       <td><video src= "https://user-images.githubusercontent.com/39739180/172660561-0819f391-f576-4d22-bfc0-854080ebc174.mp4"/>
	<td><video src= "https://user-images.githubusercontent.com/39739180/172660600-395d188c-b652-4840-95c5-4628b81329a6.mp4" />
<tr>
</table>


jorgemoya (approved):
This looks great!

packages/big-design/src/components/Table/Table.tsx
```diff
@@ -95,10 +96,30 @@ const InternalTable = <T extends TableItem>(props: TableProps<T>): React.ReactEl
...
+    const headerCellIconWidth = headerCellIconId?.getBoundingClientRect().width ?? 'auto';
+
+    const headerCellsWidths = columns.map((_column, index) => {
+      const headerCellElement = window.document.getElementById(`header-cell-${index}`);
+
+      if (headerCellElement !== null) {
+        return headerCellElement.getBoundingClientRect().width;
+      }
+
+      return 'auto';
```
chanceaclark, line 116: 🍹 
```suggestion
      return headerCellElement?.getBoundingClientRect().width ?? 'auto';
```

packages/big-design/src/components/Table/Table.tsx
```diff
@@ -95,10 +96,30 @@ const InternalTable = <T extends TableItem>(props: TableProps<T>): React.ReactEl
...
         onRowDrop(source.index, destination.index);
       }
+
+      setHeaderCellWidths([]);
     },
     [onRowDrop],
   );
 
+  const onBeforeDragStart = () => {
+    const headerCellIconId = window.document.getElementById('header-cell-icon');
```
chanceaclark, line 106: 🍹 Maybe keeping track of a `ref` for this `th` may be better than having an `id` 🤔

chanceaclark (approved):

packages/big-design/src/components/Table/Table.tsx
MariaJose: @chanceaclark you're right 👍🏼 . For getting the `width`, I used `offseetWidth` that apparently considers the padding and boarding as well.  Reference: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth. 
Is that ok?

chanceaclark: Should should be :gucci:

PR #861, opened 2022-06-09
refactor(docs): update GTM logic using recommendations

chanceaclark:
## What?

_Hint: hide whitespace will be easier for review._ 

Updates our GTM logic using [NextJS's recommendations](https://github.com/vercel/next.js/blob/748c8e8002987a68396858f4b5ca3eda9bd6d0d2/examples/with-google-analytics/pages/_app.js).

## Why?

- Uses `next/script` functionality instead of adding it manually to the `next/document` `Head` component.`
- Converts `_app.tsx` to a functional component.
- Exposes the `GTM_ID` via `.env.production` so we don't have to include it on the build script
- Added `lang="en"` to html element

### PageEvents being fired
Screen Shot 2022-06-08 at 16 58 20

### Inlined var on build
Screen Shot 2022-06-09 at 08 13 55

## Testing/Proof

See screenshots.


.gitignore
```diff
@@ -10,4 +10,5 @@ packages/docs/pages/dev.tsx
 .idea
 .DS_Store
 .parcel-cache
-.tsbuildinfo
\ No newline at end of file
+.tsbuildinfo
+.env*.local
```
chanceaclark, line 14: This is NextJS's recommendation to only `.gitignore` the `.local` postfix files.

packages/docs/.env.production
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_GTM_ID="GTM-KL3MKSK"
```
chanceaclark, line 1: Prefixing with `NEXT_PUBLIC_*` exposes it to the client.

packages/docs/pages/_document.tsx
```diff
@@ -2,14 +2,8 @@ import Document, { DocumentContext, Head, Html, Main, NextScript } from 'next/do
...
 import { resetServerContext } from 'react-beautiful-dnd';
 import { ServerStyleSheet } from 'styled-components';
-
-import { GTM_ID, GTM_URL } from '../utils/analytics/gtm';
-
-const isProd = process.env.NODE_ENV === 'production';
-
 export default class AppDocument extends Document {
   static async getInitialProps(ctx: DocumentContext) {
-    resetServerContext();
```
chanceaclark, line 12: Moved this to the `finally` block. Just for 💅

packages/docs/utils/analytics/gtm.ts
```diff
@@ -1,30 +1,15 @@
-const GTM_CONFIG = {
-  ID: process.env.GTM_ID || '',
-  AUTH: process.env.GTM_AUTH || '',
-  PREVIEW: process.env.GTM_PREVIEW || '',
-  COOKIES_WIN: process.env.GTM_COOKIES_WIN || '',
```
chanceaclark, line 5: We never used these, so removed.

rtalvarez (approved):

bc-andreadao (approved):

MariaJose (approved):
👍🏼 !

PR #864, opened 2022-06-09
fix(component): align radio buttons with custom label

jorgemoya:
## What?

Align items `center` instead of `flex-start`.

## Why?

With a custom label, radio button won't align to the content. Not sure if this is expected, if so we can close this.

|Before|After|
|-----|-----|
| Screen Shot 2022-06-09 at 3 11 13 PM | Screen Shot 2022-06-09 at 3 11 32 PM |

## Testing/Proof

Locally.

### DIFF
--- a/packages/big-design/src/components/Radio/__snapshots__/spec.tsx.snap
+++ b/packages/big-design/src/components/Radio/__snapshots__/spec.tsx.snap
@@ -23,10 +23,10 @@ exports[`render Radio (checked + disabled) 1`] = `
 }
 
 .c0 {
-  -webkit-align-items: flex-start;
-  -webkit-box-align: flex-start;
-  -ms-flex-align: flex-start;
-  align-items: flex-start;
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
+  align-items: center;
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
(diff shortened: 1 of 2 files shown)


chanceaclark (approved):
Code wise, I'm good with this however we should get someone from @bigcommerce/product-design to review this too.

bc-apostoliuk (approved):

MariaJose (approved):

eugene-polev:
Why are we making this change? The radio selector should be aligned with the first line of its label (like it's displayed in the "before" example).

jorgemoya:
> Why are we making this change? The radio selector should be aligned with the first line of its label (like it's displayed in the "before" example).

To fix use cases with custom components:

#### If I leave as is, it looks weird.
Screen Shot 2022-06-10 at 10 06 11 AM

#### If I want to align center, I can't.
Screen Shot 2022-06-10 at 10 06 34 AM

Should we avoid this pattern altogether?

eugene-polev:
> Should we avoid this pattern altogether?

IMO we shouldn't insert the input field inside the radio component in the example. There is a radio button, that consists of a radio toggle and a label, and some content/functionality dependent on the state of the radio button. In this case, we substituted the label with input (dependent content) but this input isn't a part of the radio button. So we have a container with the radio button with the label turned off and input. Both are vertically aligned to center. 

image

PR #874, opened 2022-06-14
chore(ci): Enabled type assertions lint rule

bc-juanvasquez:
## What?

Enable @typescript-eslint/consistent-type-assertions and either fix easy issues or disable lines. 

## Why?

Eslint rules are currently disabled in Big Design in order to use BigCommerce’s public eslint config. 

N/A

## Testing/Proof

All build checks pass

### DIFF
--- a/packages/big-design/src/components/Checkbox/Label/styled.tsx
+++ b/packages/big-design/src/components/Checkbox/Label/styled.tsx
@@ -9,6 +9,7 @@ export interface StyledLabelProps extends React.LabelHTMLAttributes<HTMLLabelEle
   disabled?: boolean;
 }
 
+// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
 export const StyledLabel = styled(StyleableText).attrs({
   as: 'label',
 })<StyledLabelProps>`
--- a/packages/big-design/src/components/Checkbox/Label/styled.tsx
+++ b/packages/big-design/src/components/Checkbox/Label/styled.tsx
@@ -3,14 +3,16 @@ import { hideVisually } from 'polished';
 import styled, { css, DefaultTheme, StyledComponent } from 'styled-components';
 
 import { StyleableText } from '../../Typography/private';
+import { TextProps } from '../../Typography/types';
 
 export interface StyledLabelProps extends React.LabelHTMLAttributes<HTMLLabelElement> {
   hidden?: boolean;
   disabled?: boolean;
 }
 
-// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-export const StyledLabel = styled(StyleableText).attrs({
+export const StyledLabel = styled<
+  StyledComponent<'label' | 'p', DefaultTheme, Partial<TextProps>> & StyledLabelProps
+>(StyleableText).attrs({
   as: 'label',
 })<StyledLabelProps>`
   cursor: pointer;
@@ -22,6 +24,6 @@ export const StyledLabel = styled(StyleableText).attrs({
     `}
 
   ${({ hidden }) => hidden && hideVisually()}
-` as StyledComponent<'label', DefaultTheme, StyledLabelProps>;
+`;
 
 StyledLabel.defaultProps = { theme: defaultTheme };
--- a/packages/big-design/src/components/Datepicker/spec.tsx
+++ b/packages/big-design/src/components/Datepicker/spec.tsx
@@ -3,7 +3,6 @@ import 'jest-styled-components';
 import userEvent from '@testing-library/user-event';
 import React, { createRef } from 'react';
 import type { ReactDatePicker } from 'react-datepicker';
-import { act } from 'react-dom/test-utils';
 
 import { fireEvent, render, screen, waitFor } from '@test/utils';
 
@@ -44,11 +43,7 @@ test('calls onDateChange function when a date cell is clicked', async () => {
 
   const cell = await screen.findByRole('option', { current: 'date' });
 
-  await act(async () => {
-    if (cell) {
-      await waitFor(() => fireEvent.click(cell));
-    }
-  });
+  await userEvent.click(cell);
 
   expect(changeFunction).toHaveBeenCalled();
 });
--- a/packages/big-design/src/components/Datepicker/spec.tsx
+++ b/packages/big-design/src/components/Datepicker/spec.tsx
@@ -4,25 +4,27 @@ import React, { createRef } from 'react';
 import type { ReactDatePicker } from 'react-datepicker';
 import { act } from 'react-dom/test-utils';
 
-import { fireEvent, render } from '@test/utils';
+import { fireEvent, render, waitFor } from '@test/utils';
 
 import { FormGroup } from '..';
 
 import { Datepicker } from './index';
 
-test('should use the passed in ref object if provided', () => {
+test('should use the passed in ref object if provided', async () => {
   const ref = createRef<ReactDatePicker>();
   const { container } = render(<Datepicker onDateChange={jest.fn()} ref={ref} />);
 
   const input = container.querySelector('input');
 
-  fireEvent.click(input as HTMLInputElement);
+  if (input) {
+    await waitFor(() => fireEvent.click(input));
+  }
 
   const datepicker = container.querySelector('.react-datepicker');
+  const refClassname =
+    ref.current && ref.current.props.calendarClassName ? ref.current.props.calendarClassName : '';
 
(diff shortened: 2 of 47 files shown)


packages/big-design/src/components/Modal/Modal.tsx
```diff
@@ -69,7 +69,7 @@ const InternalModal: React.FC<ModalProps> = ({
   const initialBodyOverflowYRef = useRef('');
   const internalTrap = useRef<FocusTrap | null>(null);
   const headerUniqueId = useUniqueId('modal_header');
-  const [modalRef, setModalRef] = useState<HTMLDivElement | null>(null);
+  const [modalRef, setModalRef] = useState<HTMLDivElement | HTMLElement | null>(null);
```
chanceaclark, line 72: Do we actually need this type?
```suggestion
  const [modalRef, setModalRef] = useState<HTMLDivElement | null>(null);
```

packages/big-design/src/components/Popover/Popover.tsx
```diff
@@ -115,7 +115,8 @@ const InternalPopover: React.FC<InternalPopoverProps> = ({
   });
 
   useEffect(() => {
-    const prevFocus = previousFocus.current as HTMLElement;
+    // @ts-expect-error current is HTMLElement
+    const prevFocus: HTMLElement = previousFocus.current;
 
     return () => {
       if (prevFocus && typeof prevFocus.focus === 'function') {
```
chanceaclark, line 122: We could check to see if `prevFocus` is an instance of `HTMLElement` instead 🤔

packages/big-design/src/components/StatefulTable/spec.tsx
```diff
@@ -124,9 +124,9 @@ test('renders rows without checkboxes when opting in to selectable', () => {
 
 test('items are unselected by default', () => {
   const { container } = render(getSimpleTable({ selectable: true }));
-  const checkbox = container.querySelector('tbody > tr input') as HTMLInputElement;
+  const checkboxes: NodeListOf<HTMLInputElement> = container.querySelectorAll('tbody > tr input');
```
chanceaclark, line 127: Could we use `findByRole` instead?

packages/big-design/src/components/Table/spec.tsx
```diff
@@ -544,15 +544,19 @@ describe('draggable', () => {
     const spaceKey = { keyCode: 32 };
     const downKey = { keyCode: 40 };
     const { container } = render(<Table columns={columns} items={items} onRowDrop={onRowDrop} />);
-    const dragEl = container.querySelector('[data-rbd-draggable-id]') as HTMLElement;
+    const dragEl: HTMLElement | null = container.querySelector('[data-rbd-draggable-id]');
```
chanceaclark, line 547: Could we use `findByRole` here instead?

packages/big-design/src/hooks/useComponentSize.ts
```diff
@@ -31,6 +31,7 @@ export const useComponentSize = <T extends HTMLElement>(ref: RefObject<T>): Comp
     if (typeof MutationObserver === 'function') {
```
chanceaclark, line 31: Could check if the instance of ref.current is a `Node`

packages/big-design/src/managers/alerts/spec.tsx
```diff
@@ -221,26 +221,26 @@ describe('alertsManager functionality', () => {
 });
 
 describe('AlertsManager', () => {
-  const successAlert = {
+  const successAlert: AlertProps = {
```
chanceaclark, line 224: You can always extend the type to ensure `key` is not undefined...
```ts
type KeyedAlertProps = AlertProps & {
  key: string;
};
```

That way you don't have to write conditionals

packages/docs/components/SideNav/SideNav.tsx
```diff
@@ -30,7 +30,7 @@ export const SideNav: React.FC = () => {
         <SideNavGroup title="Introduction">
           <SideNavLink href="/">Getting Started</SideNavLink>
 
-          <Link title="CodeSandbox Example" url={CodeSandboxUrl} />
+          <Link title="CodeSandbox Example" url={String(CodeSandboxUrl)} />
```
chanceaclark, line 33: This will be a `"undefined"` if it's undefined. Should we just null coalesce to an empty string?

MariaJose, line 33: wondering if we should add `""` in case `CodeSandboxUrl` is undefined or null? I think if `CodeSandboxUrl` is `undefined` or `null`, we could get `'undefined'` or `'null'`

packages/big-design/src/components/Datepicker/spec.tsx
```diff
@@ -65,36 +67,44 @@ test('renders select label', () => {
...
   const input = container.querySelector('input');
 
-  fireEvent.click(input as HTMLInputElement);
+  if (input) {
+    await waitFor(() => fireEvent.click(input));
+  }
 
   const datepicker = container.querySelector('.react-datepicker');
 
   const cell = datepicker?.querySelector('.react-datepicker__day--today');
```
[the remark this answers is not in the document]
bc-juanvasquez: Screen Shot 2022-06-14 at 1 36 12 PM

packages/big-design/src/components/Datepicker/spec.tsx
```diff
@@ -65,36 +67,44 @@ test('renders select label', () => {
...
-test('calls onDateChange function when a date cell is clicked', () => {
+test('calls onDateChange function when a date cell is clicked', async () => {
   const changeFunction = jest.fn();
   const { container } = render(<Datepicker onDateChange={changeFunction} />);
 
   const input = container.querySelector('input');
 
-  fireEvent.click(input as HTMLInputElement);
+  if (input) {
+    await waitFor(() => fireEvent.click(input));
```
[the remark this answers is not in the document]
chanceaclark: Using `findBy*` methods will throw if an element is not found. That way we can ensure an element exists later on in the test.

packages/big-design/src/components/Modal/Modal.tsx
chanceaclark: `HTMLDivElement` extends from `HTMLElement` so it's pretty much the same thing.

packages/big-design/src/components/Popover/Popover.tsx
bc-juanvasquez: this? 
```suggestion
     if (prevFocus && prevFocus instanceof HTMLElement && typeof prevFocus.focus === 'function') {
        prevFocus.focus();
      }
```

packages/big-design/src/components/StatefulTable/spec.tsx
bc-juanvasquez: Screen Shot 2022-06-14 at 2 40 55 PM

packages/big-design/src/components/Datepicker/spec.tsx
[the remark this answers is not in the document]
chanceaclark: Looks like `react-datepicker` added these accessibility roles et. al. in `4.7.0`. Opened a PR to update that: https://github.com/bigcommerce/big-design/pull/877

MariaJose (approved):
lgtm!

packages/big-design/src/components/Checkbox/Label/styled.tsx
```diff
@@ -9,6 +9,7 @@ export interface StyledLabelProps extends React.LabelHTMLAttributes<HTMLLabelEle
   disabled?: boolean;
 }
 
+// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
chanceaclark, line 12: I wonder if there is a TS Generic way to resolve this 🤔

packages/big-design/src/components/Form/spec.tsx
```diff
@@ -17,12 +17,14 @@ test('forwards ref', () => {
   expect(form).toBe(ref.current);
 });
 
-test('calls onSubmit', () => {
+test('calls onSubmit', async () => {
   const onSubmit = jest.fn();
   const { container } = render(<Form onSubmit={onSubmit} />);
-  const form = container.querySelector('form') as HTMLFormElement;
+  const form = container.querySelector('form');
```
chanceaclark, line 23: If you give the `Form` a `name` attribute you should be able to use `findByRole('form')` query.

PR #878, opened 2022-06-15
chore(ci): enabled no-unsafe-member-access rule

bc-juanvasquez:
## What?

Enable @typescript-eslint/no-unsafe-member-access and either fix easy issues or disable lines. 

## Why?

Eslint rules are currently disabled in big design in order to use BigCommerce’s public eslint config. We want to enable the rules from the public config.

N/A

## Testing/Proof

Build Checks pass


packages/big-design/src/components/ButtonGroup/spec.tsx
```diff
@@ -10,14 +10,17 @@ beforeAll(() => {
   Object.defineProperties(window.HTMLElement.prototype, {
     offsetWidth: {
       get() {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
rtalvarez, line 13: You can fix/remove all these by providing typings to `this`

```
      get(this: HTMLElement) {
```

(or whatever other element you're using)

packages/big-design/src/components/PillTabs/spec.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
rtalvarez, line 1: I assume this is the same issue as above, we might be able to get rid of this one with the same approach

packages/big-design/src/components/Dropdown/types.ts
```diff
@@ -1,14 +1,16 @@
...
+
 export interface DropdownProps extends Omit<React.HTMLAttributes<HTMLUListElement>, 'children'> {
   autoWidth?: boolean;
   disabled?: boolean;
   items: Array<DropdownItem | DropdownLinkItem> | DropdownItemGroup[];
   maxHeight?: number;
   placement?: Placement;
   positionFixed?: boolean;
-  toggle: React.ReactElement;
+  toggle: React.ReactElement<ButtonProps>;
```
chanceaclark, line 13: Would this be `ButtonProps` or actually `unknown`? We don't actually enforce this as a `Button` element, so it could be anything.

packages/big-design/src/components/ProgressCircle/styled.tsx
```diff
@@ -11,12 +11,14 @@ export const StyledProgressCircle = styled.svg<ProgressCircleProps>`
...
 `;
 
-export const StyledCircle = styled.circle.attrs(({ size, theme }: any) => ({
-  // rem not usable for circle svg cx, cy, and r values in Safari 14
-  cx: theme.helpers.emCalc(getDimensions(size) / 2),
-  cy: theme.helpers.emCalc(getDimensions(size) / 2),
-  r: theme.helpers.emCalc(getDimensions(size) / 2 - getStrokeWidth(size) / 2),
-}))<ProgressCircleProps>`
+export const StyledCircle = styled.circle.attrs(
+  ({ size, theme }: { theme: DefaultTheme; size: ProgressCircleProps['size'] }) => ({
```
chanceaclark, line 15: I think we could just simplify this
```suggestion
export const StyledCircle = styled.circle.attrs<ProgressCircleProps>(({ size, theme }) => ({
```

packages/big-design/src/components/Table/spec.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
chanceaclark, line 1: Looks like we should actually update the types on `columns` and `items` in this file
```ts
interface SimpleTableOptions {
  // ...
  columns?: Array<TableColumn<Record<string, unknown>>>;
  items?: Array<Record<string, unknown>>;
  // ...
}
```

packages/docs/components/CodePreview/CodePreview.tsx
```diff
@@ -31,6 +31,7 @@ function getInitialCode(children: React.ReactNode, language: Language): string {
     return children;
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
chanceaclark, line 34: `yarn add @types/babel__standalone -D` then fix the issues from there.

packages/docs/components/PropTable/PropTable.tsx
```diff
@@ -117,13 +117,16 @@ const TypesData: React.FC<TypesDataProps> = (props): any => {
   if (Array.isArray(types)) {
     return types.map((type, index) => {
       return (
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
chanceaclark, line 120: It'll be some effort, but let's fix this.

packages/docs/pages/form.tsx
```diff
@@ -239,7 +239,7 @@ const FormPage = () => {
                       const [value, setValue] = useState('BigCommerce');
                       const [error, setError] = useState(ERROR_MSG);
 
-                      const handleSubmit = (event) => {
+                      const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
```
chanceaclark, line 242: 🍹 We generally like to use the component props to type `event`. In case we ever change the `onSubmit` api, we'll be able to tell more easily what's wrong.
```suggestion
                      const handleSubmit: FormProps['onSubmit'] = (event) => {
```

packages/docs/components/CodePreview/CodePreview.tsx
```diff
@@ -31,11 +32,13 @@ function getInitialCode(children: React.ReactNode, language: Language): string {
     return children;
   }
 
-  const code = transform(children, {
+  const transformResult: BabelFileResult = transform(children, {
```
chanceaclark, line 35: We should be able to get rid of this.
```suggestion
  const transformResult = transform(children, {
```

packages/docs/components/CodePreview/CodePreview.tsx
```diff
@@ -31,11 +32,13 @@ function getInitialCode(children: React.ReactNode, language: Language): string {
...
     compact: false,
     retainLines: true,
     presets: [['typescript', { allExtensions: true, isTSX: true, jsxPragma: 'preserve' }]],
-  }).code;
+  });
+
+  const code = transformResult.code ?? '';
```
chanceaclark, line 41: Why don't we just return `children` after the null coalescing.

packages/big-design/src/components/Table/spec.tsx
```diff
@@ -1,14 +1,22 @@
...
+  sku: string;
+  name: string;
+  stock: number;
+}
 
 interface SimpleTableOptions {
   className?: string;
-  columns?: any[];
-  items?: any[];
+  columns?: Array<TableColumn<SimpleTableItem>>;
```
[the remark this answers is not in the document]
bc-juanvasquez: Isn't the point of having TableColumn generic to pass implementation's defined type?

chanceaclark (approved):
Looks great! Thanks for updating!

Issue #859, opened 2022-06-08
`selectable` + draggable columns don't render nicely


chanceaclark:
**Describe the bug**
Using selectable columns alongside draggable columns doesn't render correctly.

**To Reproduce**
```tsx
function Example() {
  const [items, setItems] = useState(data);
  const [selectedItems, setSelectedItems] = useState<Item[]>([]);

  const onDragEnd = (from: number, to: number) =>
    setItems((currentItems) => dragEnd(currentItems, from, to));

  return (
    <Table
      columns={[
        { header: 'Sku', hash: 'sku', render: ({ sku }) => sku },
        { header: 'Name', hash: 'name', render: ({ name }) => name },
        { header: 'Stock', hash: 'stock', render: ({ stock }) => stock },
      ]}
      items={items}
      onRowDrop={onDragEnd}
      selectable={{
        selectedItems,
        onSelectionChange: setSelectedItems,
      }}
    />
  );
}
```

**Expected behavior**
Draggable column should be the same width as its icon.

**Screenshots**
<img width="242" alt="Screen Shot 2022-06-08 at 10 30 35" src="https://user-images.githubusercontent.com/10539418/172657390-d63bac21-e344-4bf6-945d-7340ca5399fe.png">

**Desktop (please complete the following information, if applicable):**
N/A

**Smartphone (please complete the following information, if applicable):**
N/A

**Additional context**

PR #880, opened 2022-06-16
fix(component): add set width if unset

tildemarky:
## What?

Keeps the draggable icon constrained instead of stretching potentially fixes: #859

## Why?

Adding a set width instead of auto keeps the draggable icon column constrained which is useful for when you're creating columns with draggable and selectable elements together. 50px was chosen as it's small enough to still be reasonable on smaller screens while also not constraining the drag icon svg at larger sizes.

https://user-images.githubusercontent.com/99497246/174160129-be342025-91db-45a1-92ab-1893df8c2c0c.mp4

https://user-images.githubusercontent.com/99497246/174160239-3150f090-703e-4601-b0ff-9033bb2caf04.mp4

## Testing/Proof

Test coverage already accommodated the component and this change. Ran the test suite prior to this PR.

### DIFF
--- a/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
+++ b/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
@@ -127,16 +127,10 @@ export const HeaderCheckboxCell: React.FC<HeaderCheckboxCellProps> = memo(
 );
 
 export const DragIconHeaderCell: React.FC<DragIconCellProps> = memo(
-  ({ actionsRef, headerCellIconRef, width }) => {
+  ({ actionsRef, headerCellIconRef }) => {
     const actionsSize = useComponentSize(actionsRef);
 
-    return (
-      <StyledTableHeaderIcon
-        ref={headerCellIconRef}
-        stickyHeight={actionsSize.height}
-        width={width}
-      />
-    );
+    return <StyledTableHeaderIcon ref={headerCellIconRef} stickyHeight={actionsSize.height} />;
   },
 );
 
--- a/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
+++ b/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
@@ -12,7 +12,7 @@ import { Tooltip } from '../../Tooltip';
 import { TableColumnDisplayProps } from '../mixins';
 import { TableColumn, TableItem } from '../types';
 
-import { StyledFlex, StyledTableHeaderCell, StyledTableHeaderCheckbox } from './styled';
+import { StyledFlex, StyledTableHeaderCell, StyledTableHeaderIcon } from './styled';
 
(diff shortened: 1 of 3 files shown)


packages/big-design/src/components/Table/Table.tsx
```diff
@@ -138,7 +138,7 @@ const InternalTable = <T extends TableItem>(
           <DragIconHeaderCell
             actionsRef={actionsRef}
             headerCellIconRef={headerCellIconRef}
-            width={headerCellWidths.length ? headerCellWidths[0] : 'auto'}
+            width={headerCellWidths.length ? headerCellWidths[0] : '50px'}
```
chanceaclark, line 141: Instead of hardcoding this width, inside `/src/components/Table/HeaderCell`, what if we renamed `StyledTableHeaderCheckbox` to be more generic (something like `StyledTableHeaderIcon`) and reuse that component inside `DragIconHeaderCell` instead?

chanceaclark (commented):
Hey @tildemarky! Thanks for making your first contribution and getting your feet wet!

I do have a comment to see if we can change some stuff around, let me know if you have any questions and I'll do my best to clarify!

packages/big-design/src/components/Table/Table.tsx
tildemarky: Oh that's brilliant, a much more reusable solution! Pushing those changes shortly

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -133,7 +131,7 @@ export const DragIconHeaderCell: React.FC<DragIconCellProps> = memo(
     const actionsSize = useComponentSize(actionsRef);
 
     return (
-      <StyledTableHeaderCell
+      <StyledTableHeaderIcon
         ref={headerCellIconRef}
         stickyHeight={actionsSize.height}
         width={width}
```
chanceaclark, line 137: I believe we should be able to remove the `width` prop now!

chanceaclark (commented):
One minor comment, but overall looks so much better! Thanks!

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
tildemarky: Good point, thought it was being used somewhere else in the file

tildemarky: Would it be worth it also removing the width from the `DragIconCellProps` interface as well?

chanceaclark: That's amazing!

tildemarky:
> Thanks again for this! I'll go ahead and merge it! Tbd on the release though

Amazing! And thank you for helping me out 😄

PR #902, opened 2022-07-05
feat(component): remove tooltip wrapper

chanceaclark:
## What?

Removes the `StyledTooltipTrigger` wrapper from the `Tooltip` component. 

This is a breaking change as a result of it:

```
BREAKING CHANGE:
- Removes the `inline` prop on the tooltip
- Strings are not allowed in the `trigger` prop
```

## Why?

This was causing some layout issues when using it with `Icon` + `Button` components. For example,
```tsx
<Button
  iconLeft={
    <Tooltip placement="bottom" trigger={<InfoIcon size={20} />}>
      Tooltip
    </Tooltip>
  }
  iconRight={<ArrowDropDownIcon />}
>
   Label
</Button>
```

Would misalign the icon within the button:
Screen Shot 2022-07-05 at 13 35 11

This leaves the layout up to the consumer to position it with a button. I have found `inline-flex` and `inline-grid` good alignment `display` values.

Screen Shot 2022-07-05 at 17 18 43

## Testing/Proof

Updated tests to reflect changes.

### DIFF
--- a/packages/big-design/src/components/List/Item/Content.tsx
+++ b/packages/big-design/src/components/List/Item/Content.tsx
@@ -51,9 +51,8 @@ export const Content = memo(({ item, isHighlighted }: ContentProps) => {
   );
 
   const wrapInTooltip = useCallback(
-    (tooltip: string, tooltipTrigger: React.ReactChild) => (
+    (tooltip: string, tooltipTrigger: React.ReactElement) => (
       <Tooltip
-        inline={false}
         modifiers={[{ name: 'preventOverflow' }, { name: 'offset', options: { offset: [0, 20] } }]}
         placement="left"
         trigger={tooltipTrigger}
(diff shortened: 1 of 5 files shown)


rtalvarez (commented):
I wonder if anyone remembers what was the original goal with having this inline prop, can't recall a specific reason myself

Also, if this is a breaking change, I think you can denote so in your commit message, so the release notes reflect this too. I believe you can do this with a `!` after the scope: `feat(component)!: remove tooltip wrapper`

MariaJose (approved):

chanceaclark:
> I wonder if anyone remembers what was the original goal with having this inline prop, can't recall a specific reason myself

I believe it's because that element was a `div` so we didn't want to be a block element when using it alongside text or wrapping a button.

> Also, if this is a breaking change, I think you can denote so in your commit message, so the release notes reflect this too. I believe you can do this with a `!` after the scope: `feat(component)!: remove tooltip wrapper`

With conventional commits, you can either do the `!` or have `BREAKING CHANGE:` in the commit body.

jorgemoya (approved):

bc-andreadao (approved):

PR #903, opened 2022-07-07
chore(icons): remove .tsbuildinfo before releasing

chanceaclark:
## What?

Removes the `.tsbuildinfo` file before releasing the icon package.

## Why?

The current `prepublishOnly` script removes `dist/` folder, thus removing existing types and because we incrementally build icons then next run of the `build` script won't generate the types again. This causes issues when trying to release big-design as it errors out when running `lerna run build --stream`.

Screen Shot 2022-07-07 at 09 37 05

## Testing/Proof

See screenshot.


golcinho (approved):

deini (approved):
lol

PR #905, opened 2022-07-08
feat(component): support margin auto

deini:
## What?
Adds support for margin auto. Can be used in any of our margin props, `margin`, `marginTop`, `marginHorizontal`, etc.

Also in their responsive version `margin={{ mobile: 'auto' }}`

## Why?
Using `margin: auto` is pretty common. Right now you need to either create a custom component, or inline some styles in a utility component like `<Box style={{ marginRight: 'auto' }}>`. Which works fine but doesn't support our responsive api. And can only be applied to utility components.

Screenshot 2022-07-08 at 14 57 33

## Testing/Proof
Unit tests 👀 

Working on non-responsive margin:
Screenshot 2022-07-08 at 14 55 12

Working for responsive margin:
Screenshot 2022-07-08 at 14 54 43

Padding doesn't include `auto`, relevant since both use `spacing` keys for values:
Screenshot 2022-07-08 at 14 55 56


chanceaclark (approved):

chanceaclark:
Might want to add a test, but looks good.

MariaJose (approved):

jorgemoya (approved):
Cool!

PR #913, opened 2022-07-14
TableNext with expandable functionality

MariaJose:
## What?
Create a new `<TableNext/>` component which has all the `Table` component's functionality + expandable functionality. This PR just focus on adding the expandable functionality that works together with Select and Drag and Drop. 

Pagination + Expandable is gonna be in another PR (Plan start working on that today)

If we want to use the expandable functionality we will need to use the `<TableNext/>` component instead, and not the `<Table/>` component. For now we are gonna use this component internally, so we don't plan to add it in the BD documentation until we add tan-stack in the component.

Tan-stack: https://tanstack.com/table/v8

## Why?
We decided to create this new component since we wanted to avoid a breaking change for now because our current Select API in our `<Table/>` component was complicating the things when adding the expandable functionality so it was necessary to change the API. 

**Pros and Cons with this approach:**

**Pros:**
- We can trying out for a bit without breaking current Table usages
- We get a similar api to tan-stack, so its prob easier to port later
- Avoid breaking change (for now)

**Cons:**
- Bigger bundle size since we now have 2 tables
- If we want to add things to the table, we now have multiple tables

**Things to keep in mind:**

- We can reduce the bundle size by not including by default and have a special import import TableNext from '@bigcommerce/big-design/table-next' or something.
- If we make a change to the table, we can temporary add the changes in both tables.
- We need to be "all-in" in terms of getting TableNext be the real Table soon.
- We have a lot of different functionality of our table component that we might need to rethink the future of the component and how we can have it be extensible.
- Prefer we consume the new API of TableNext for a bit, what if we have to do changes and end up with 2 breaking Table changes back 2 back.

The new select API it's gonna be the same as the one that tan-stack uses, since we are planning to add this library in our <TableNext/> component. 

New API for select functionality:
```js
{
0: true
0.0: true
0.1: true
1: true
1.0: true
1.1: true
2: true
2.0: true
2.1: true
}
```
Example:
`0: true` <--- this means that a parent row was selected.
`0.0: true` <--- this means that a parent's child row was selected.

API for expandable functionality:

```js
{0: true, 1: true, 2: true}
```

https://user-images.githubusercontent.com/39739180/179029437-4343c25c-82e3-42a2-aa6c-40dc884b2ad6.mp4

https://user-images.githubusercontent.com/39739180/179029463-11bc738b-eb49-45b6-aedc-a5473e3a32f1.mp4

https://user-images.githubusercontent.com/39739180/179029471-4285e6f0-b695-4eb0-8427-95170ecac084.mp4

## Testing/Proof
Tests pass

### DIFF
--- a/packages/big-design/src/components/TableNext/Actions/Actions.tsx
+++ b/packages/big-design/src/components/TableNext/Actions/Actions.tsx
@@ -5,7 +5,7 @@ import { FlexItem } from '../../Flex';
 import { Text } from '../../Typography';
 import { SelectAll } from '../SelectAll';
 import { TablePagination } from '../TablePagination';
-import { TableItem, TablePaginationProps, TableSelectable } from '../types';
+import { TableExpandable, TableItem, TablePaginationProps, TableSelectable } from '../types';
 
 import { StyledFlex } from './styled';
 
@@ -14,11 +14,13 @@ export interface ActionsProps<T> {
   forwardedRef: RefObject<HTMLDivElement>;
   itemName?: string;
   items: T[];
+  isExpandable: boolean;
   pagination?: TablePaginationProps;
-  onSelectionChange?: TableSelectable<T>['onSelectionChange'];
-  selectedItems: Set<T>;
+  selectedItems: TableSelectable['selectedItems'];
   stickyHeader?: boolean;
   tableId: string;
+  expandedRowSelector?: TableExpandable<T>['expandedRowSelector'];
+  onSelectionChange?: TableSelectable['onSelectionChange'];
 }
 
 const InternalActions = <T extends TableItem>({
@@ -26,11 +28,13 @@ const InternalActions = <T extends TableItem>({
   forwardedRef,
   itemName,
   items = [],
-  onSelectionChange,
   pagination,
   selectedItems,
   stickyHeader,
   tableId,
+  isExpandable,
+  expandedRowSelector,
+  onSelectionChange,
   ...props
 }: ActionsProps<T>) => {
   const isSelectable = typeof onSelectionChange === 'function';
@@ -64,12 +68,16 @@ const InternalActions = <T extends TableItem>({
       stickyHeader={stickyHeader}
       {...props}
     >
-      <SelectAll
-        items={items}
-        onChange={onSelectionChange}
-        selectedItems={selectedItems}
-        totalItems={totalItems}
-      />
+      {isSelectable && (
+        <SelectAll
+          expandedRowSelector={expandedRowSelector}
+          isExpandable={isExpandable}
+          items={items}
+          onChange={onSelectionChange}
+          selectedItems={selectedItems}
+          totalItems={totalItems}
+        />
+      )}
       {renderItemName()}
       {renderActions()}
 
--- a/packages/big-design/src/components/TableNext/Actions/Actions.tsx
+++ b/packages/big-design/src/components/TableNext/Actions/Actions.tsx
@@ -1,0 +1,81 @@
+import React, { RefObject } from 'react';
+
+import { typedMemo } from '../../../utils';
+import { FlexItem } from '../../Flex';
+import { Text } from '../../Typography';
+import { SelectAll } from '../SelectAll';
+import { TablePagination } from '../TablePagination';
+import { TableItem, TablePaginationProps, TableSelectable } from '../types';
+
+import { StyledFlex } from './styled';
+
+export interface ActionsProps<T> {
+  customActions?: React.ReactNode;
+  forwardedRef: RefObject<HTMLDivElement>;
+  itemName?: string;
+  items: T[];
+  pagination?: TablePaginationProps;
+  onSelectionChange?: TableSelectable<T>['onSelectionChange'];
+  selectedItems: Set<T>;
+  stickyHeader?: boolean;
+  tableId: string;
+}
+
+const InternalActions = <T extends TableItem>({
+  customActions,
+  forwardedRef,
+  itemName,
+  items = [],
+  onSelectionChange,
+  pagination,
+  selectedItems,
+  stickyHeader,
+  tableId,
+  ...props
+}: ActionsProps<T>) => {
+  const isSelectable = typeof onSelectionChange === 'function';
+  const totalItems = pagination ? pagination.totalItems : items.length;
+
+  const renderItemName = () => {
+    if (typeof itemName !== 'string') {
+      return null;
+    }
+
+    const text = isSelectable ? itemName : `${totalItems} ${itemName}`;
+
+    return (
+      <FlexItem flexShrink={0} marginRight="medium">
+        <Text margin="none">{text}</Text>
+      </FlexItem>
+    );
+  };
+
+  const renderActions = () => {
+    return customActions ?? null;
+  };
+
+  return (
+    <StyledFlex
+      alignItems="center"
+      aria-controls={tableId}
+      flexDirection="row"
+      justifyContent="stretch"
+      ref={forwardedRef}
+      stickyHeader={stickyHeader}
+      {...props}
+    >
+      <SelectAll
+        items={items}
+        onChange={onSelectionChange}
+        selectedItems={selectedItems}
+        totalItems={totalItems}
+      />
+      {renderItemName()}
+      {renderActions()}
+
+      {pagination && <TablePagination {...pagination} />}
+    </StyledFlex>
+  );
+};
+
+export const Actions = typedMemo(InternalActions);
--- a/packages/big-design/src/components/TableNext/Actions/styled.tsx
+++ b/packages/big-design/src/components/TableNext/Actions/styled.tsx
@@ -1,0 +1,25 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled, { css } from 'styled-components';
+
+import { FlexProps } from '../../Flex';
+import { withFlexedContainer } from '../../Flex/withFlex';
+
+export const StyledFlex = styled.div<FlexProps & { stickyHeader?: boolean }>`
+  ${withFlexedContainer()}
+
+  background-color: ${({ theme }) => theme.colors.white};
+  display: flex;
+  padding: ${({ theme }) => `${theme.spacing.small} ${theme.spacing.xLarge}`};
+
+  ${({ theme, stickyHeader }) =>
+    stickyHeader &&
+    css`
+      ${theme.breakpoints.tablet} {
+        position: sticky;
+        top: 0;
+        z-index: ${theme.zIndex.sticky + 1};
+      }
+    `}
+`;
+
+StyledFlex.defaultProps = { theme: defaultTheme };
--- a/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
+++ b/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
@@ -1,19 +1,21 @@
 import React, { memo, TableHTMLAttributes } from 'react';
 
+import { PaddingProps } from '../../../mixins';
 import { TableColumnDisplayProps } from '../mixins';
 
 import { StyledTableDataCell, StyledTableDataCheckbox } from './styled';
 
 export interface DataCellProps
   extends TableHTMLAttributes<HTMLTableCellElement>,
-    TableColumnDisplayProps {
+    TableColumnDisplayProps,
+    PaddingProps {
   align?: 'left' | 'center' | 'right';
   children?: React.ReactNode;
+  isExpandable?: boolean;
   isCheckbox?: boolean;
   verticalAlign?: 'top' | 'middle';
   width?: number | string;
   withBorder?: boolean;
-  withPadding?: boolean;
 }
 
 export const DataCell: React.FC<DataCellProps> = memo(
@@ -22,15 +24,19 @@ export const DataCell: React.FC<DataCellProps> = memo(
     children,
     display,
     isCheckbox,
+    isExpandable = false,
     verticalAlign,
     width,
     withBorder = true,
-    withPadding = true,
+    paddingHorizontal,
+    paddingVertical,
+    padding,
   }: DataCellProps) => {
     return isCheckbox ? (
       <StyledTableDataCheckbox
         align={align}
         display={display}
+        isExpandable={isExpandable}
         width={width}
         withBorder={withBorder}
       >
@@ -40,10 +46,12 @@ export const DataCell: React.FC<DataCellProps> = memo(
       <StyledTableDataCell
         align={align}
         display={display}
+        padding={padding}
+        paddingHorizontal={paddingHorizontal}
+        paddingVertical={paddingVertical}
         verticalAlign={verticalAlign}
         width={width}
         withBorder={withBorder}
-        withPadding={withPadding}
       >
         {children}
       </StyledTableDataCell>
--- a/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
+++ b/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
@@ -1,0 +1,52 @@
+import React, { memo, TableHTMLAttributes } from 'react';
+
+import { TableColumnDisplayProps } from '../mixins';
+
+import { StyledTableDataCell, StyledTableDataCheckbox } from './styled';
+
+export interface DataCellProps
+  extends TableHTMLAttributes<HTMLTableCellElement>,
+    TableColumnDisplayProps {
+  align?: 'left' | 'center' | 'right';
+  children?: React.ReactNode;
+  isCheckbox?: boolean;
+  verticalAlign?: 'top' | 'middle';
+  width?: number | string;
+  withBorder?: boolean;
+  withPadding?: boolean;
+}
+
+export const DataCell: React.FC<DataCellProps> = memo(
+  ({
+    align,
+    children,
+    display,
+    isCheckbox,
+    verticalAlign,
+    width,
+    withBorder = true,
+    withPadding = true,
+  }: DataCellProps) => {
+    return isCheckbox ? (
+      <StyledTableDataCheckbox
+        align={align}
+        display={display}
+        width={width}
+        withBorder={withBorder}
+      >
+        {children}
+      </StyledTableDataCheckbox>
+    ) : (
+      <StyledTableDataCell
+        align={align}
+        display={display}
+        verticalAlign={verticalAlign}
+        width={width}
+        withBorder={withBorder}
+        withPadding={withPadding}
+      >
+        {children}
+      </StyledTableDataCell>
+    );
+  },
+);
--- a/packages/big-design/src/components/TableNext/DataCell/styled.tsx
+++ b/packages/big-design/src/components/TableNext/DataCell/styled.tsx
@@ -1,25 +1,28 @@
 import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import styled, { css } from 'styled-components';
 
+import { withPaddings } from '../../../mixins';
 import { withTableColumnDisplay } from '../mixins';
 
 import { DataCellProps } from './DataCell';
 
 export const StyledTableDataCell = styled.td<DataCellProps>`
   ${withTableColumnDisplay()}
-
+  ${withPaddings()}
+  
   background-color: ${({ theme }) => theme.colors.white};
   box-sizing: border-box;
   color: ${({ theme }) => theme.colors.secondary70};
   font-size: ${({ theme }) => theme.typography.fontSize.medium};
-  padding: ${({ theme, withPadding }) => (withPadding ? theme.spacing.small : 0)};
 
   &:first-of-type {
-    padding-left: ${({ theme, withPadding }) => (withPadding ? theme.spacing.xLarge : 0)};
+    padding-left: ${({ theme, paddingHorizontal, padding }) =>
+      padding || paddingHorizontal ? theme.spacing.xLarge : 0};
   }
 
   &:last-of-type {
-    padding-right: ${({ theme, withPadding }) => (withPadding ? theme.spacing.xLarge : 0)};
+    padding-right: ${({ theme, paddingHorizontal, padding }) =>
+      padding || paddingHorizontal ? theme.spacing.xLarge : 0};
   }
 
   ${({ theme, withBorder }) =>
@@ -57,9 +60,11 @@ export const StyledTableDataCheckbox = styled(StyledTableDataCell)`
     padding-left: ${({ theme }) => theme.spacing.xLarge};
   }
 
-  &:last-of-type {
-    padding-right: ${({ theme }) => theme.spacing.xLarge};
-  }
+  ${({ isExpandable }) =>
+    isExpandable &&
+    css`
+      padding-right: 0;
+    `}
 
   ${(props) =>
     props.isCheckbox &&
--- a/packages/big-design/src/components/TableNext/DataCell/styled.tsx
+++ b/packages/big-design/src/components/TableNext/DataCell/styled.tsx
@@ -1,0 +1,73 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled, { css } from 'styled-components';
+
+import { withTableColumnDisplay } from '../mixins';
+
+import { DataCellProps } from './DataCell';
+
+export const StyledTableDataCell = styled.td<DataCellProps>`
+  ${withTableColumnDisplay()}
+
+  background-color: ${({ theme }) => theme.colors.white};
+  box-sizing: border-box;
+  color: ${({ theme }) => theme.colors.secondary70};
+  font-size: ${({ theme }) => theme.typography.fontSize.medium};
+  padding: ${({ theme, withPadding }) => (withPadding ? theme.spacing.small : 0)};
+
+  &:first-of-type {
+    padding-left: ${({ theme, withPadding }) => (withPadding ? theme.spacing.xLarge : 0)};
+  }
+
+  &:last-of-type {
+    padding-right: ${({ theme, withPadding }) => (withPadding ? theme.spacing.xLarge : 0)};
+  }
+
+  ${({ theme, withBorder }) =>
+    withBorder &&
+    css`
+      border-bottom: ${theme.border.box};
+    `}
+
+  ${({ align }) =>
+    align &&
+    css`
+      text-align: ${align};
+    `};
+
+  ${({ verticalAlign }) =>
+    verticalAlign &&
+    css`
+      vertical-align: ${verticalAlign};
+    `};
+
+  ${({ width }) =>
+    width !== undefined &&
+    css`
+      width: ${typeof width === 'string' ? width : `${width}px`};
+    `};
+`;
+
+export const StyledTableDataCheckbox = styled(StyledTableDataCell)`
+  ${withTableColumnDisplay()}
+
+  background-color: ${({ theme }) => theme.colors.white};
+  padding: ${({ theme }) => `0 ${theme.spacing.small}`};
+
+  &:first-of-type {
+    padding-left: ${({ theme }) => theme.spacing.xLarge};
+  }
+
+  &:last-of-type {
+    padding-right: ${({ theme }) => theme.spacing.xLarge};
+  }
+
+  ${(props) =>
+    props.isCheckbox &&
+    css`
+      width: ${({ theme }) => theme.helpers.addValues(theme.spacing.xLarge, theme.spacing.small)};
+      white-space: nowrap;
+    `};
+`;
+
+StyledTableDataCell.defaultProps = { theme: defaultTheme };
+StyledTableDataCheckbox.defaultProps = { theme: defaultTheme };
--- a/packages/big-design/src/components/TableNext/HeaderCell/HeaderCell.tsx
+++ b/packages/big-design/src/components/TableNext/HeaderCell/HeaderCell.tsx
@@ -1,0 +1,137 @@
+import {
+  ArrowDownwardIcon,
+  ArrowUpwardIcon,
+  BaselineHelpIcon,
+} from '@bigcommerce/big-design-icons';
+import React, { memo, RefObject, TableHTMLAttributes } from 'react';
+
+import { useComponentSize, useUniqueId } from '../../../hooks';
+import { typedMemo } from '../../../utils';
+import { Box } from '../../Box';
+import { Tooltip } from '../../Tooltip';
+import { TableColumnDisplayProps } from '../mixins';
+import { TableColumn, TableItem } from '../types';
+
+import { StyledFlex, StyledTableHeaderCell, StyledTableHeaderIcon } from './styled';
+
+export interface HeaderCellProps<T>
+  extends TableHTMLAttributes<HTMLTableCellElement>,
+    TableColumnDisplayProps {
+  actionsRef: RefObject<HTMLDivElement>;
+  children?: React.ReactNode;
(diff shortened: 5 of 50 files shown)


packages/big-design/src/components/TableNext/Actions/Actions.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { RefObject } from 'react';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/DataCell/styled.tsx
```diff
@@ -0,0 +1,75 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/DataCell/styled.tsx
```diff
@@ -0,0 +1,75 @@
...
+  &:first-of-type {
+    padding-left: ${({ theme }) => theme.spacing.xLarge};
+  }
+
+  ${({ isExpandable }) =>
+    isExpandable &&
+    css`
+      padding-right: 0;
+    `}
+
```
MariaJose, line 65: Note: There was this styled used in the `<Table/>` component, but I think that is not used 🤔 so I deleted it. Unless I'm missing something.

```css
  &:last-of-type {
    padding-right: ${({ theme }) => theme.spacing.xLarge};
  }
```

packages/big-design/src/components/TableNext/HeaderCell/HeaderCell.tsx
```diff
@@ -0,0 +1,143 @@
+import {
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/Row/Row.tsx
```diff
@@ -0,0 +1,212 @@
+import { ChevronRightIcon, DragIndicatorIcon, ExpandMoreIcon } from '@bigcommerce/big-design-icons';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/Row/useRowState.ts
```diff
@@ -0,0 +1,68 @@
+import { TableSelectable } from '../types';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/RowContainer/RowContainer.tsx
```diff
@@ -0,0 +1,100 @@
+import React, { forwardRef } from 'react';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/SelectAll/SelectAll.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/SelectAll/helpers.ts
```diff
@@ -0,0 +1,115 @@
+import { TableProps, TableSelectable } from '../types';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/SelectAll/useSelectAllState.ts
```diff
@@ -0,0 +1,60 @@
+import { TableProps, TableSelectable } from '../types';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/TableNext.tsx
```diff
@@ -0,0 +1,285 @@
+import React, { memo, useCallback, useRef, useState } from 'react';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/hooks/useExpandable/useExpandable.ts
```diff
@@ -0,0 +1,45 @@
+import { useEffect, useState } from 'react';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/hooks/useSelectable/helpers.ts
```diff
@@ -0,0 +1,84 @@
+import { TableSelectable } from '../../types';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/hooks/useSelectable/useSelectable.ts
```diff
@@ -0,0 +1,73 @@
+import { useEffect, useState } from 'react';
```
MariaJose, line 1: Please review this file

packages/big-design/src/components/TableNext/spec.tsx
```diff
@@ -0,0 +1,849 @@
+import userEvent from '@testing-library/user-event';
```
MariaJose, line 1: Please review this file.

packages/big-design/src/components/TableNext/types.ts
```diff
@@ -0,0 +1,63 @@
+import { ReactNode } from 'react';
```
MariaJose, line 1: Please review this file

packages/big-design/src/hooks/useEventCallback.ts
```diff
@@ -1,16 +1,16 @@
 import { useCallback, useEffect, useRef } from 'react';
```
MariaJose, line 1: Please review this file

MariaJose (commented):
Note:  I added a comment to the files that are necessary to review. You can ignore the ones that don't have since they are just a copy paste form the Table component.

packages/big-design/src/components/TableNext/HeaderCell/HeaderCell.tsx
```diff
@@ -0,0 +1,143 @@
...
+      return (
+        <Tooltip
+          id={tooltipId}
+          placement="right"
+          trigger={
+            <Box as="span" marginLeft="xxSmall">
+              <BaselineHelpIcon
+                aria-describedby={tooltipId}
+                size="medium"
+                title="Hover or focus for additional context."
```
bc-juanvasquez, line 80: `title` need `getText`?

MariaJose: Good call out! I think we don't translate texts in BD for the moment, however, this is one of our future plans. If I recall correctly the Design team created a doc talking about this. So might be good to wait until we decided what's the next step?

packages/big-design/src/components/TableNext/DataCell/styled.tsx
```diff
@@ -0,0 +1,75 @@
...
+`;
+
+export const StyledTableDataCheckbox = styled(StyledTableDataCell)`
+  ${withTableColumnDisplay()}
+
+  background-color: ${({ theme }) => theme.colors.white};
+  padding: ${({ theme }) => `0 ${theme.spacing.small}`};
+
+  &:first-of-type {
+    padding-left: ${({ theme }) => theme.spacing.xLarge};
```
bc-juanvasquez, line 57: was this style being applied as well?

packages/big-design/src/components/TableNext/Row/Row.tsx
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    return null;
+  };
+
+  const renderExpandedIconCell = () => {
+    if (isExpandable && isParentRow && hasChildrenRows) {
+      return (
+        <DataCell
+          cellPadding={isSelectable || isDraggable ? 0 : undefined}
```
bc-juanvasquez, line 125: can make this boolean test a `const`

packages/big-design/src/components/TableNext/HeaderCell/HeaderCell.tsx
rtalvarez: Good callout, but yeah lets not worry too much about gettext for now

rtalvarez (commented):
Since this PR is quite big, I'll devote some focused time with @chanceaclark to get it properly reviewed on Wednesday.

packages/big-design/src/components/TableNext/Row/useRowState.ts
```diff
@@ -0,0 +1,68 @@
...
+      return selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined;
+    });
+
+  const someChildrenRowsSelected: boolean =
+    isExpandable &&
+    childrenRows.some((_childRow, childRowIndex) => {
+      return selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined;
+    });
+
+  const label = isSelected ? `Selected` : `Unselected`;
```
bc-juanvasquez, line 58: `gettext`?

packages/big-design/src/components/TableNext/Row/useRowState.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (onExpandedRow) {
+      onExpandedRow(parentRowIndex ?? null);
+    }
+  };
+
+  const hasChildrenRows = childrenRows.length > 0;
+
+  const allChildrenRowsSelected: boolean =
+    isExpandable &&
+    childrenRows.every((_childRow, childRowIndex) => {
```
bc-juanvasquez, line 48: could create a helper function for this and pass in an argument for the 'filter' function. (`some`, `every`)  since the logic is the same for `allChildrenRowsSelected` and `someChildrenRowsSelected`

packages/big-design/src/components/TableNext/hooks/useSelectable/helpers.ts
bc-juanvasquez: i think if we change `selectedItems` data structure, this logic could be simplified. I'd like to hear what others have to think about that tho 🤔

packages/big-design/src/components/TableNext/Actions/Actions.tsx
```diff
@@ -14,23 +14,28 @@ export interface ActionsProps<T> {
   forwardedRef: RefObject<HTMLDivElement>;
   itemName?: string;
   items: T[];
+  isExpandable: boolean;
+  isSelectable: boolean;
```
MariaJose, line 18: I'll remove this one, I think this isn't necessary

packages/big-design/src/components/TableNext/types.ts
```diff
@@ -44,13 +49,15 @@ export interface TableProps<T> extends React.TableHTMLAttributes<HTMLTableElemen
   actions?: React.ReactNode;
   columns: Array<TableColumn<T>>;
   emptyComponent?: React.ReactElement;
+  expandedRowSelector?: (item: T) => T[] | undefined;
```
MariaJose, line 52: I think should be better to put `expandedRowSelector` inside `TableExpandable` 🤔

packages/big-design/src/components/TableNext/DataCell/styled.tsx
MariaJose: Just re-check, and this one is used 👍🏼

packages/big-design/src/components/TableNext/Row/useRowState.ts
MariaJose: Tried to create the helper function, but I think is now looking too bulky or distracting? 🤔  was there another way to implement it that I might be missing?

```ts
Could be outside the hook: 
const childrenRowsSelected = <T>(
  mode: 'every' | 'some',
  isExpandable: boolean,
  childrenRows: T[],
  parentRowIndex: number,
  selectedItems: TableSelectable['selectedItems'],
) => {
  if (!isExpandable) {
    return false;
  }

  if (mode === 'every') {
    return childrenRows.every((_childRow, childRowIndex) => {
      return selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined;
    });
  }

  return childrenRows.some((_childRow, childRowIndex) => {
    return selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined;
  });
};

or Inside the hook:
  const childrenRowsSelected = <T>(
    mode: 'every' | 'some',
  ) => {
    if (!isExpandable) {
      return false;
    }
  
    if (mode === 'every') {
      return childrenRows.every((_childRow, childRowIndex) => {
        return selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined;
      });
    }
  
    return childrenRows.some((_childRow, childRowIndex) => {
      return selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined;
    });
  };
```

bc-juanvasquez: `const filterChildrenRows = (type) => {
	const dictionary = {
  	'some': Array.some,
    'every': Array.every,
    //extensible to other filters
  }
  const filterFunction = dictionary[type] ?? null;
  const cb = (el, childRowIndex, childrenRows) => selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined
  
  return filterFunction ? filterFunction(cb) : false;
}

const allChildrenRowsSelected = isExpandable && filterChildrenRows('every')`

 was thinking something along these lines, but not sure if it will work 🤔

packages/big-design/src/components/TableNext/Row/styled.tsx
```diff
@@ -20,4 +20,12 @@ export const StyledTableRow = styled.tr<StyledTableRowProps>`
   }
 `;
 
+export const StyledExpandedIcon = styled.button`
```
chanceaclark, line 23: Is there a reason why we are not using our private [`StylableButton`](https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Button/Button.tsx#L50)? That way we can use our `Button` props.

packages/big-design/src/components/TableNext/Row/useRowState.ts
chanceaclark: I'd avoid using a dictionary to point to a prototype method. I'm actually more fond of leaving it as is since it's a bit more readable. If we did want to make it a little better, we could rip it outside of the hook and have it be a bit more generic. But I would only do that if we are going to reuse it in other places 🤷 
```ts
// "All" example
const areAllRowsSelected = (rows, selectedItems) => {
  return rows.some((row, index) => {
      const parentIndex = getParentRowIndex(row); // TODO: Define utility function;

      return selectedItems[`${parentIndex}.${index}`] !== undefined;
  });
}
```

packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
```diff
@@ -9,11 +9,13 @@ export interface DataCellProps
...
   align?: 'left' | 'center' | 'right';
   children?: React.ReactNode;
+  isExpandable?: boolean;
   isCheckbox?: boolean;
   verticalAlign?: 'top' | 'middle';
   width?: number | string;
   withBorder?: boolean;
-  withPadding?: boolean;
+  withVerticalPadding?: boolean;
+  withHorizontalPadding?: boolean;
```
chanceaclark, line 18: Instead of defining new props, why don't we try and extend off of our `PaddingProps`? May not be possible but worth a shot.

packages/big-design/src/components/TableNext/Row/Row.tsx
```diff
@@ -1,60 +1,165 @@
...
-
-interface PrivateProps {
-  forwardedRef?: React.Ref<HTMLTableRowElement>;
+  onExpandedRow?(parentRowIndex: number | null): void;
+  onItemSelect?(
+    childRowIndex: number | null,
+    index: number,
+    childrenRows: T[],
+    isParentRow: boolean,
+    isExpandable: boolean,
```
chanceaclark, line 39: Is there a way where we can reduce the amount of arguments passed? Anything greater than 2-3 arguments can be quite cumbersome. Could this be a single interface a pass in instead (like it was before)?

packages/big-design/src/components/TableNext/Row/Row.tsx
```diff
@@ -1,60 +1,165 @@
...
+
+  const renderExtraCellsForParentRow = () => {
+    const extraDataCells: ReactNode[] = [];
+
+    if (!hasChildrenRows) {
+      extraDataCells.push(<DataCell key={`parent-extra-cell-${parentRowIndex}-1`} />);
+    }
+
+    return extraDataCells;
+  };
```
chanceaclark, line 140: We could simplify this a bit instead of relying on array mutations
```suggestion
  const renderExtraCellsForParentRow = (): ReactNode[] => {
    if (!hasChildrenRows) {
      return [<DataCell key={`parent-extra-cell-${parentRowIndex}-1`} />];
    }

    return [];
  };
```

packages/big-design/src/components/TableNext/hooks/useSelectable/helpers.ts
```diff
@@ -0,0 +1,84 @@
...
+    return;
+  }
+
+  const isSelectedChild = selectedItems[`${parentRowIndex}.${childRowIndex}`] !== undefined;
+
+  if (isSelectedChild) {
+    delete selectedItems[`${parentRowIndex}.${childRowIndex}`];
+
+    if (isTheOnlySelectedChildRow) {
+      delete selectedItems[`${parentRowIndex}`];
```
chanceaclark, line 61: 😬 Can we try and find a better way for this instead of mutating the array? Mutations are side-effects and are hard to track down if there are any issues with it. It's better to return an immutable array instead.

packages/big-design/src/components/TableNext/hooks/useExpandable/useExpandable.ts
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+
+    const { onExpandedChange } = expandable;
+
+    const isExpandedRow = expandedRows[parentRowIndex] !== undefined;
+    const newExpandedRows = { ...expandedRows };
+
+    if (isExpandedRow) {
+      delete newExpandedRows[parentRowIndex];
```
chanceaclark, line 21: Same here.

packages/big-design/src/components/TableNext/SelectAll/helpers.ts
```diff
@@ -0,0 +1,115 @@
...
+  return expandedRowSelector(parentRow) ?? [];
+}
+
+export function areAllInPageSelected<T>(
+  isExpandable: boolean,
+  items: T[],
+  selectedItems: TableSelectable['selectedItems'],
+  expandedRowSelector?: TableExpandable<T>['expandedRowSelector'],
+) {
+  return (
```
rtalvarez, line 35: Hmm I think these `return`s in this file could be tweaked a bit. ATM I think it's hard to follow them because theres a `return` -> `items.every` -> more `return` branches -> nested `childrenRows.every` -> `return` branches

Maybe pulling this into logical helper functions? Or moving some stuff around 🤔

packages/big-design/src/components/TableNext/SelectAll/useSelectAllState.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    if (allInPageSelected) {
+      selectAll(isExpandable, items, newSelectedItems, expandedRowSelector);
+
+      return onChange(newSelectedItems);
+    }
+
+    unselectAll(isExpandable, items, newSelectedItems, expandedRowSelector);
+
+    return onChange(newSelectedItems);
```
jorgemoya, line 58: Why does it look like the `selectAll` and `unselectAll` are reversed 😆 . I'm sure it's working fine, just having hard time in understanding why we wouldn't `unselectAll` if `allInPageSelected`?

packages/big-design/src/components/TableNext/mixins/display/display.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const getSimpleDisplay = (
+  displayProp: string | number,
+  cssKey: string,
+): FlattenSimpleInterpolation => css`
+  ${cssKey}: ${displayProp}
+`;
+
+const getResponsiveDisplay: TableColumnDisplayOverload = (
```
jorgemoya, line 33: Wondering why we are not using the mixins that already exist in `src/mixins/display`?

jorgemoya (commented):
Looking good!

packages/big-design/src/components/TableNext/Actions/styled.tsx
```diff
@@ -0,0 +1,25 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled, { css } from 'styled-components';
+
+import { FlexProps } from '../../Flex';
+import { withFlexedContainer } from '../../Flex/withFlex';
+
+export const StyledFlex = styled.div<FlexProps & { stickyHeader?: boolean }>`
+  ${withFlexedContainer()}
```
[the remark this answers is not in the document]
chanceaclark: I guess it really depends on the situation. If we extend via `styled(Flex)` we introduce a lot more props than just the `FlexProps`; this would introduce `BoxProps`, `PaddingProps`, `MarginProps`, & `DisplayProps`. If we do it the `styled.div<FlexProps>` way we can narrow the props to what we want without having to filter out the unwanted props.

packages/big-design/src/components/TableNext/Row/styled.tsx
MariaJose: Tried with the public one, but didn't know we could use that one ^  😓 , so decided to create a new one because I was having issues with the styles, but good to know that this one is more flexible.

Also the `StylableButton` adds styles when button is active, focused,  and hover. I guess I'm gonna leave these styles for now to have better accessibility even-though that is not specified in the designs, I think we could change it later in case this is not required.

Screen Shot 2022-07-21 at 1 33 19 PM

packages/big-design/src/components/TableNext/hooks/useSelectable/helpers.ts
MariaJose: I think I'm not mutating the object. Before calling this helper fn I'm creating a copy and passed the copy to the helper function. Reference: packages/big-design/src/components/TableNext/hooks/useSelectable/useSelectable.ts

I think it might be confused because I naming this prop as selectedItems, probably I should name it differently, or make the copy inside the this helper function to prevent confusion.

packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
MariaJose: Tried to extended it, but things got messed up with styles, I'll leave this suggestion until the end.

packages/big-design/src/components/TableNext/hooks/useSelectable/helpers.ts
chanceaclark: You are still mutating a passed object and mutating it's contents. Could we try and make this function a bit smaller, reduce the amount of arguments and have it do one thing: take an object and return a new one. Smaller functions are highly preferred.

```ts
const selectChildRow = (selectedItems: TableSelectable['selectedItems'], childRowIndex: number, ): TableSelectable['selectedItems'] => {
  const parentRowIndex = getParentIndex(childRowIndex);
  const isSelectedParent = selectedItems[parentRowIndex] !== undefined;

  if (!isSelectedParent) {
    selectedItems[`${parentRowIndex}`] = true;
    selectedItems[`${parentRowIndex}.${childRowIndex}`] = true;

    return {
      ...selectedItems,
      [`${parentRowIndex}`]: true,
      [`${parentRowIndex}.${childRowIndex}`] = true;
    };
  }

  // ...
}

// Usage:

const newSelectedItems = selectChildRow(selectedItems, childRowIndex);

onSelectionChange(newSelectedItems);
```

Instead of deleting keys could we just update the value to false?

deini: This example is still mutating, left-over copy paste?

packages/big-design/src/components/TableNext/SelectAll/useSelectAllState.ts
MariaJose: lol, good catch, thanks! 😓  I swapped the names

packages/big-design/src/components/TableNext/mixins/display/types.ts
```diff
@@ -0,0 +1,16 @@
+import { ThemeInterface } from '@bigcommerce/big-design-theme';
+import { FlattenSimpleInterpolation } from 'styled-components';
+
+import { ResponsiveProp } from '../../../../types';
+
+type TableColumnDisplayProp = ResponsiveProp<'table-cell' | 'none'>;
+
+export type TableColumnDisplayProps = Partial<{
+  display: TableColumnDisplayProp;
+}>;
```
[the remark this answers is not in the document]
MariaJose: I agree, we could use `?` instead of `Partial`. I just gonna change `type` to `interface` since it's causing eslint errors `Use an `interface` instead of a `type`

would be like this:
```
export interface TableColumnDisplayProps {
  display?: TableColumnDisplayProp;
}
```

packages/big-design/src/components/TableNext/mixins/display/display.tsx
MariaJose: Not sure tbh, this was a copy paste from the Table component, it was like that before. But I'll take a look and see what I can do

packages/big-design/src/components/TableNext/SelectAll/helpers.ts
MariaJose: That's true, I created a helper function so it can be easier to follow 👍🏼

packages/big-design/src/components/TableNext/mixins/display/display.tsx
MariaJose: I already checked and I think is like that because the types are different, so I imagine that even-though it's almost the same, it was better to keep it separately to make it more readable ? 🤔  I think I'm gonna leave it like that and we can create another ticket to handle that scope.

MariaJose:
♻️  @bigcommerce/channels-fe

jorgemoya (approved):
lgtm!

packages/big-design/src/components/TableNext/SelectAll/useSelectAllState.ts
```diff
@@ -0,0 +1,69 @@
+import { TableExpandable, TableSelectable } from '../types';
+
+import {
+  areAllInPageSelected,
+  areSomeInPageSelected,
+  getTotalSelectedItems,
+  selectAll,
+  // unselectAll,
```
chanceaclark, line 8: Unused?

packages/big-design/src/components/TableNext/hooks/useSelectable/useSelectable.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+  );
+
+  useEffect(() => {
+    if (selectable) {
+      const newSelectedItems = { ...selectable.selectedItems };
+
+      setSelectedItems(newSelectedItems);
+    }
+  }, [selectable, selectableConditionalDep]);
```
chanceaclark, line 76: `const selectableConditionalDep = selectable ? selectable.selectedItems : null;` is generally not ideal. We should be able to use optional chaining instead.
```suggestion
  useEffect(() => {
    if (selectable?.selectedItems) {
      setSelectedItems({ ...selectable.selectedItems });
    }
  }, [selectable?.selectedItems]);
```

packages/big-design/src/components/TableNext/hooks/useSelectable/helpers.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+export function getTotalSelectedChildRows<T>({
+  childrenRows,
+  parentRowIndex,
+  selectedItems,
+}: SelectRowArg<T>) {
+  return (
+    childrenRows !== undefined &&
+    childrenRows.reduce((acc, _childRow, childRowIndex) => {
```
chanceaclark, line 201: Can we use optional chaining instead `childrenRows?.reduce`?

packages/big-design/src/components/TableNext/hooks/useSelectable/helpers.ts
```diff
@@ -0,0 +1,209 @@
...
+}: SelectRowArg<T>) {
+  const newSelectedItems = { ...selectedItems };
+
+  delete newSelectedItems[`${parentRowIndex}.${childRowIndex}`];
+
+  if (isTheOnlySelectedChildRow) {
+    delete newSelectedItems[`${parentRowIndex}`];
+  }
+
+  return newSelectedItems;
```
chanceaclark, line 191: Again, we probably shouldn't mutate even if we are creating a new object/array. Maybe we can destructure where we delete instead?
```ts
const parentChildIndex = `${parentRowIndex}.${childRowIndex}`;
const { [parentChildIndex]: _child, ...removedChildSelectedItems } = selectedItems;

if (isTheOnlySelectedChildRow) {
  const parentIndex = `${parentRowIndex}`;
  const { [parentIndex]: _parent, ...removedParentSelectedItems } = removedChildSelectedItems;
   
  return removedParentSelectedItems;
}

return removedChildSelectedItems;
```

MariaJose:
♻️

MariaJose:
♻️  @bigcommerce/channels-fe

packages/big-design/src/components/TableNext/SelectAll/helpers.ts
```diff
@@ -0,0 +1,127 @@
...
+      childrenRows.forEach((_childRow, childRowIndex) => {
+        if (newSelectedItems[`${parentRowIndex}.${childRowIndex}`] === undefined) {
+          newSelectedItems[`${parentRowIndex}.${childRowIndex}`] = true;
+        }
+      });
+    }
+  });
+
+  return newSelectedItems;
+}
```
chanceaclark, line 127: Do we need to know about previous selected items or can we just build a new object from scratch? This may not be 100% right, but this was what I was thinking:
```suggestion
export function selectAll<T>({
  expandedRowSelector,
  isExpandable,
  items,
}: SelectAllRowsArg<T>): TableSelectable['selectedItems'] {
  return Object.fromEntries(items.map((parent, parentIndex) => {
    const children = getChildrenRows(parent, expandedRowSelector);

    const selectedChildren = isExpandable ? children.map((_child, childIndex) => {
        return [`${parentIndex}.${childIndex}`, true];
    }) : [];

    return [
        [`${parentIndex}`, true],
        ...selectedChildren,
    ];
  }));
}
```

chanceaclark (commented):
Few more comments, otherwise looking really good.

packages/big-design/src/components/TableNext/RowContainer/RowContainer.tsx
```diff
@@ -0,0 +1,100 @@
...
+  expandedRowSelector,
+  getItemKey,
+  onItemSelect,
+  onExpandedRow,
+  selectedItems,
+  ...rest
+}: InternalRowContainerProps<T> & PrivateProps) => {
+  const isParentRowSelected = selectedItems[parentRowIndex] !== undefined;
+  const isExpanded = expandedRows[parentRowIndex] !== undefined;
+  const childrenRows: T[] | undefined = expandedRowSelector ? expandedRowSelector(item) : [];
```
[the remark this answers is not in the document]
MariaJose: That's true, looks better 👍🏼

MariaJose:
♻️

jorgemoya (commented):
lgtm!

PR #920, opened 2022-07-20
TableNext with helperRow component

MariaJose:
## What?

Add `renderHelperRow` prop. This prop accepts a functional component or class component. This component is gonna help the user render children rows dynamically and showing loader when doing a request. Also, this component must receive a `parentRowIndex` prop which is gonna be helpful to request additional children rows by parent.

Also, another change that was made was inside the expanded props, specifically on the `onExpandedChange` fn. This function is gonna receives two arguments. `expandedItems` and `parentRowIndex`. The last one is gonna help the user fetch data dynamically after clicking on the expanded icon.

Example: 

```tsx

    const HelperRow = ({ parentRowIndex }) => {
        if (state[parentIndex].isLoading) {
           return <LoadingIcon/>
        }

        if (state[parentIndex] !== undefined && state[parentIndex].remaining) {
            const itemsToView = state[parentIndex].remaining;

            return <Button onClick={() => viewMore(parentRowIndex)} />
        }

        return null
    };

    <TableNext
      columns={columns}
      expandable={{
        expandedRows,
        onExpandedChange: (newExpandedRows: Record<string, true>, parentRowIndex: number) => {
          setExpandedRows(newExpandedRows);

          if (newExpandedRows[parentRowIndex]) {
            handleOnClickFirstFetch(parentRowIndex);
          }
        },
        expandedRowSelector: ({ subRows }) => subRows,
        helperRowRenderer: RenderRow,
      }}
      itemName="Products"
      items={items}
      keyField="sku"
      selectable={{ selectedItems, onSelectionChange: setSelectedItems }}
    />
  );
```

## Why?

These changes were made in order to fetch data dynamically.

https://user-images.githubusercontent.com/39739180/180834073-9ffbcee4-3a99-439f-8b85-fbdb13e2dfd2.mp4

## Testing/Proof
Test pass locally

### DIFF
--- a/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
+++ b/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
@@ -1,4 +1,4 @@
-import React, { memo, TableHTMLAttributes } from 'react';
+import React, { memo, TdHTMLAttributes } from 'react';
 
 import { PaddingProps } from '../../../mixins';
 import { TableColumnDisplayProps } from '../mixins';
@@ -6,13 +6,12 @@ import { TableColumnDisplayProps } from '../mixins';
 import { StyledTableDataCell, StyledTableDataCheckbox } from './styled';
 
 export interface DataCellProps
-  extends TableHTMLAttributes<HTMLTableCellElement>,
+  extends TdHTMLAttributes<HTMLTableCellElement>,
     TableColumnDisplayProps,
     PaddingProps {
   align?: 'left' | 'center' | 'right';
   checkEmptyCell?: boolean;
   children?: React.ReactNode;
-  colSpan?: number;
   isExpandable?: boolean;
   isCheckbox?: boolean;
   verticalAlign?: 'top' | 'middle';
--- a/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
+++ b/packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
@@ -10,7 +10,9 @@ export interface DataCellProps
     TableColumnDisplayProps,
     PaddingProps {
   align?: 'left' | 'center' | 'right';
+  checkEmptyCell?: boolean;
   children?: React.ReactNode;
+  colSpan?: number;
   isExpandable?: boolean;
   isCheckbox?: boolean;
   verticalAlign?: 'top' | 'middle';
@@ -21,7 +23,9 @@ export interface DataCellProps
 export const DataCell: React.FC<DataCellProps> = memo(
   ({
     align,
+    checkEmptyCell,
     children,
+    colSpan,
     display,
     isCheckbox,
     isExpandable = false,
@@ -45,6 +49,8 @@ export const DataCell: React.FC<DataCellProps> = memo(
     ) : (
       <StyledTableDataCell
         align={align}
+        checkEmptyCell={checkEmptyCell}
+        colSpan={colSpan}
         display={display}
         padding={padding}
         paddingHorizontal={paddingHorizontal}
--- a/packages/big-design/src/components/TableNext/DataCell/styled.tsx
+++ b/packages/big-design/src/components/TableNext/DataCell/styled.tsx
@@ -43,6 +43,17 @@ export const StyledTableDataCell = styled.td<DataCellProps>`
       vertical-align: ${verticalAlign};
     `};
 
+  ${({ checkEmptyCell }) => {
+    return (
+      checkEmptyCell &&
+      css`
+        &:empty {
+          display: none;
+        }
+      `
+    );
+  }};
+
   ${({ width }) =>
     width !== undefined &&
     css`
--- a/packages/big-design/src/components/TableNext/Row/useRowState.ts
+++ b/packages/big-design/src/components/TableNext/Row/useRowState.ts
@@ -3,7 +3,7 @@ import { TableSelectable } from '../types';
 
 interface UseRowStateProps<T> {
   childRowIndex?: number;
-  childrenRows: T[];
+  childrenRows?: T[];
   isExpandable: boolean;
   isParentRow: boolean;
   isSelected?: boolean;
@@ -28,7 +28,7 @@ export const useRowState = <T>({
     if (onItemSelect) {
       onItemSelect({
         childRowIndex: childRowIndex ?? null,
-        childrenRows,
+        childrenRows: childrenRows ?? [],
         isParentRow,
         isExpandable,
         parentRowIndex,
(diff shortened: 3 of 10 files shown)


MariaJose:
Note: This PR has changes from  #930 Pr. To review this PR check the last commit to see the changes ([add helper row component](https://github.com/bigcommerce/big-design/pull/920/commits/2033d0245cb727032758e356d7ff811160c5b1ab)).

packages/big-design/src/components/TableNext/Row/useRowState.ts
```diff
@@ -42,23 +42,28 @@ export const useRowState = <T>({
     }
   };
 
-  const hasChildrenRows = childrenRows.length > 0;
+  const hasChildrenRows = childrenRows !== undefined;
```
bc-juanvasquez, line 45: why not `const hasChildrenRows = totalChildrenRows > 0`

packages/big-design/src/components/TableNext/types.ts
```diff
@@ -12,8 +12,11 @@ export interface TableSelectable {
 
 export interface TableExpandable<T> {
   expandedRows: Record<string, true>;
-  onExpandedChange(expandedItems: Record<string, true>): void;
   expandedRowSelector: (item: T) => T[] | undefined;
+  onExpandedChange(expandedItems: Record<string, true>, expandedIndex: number): void;
+  renderHelperRow?:
+    | ComponentType<{ parentRowIndex: number; children?: ReactNode }>
```
bc-juanvasquez, line 18: are we planning to have different implementations of the helper row because i'm not sure why the type is a renderer. Also, should probably rename it to helperRowRenderer because at first thought, it seems like a boolean property

packages/big-design/src/components/TableNext/Row/useRowState.ts
MariaJose: Good question, this helps to identify when rendering the data dynamically. There could be two scenarios: A row that might have children but at the beginning they are empty because is dynamic, and a row that does not have children at all.  👀  However, I'll recheck again and see how I can better name this ^ 🤔

packages/big-design/src/components/TableNext/types.ts
MariaJose: Good point, I think it's more descriptive 👍🏼 . I'll update it

packages/big-design/src/components/TableNext/RowContainer/RowContainer.tsx
```diff
@@ -89,6 +93,15 @@ const InternalRowContainer = <T extends TableItem>({
             />
           );
         })}
+      {isExpanded && childrenRows !== undefined && hasHelperRowComponent && (
+        <tr key={`extra-helper-row-${parentRowIndex}`}>
+          <DataCell checkEmptyCell={true} colSpan={10000} width={100}>
```
chanceaclark, line 98: Few notes here:
1. Can we calculate the `colSpan` instead of using a magic number?
2. What does `width={100}` represent (another magic number)?
```suggestion
          <DataCell checkEmptyCell={true} colSpan={columns.length + <isSelectable> + <isDragable>} width={REPRESENTATION_OF_WIDTH}>
```

packages/big-design/src/components/TableNext/DataCell/styled.tsx
```diff
@@ -43,6 +43,17 @@ export const StyledTableDataCell = styled.td<DataCellProps>`
...
+  ${({ checkEmptyCell }) => {
+    return (
+      checkEmptyCell &&
+      css`
+        &:empty {
+          display: none;
+        }
+      `
+    );
+  }};
```
chanceaclark, line 55: 🍹 I'm not very familiar with this. Could you explain the need for this?

packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
```diff
@@ -10,7 +10,9 @@ export interface DataCellProps
     TableColumnDisplayProps,
     PaddingProps {
   align?: 'left' | 'center' | 'right';
+  checkEmptyCell?: boolean;
   children?: React.ReactNode;
+  colSpan?: number;
```
chanceaclark, line 15: 🍹 I believe this should already be apart of `TableHTMLAttributes<HTMLTableCellElement>` 🤔

jorgemoya (approved):
lgtm!

packages/big-design/src/components/TableNext/types.ts
MariaJose: ohh this is also gonna help render the button, or the loading icon based on the status of each parent row. 

Example: 
```tsx
  const HelperRow = ({ parentRowIndex }) => {
  
    if (state[parentIndex].isLoading) {
       return <LoadingIcon/>
    }

   if (state[parentIndex] !== undefined && state[parentIndex].remaining) {
    const itemsToView = state[parentIndex].remaining;
     return <Button onClick={viewMore(parentRowIndex)} />
   }

   return null
  };
```

packages/big-design/src/components/TableNext/RowContainer/RowContainer.tsx
MariaJose: 1. oops forgot to remove the width 😅, good catch! Was trying to figure it out how to make the cell bigger.  Thanks!
2. I set a big number since didn't want to complicate too much the logic of calculating if columns + extra cell. I'll try to calculated it, and see how is gonna look 👍🏼  Thanks!

packages/big-design/src/components/TableNext/DataCell/styled.tsx
MariaJose: Sure, so if the HelperRowComponent returns null, then the cell is gonna be empty. So in order to disappear the cell, I had to set this css logic. 

```tsx
 const HelperRow = ({ parentRowIndex }) => {
  
    if (state[parentIndex].isLoading) {
       return <LoadingIcon/>
    }

   if (state[parentIndex] !== undefined && state[parentIndex].remaining) {

     const itemsToView = state[parentIndex].remaining;

     return <Button onClick={() => viewMore(parentRowIndex)} />
   }

   return null
  };
```

packages/big-design/src/components/TableNext/DataCell/DataCell.tsx
MariaJose: Ohh yes, I remember ts was complaining, let me try to check it again 🤔

bc-juanvasquez:
This is more of a UX question, but curious if we want to still render the helper row when all children are expanded? That way we can collapse expanded if we wanted to, or would collapsing the parent provide this functionality? Think we should provide a way for the user to undo the 'more' children expansion if they wanted?

MariaJose:
> This is more of a UX question, but curious if we want to still render the helper row when all children are expanded? That way we can collapse expanded if we wanted to, or would collapsing the parent provide this functionality? Think we should provide a way for the user to undo the 'more' children expansion if they wanted?

I think this can be possible with the `helperRowRenderer` prop. The user can pass a component and this component is gonna be the one who can decide what to do (displaying a button, loading icon, collapse children, etc).

PR #921, opened 2022-07-21
fix(component): fix worksheet's onError triggering

bc-apostoliuk:
## What? Why?
1. Remove the checking of the condition `invalidCells.length > 0` before triggering `onError`.
2. Add the fix to avoid extra triggering of the `onError` when passing the new items to the `Worksheet` from the outside

This is the same issue as it was with `onChange`: [PR](https://github.com/bigcommerce/big-design/pull/896)

## Testing/Proof

Locally.


jorgemoya (approved):
Great catch!

MariaJose (approved):

bc-juanvasquez (approved):

jorgemoya:
@bc-apostoliuk Do you plan to fix the tests for this? Or figure out why it errors?

PR #927, opened 2022-08-01
feat(component): TableNext - add loadAction prop

MariaJose:
## What?

Add `loadMoreAction` prop in `TableNext` component. This prop accepts an object with the following keys: `isLoading`, `onClick`, `text`, and `showLoadMore`.  This new props is gonna help the user render children rows dynamically and showing loader when doing a request. 

Example how to use this prop:

```tsx
 loadMoreAction:  (parentRowIndex) => {
     // we can identify when not rendering the cell with the button.
     if (//condition goes here) {
        return undefined;
     }
       
       
     return {
        text: displayText(parentRowIndex),
        isLoading: isLoading(parentRowIndex),
        onClick: (event, parentRowIndex) => {
            handleOnClickViewMore(event, parentRowIndex);
        }     
     }
 }
```

## Why?
`helperRowRenderer` prop is not gonna be used anymore, we are gonna use this new prop instead.

https://user-images.githubusercontent.com/39739180/182249517-2a54c28f-2470-47e4-9b5b-9b1e861bb17d.mp4

## Testing/Proof

Locally tested


chanceaclark (approved):
Way better! It's still a little weird how every prop is a function, but would love to know what other people think?

jorgemoya:
> Way better! It's still a little weird how every prop is a function, but would love to know what other people think?

I can see how it can be weird. Maybe allow string on `text` and boolean on `isLoading` and `showMore`?

MariaJose:
> > Way better! It's still a little weird how every prop is a function, but would love to know what other people think?
> 
> I can see how it can be weird. Maybe allow string on `text` and boolean on `isLoading` and `showMore`?

@jorgemoya did you mean something like this?
```
 isLoading: (parentRowIndex: number) => boolean | boolean;
```

bc-juanvasquez:
Either approach seems fine, but I do think we can make the prop name more descriptive such as `loadMoreActionCallbacks` or something similar.  I might prefer the 'function' callback approach because you can short circuit if need be.

bc-juanvasquez (approved):

MariaJose:
> Either approach seems fine, but I do think we can make the prop name more descriptive such as `loadMoreActionCallbacks` or something similar. I might prefer the 'function' callback approach because you can short circuit if need be.

 I think I'm gonna change to the single fn  👍🏼

MariaJose:
♻️

packages/big-design/src/components/TableNext/spec.tsx
```diff
@@ -879,10 +879,8 @@ describe('expandable', () => {
     expect(onExpandedChange).not.toHaveBeenCalledWith({ 0: true, 1: true, 2: true });
   });
 
-  test('renders a cell with a helper row component in a expanded parent', async () => {
-    const HelperRow: FC<{ parentRowIndex: number }> = ({ parentRowIndex }) => {
-      return <button>View more {parentRowIndex}</button>;
-    };
+  test('when loadMoreAction exists a cell with a button is rendered', async () => {
```
chanceaclark, line 882: 🍹 I think we've been trying to use `it` vs. `test`
```suggestion
  it('renders a "View more" button on an expandable row', async () => {
```

chanceaclark (approved):
A lot better!

jorgemoya (approved):

packages/big-design/src/components/TableNext/spec.tsx
MariaJose: I think I'm gonna keep for now `test` since this is what we are using for now in BD 😅

bc-juanvasquez (approved):

PR #933, opened 2022-08-02
feat(component): add AccordionPanel component

bc-athorne:
## What?

Add `AccordionPanel` and `Accordion` components to BigDesign

## Why?

[CHI-1984](https://bigcommercecloud.atlassian.net/browse/CHI-1984) - This component is needed for current Feedonomics work by CMX team. 

### Component
https://user-images.githubusercontent.com/95502352/183974605-5148f708-a8a3-4462-8458-b54a748ae4e1.mov

### Docs
Screen Shot 2022-08-02 at 1 49 15 PM
Screen Shot 2022-08-16 at 2 15 47 PM
Screen Shot 2022-08-16 at 2 15 56 PM
Screen Shot 2022-08-16 at 2 16 02 PM
Screen Shot 2022-08-16 at 2 16 28 PM
Screen Shot 2022-08-10 at 12 10 14 PM

## Testing/Proof
All unit tests are passing.


bc-athorne:
I'm not sure why this test is failing in CircleCI. It logs the correct value and passes in my local environment.

Screen Shot 2022-08-02 at 3 37 37 PM

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+            </Text>
+          </StyledGridItem>
+          <StyledGridItem>
+            {isExpanded ? (
+              <ExpandLessIcon color="secondary70" />
+            ) : (
+              <ExpandMoreIcon color="secondary70" />
+            )}
+          </StyledGridItem>
+        </StyledAccordionButton>
```
MariaJose, line 48: should we create a component that renders this logic ^ and then use this one in here `AccordionButton` ? 🍹

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+          {...props}
+          aria-labelledby={accordionId}
+          display={isExpanded ? 'block' : 'none'}
+          hidden={!isExpanded}
+          iconLeft={iconLeft}
+          id={accordionItemId}
+          role="region"
+        >
+          {children}
+        </StyledBox>
```
MariaJose, line 59: Should we name this one as `AccordionContent`? 🍹

packages/big-design/src/components/AccordionGroup/hooks/useAccordion.tsx
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from 'react';
+
+interface UseAccordionProps {
+  defaultExpanded?: boolean;
+}
+
+export const useAccordion = ({ defaultExpanded }: UseAccordionProps) => {
+  const [isExpanded, setisExpanded] = useState(false);
```
MariaJose, line 8: `setIsExpanded` ? 🍹

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export const AccordionPanel: React.FC<AccordionPanelProps & BoxProps> = memo(
+  ({ children, header, iconLeft, defaultExpanded = false, ...props }) => {
+    const { isExpanded, onClick } = useAccordion({ defaultExpanded });
+    const accordionId = useUniqueId('accordion');
+    const accordionItemId = useUniqueId('accordion-item');
+
+    return (
+      <StyledAccordion {...props} isExpanded={isExpanded}>
```
MariaJose, line 27: Just curious, not sure the requirements of the accordion component in terms of usability, but do we want this component to let the user style the accordion or should we create a private accordion so it can prevent that?

packages/big-design/src/components/AccordionGroup/hooks/useAccordion.tsx
```diff
@@ -0,0 +1,26 @@
...
+  useEffect(() => {
+    if (defaultExpanded) {
+      setisExpanded(defaultExpanded);
+    }
+  }, [defaultExpanded]);
+
+  const onClick = () => {
+    const nextIsExpanded = !isExpanded;
+
+    setisExpanded(nextIsExpanded);
```
MariaJose, line 19: `setIsExpanded(!nextIsExpanded)` 🍹 ?

MariaJose (approved):
Lgtm 👏🏼 ! Just few comments

MariaJose:
> I'm not sure why this test is failing in CircleCI. It logs the correct value and passes in my local environment.
> 
> Screen Shot 2022-08-02 at 3 37 37 PM
> 
> Local env:
> 
> ```
> test('if defaultExpanded is set to true, accordion is expanded', async () => {
>   render(
>     <AccordionPanel defaultExpanded={true} header="Panel Header" isExpanded onClick={handleClick}>
>       <Text>This is a child component</Text>
>     </AccordionPanel>,
>   );
> 
>   const panel = await screen.findByRole('button');
> 
>   console.log(panel.getAttribute('aria-expanded'));
> 
>   expect(panel.getAttribute('aria-expanded')).toBe('true');
> });
> ```
> 
> Screen Shot 2022-08-02 at 3 39 37 PM

> It failed for me locally as well :eyes. But I add this and the test passed :

```ts
const [isExpanded, setisExpanded] = useState(defaultExpanded ?? false);
```

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
bc-athorne: Could you elaborate on what you mean by creating a private accordion?

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+          variant="subtle"
+        >
+          <StyledGridItem>
+            {iconLeft}
+            <Text as="span" marginLeft={iconLeft ? 'small' : 'none'}>
+              {header}
+            </Text>
+          </StyledGridItem>
+          <StyledGridItem>
+            {isExpanded ? (
```
chanceaclark, line 42: Maybe we can try and use the button props instead of trying to manually layout the icons?

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
MariaJose: Sure, I've seen some components in BD that they are private or I think they are called raw 😅 . This component is rendered by other component(kind of like a public one), and we can decided what props to pass to the raw component.

Example Button component renders a raw button without style props: https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Button/Button.tsx#L54

packages/big-design/src/components/AccordionGroup/AccordionGroup.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export interface AccordionGroupProps {
+  header: string;
+  children?: React.ReactNode;
+}
+
+export const AccordionGroup: React.FC<AccordionGroupProps> = memo(({ header, children }) => {
+  return (
+    <StyledAccordionGroup>
+      <StyledH2>{header}</StyledH2>
```
chanceaclark, line 13: Why don't we just use a `Panel`?

chanceaclark (changes requested):
Hey @bc-athorne! Congrats opening up your first contribution to BigDesign! I do have a few comments we should address, including some API updates we should make. Let me know if you have any questions!

_FYI I'm just requesting changes to make sure we address some of the API concerns before we merge._

packages/big-design/src/components/AccordionGroup/spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  await waitFor(() => {
+    expect(panel.getAttribute('aria-expanded')).toBe('true');
+  });
+
+  userEvent.click(panel);
+
+  await waitFor(() => {
+    expect(panel.getAttribute('aria-expanded')).toBe('false');
+  });
```
chanceaclark, line 68: `panel` seem to refer to the button 🤔 maybe rename that?

chanceaclark: Also, to fix the issues you were seeing in CI, we should rewrite these tests to be a bit more semantic instead of using `waitFor`...
```suggestion
  const button = await screen.findByRole('button');

  userEvent.click(button);
  
  const panel = await screen.findByRole('region');
  
  expect(panel).toBeVisible();

  userEvent.click(button);

  expect(panel).not.toBeVisible();
```

Might have to tinker with some `findByRole` options, but I believe that should be a bit more meaningful. You should probably update the other tests as well.

packages/big-design/src/components/AccordionGroup/hooks/useAccordion.tsx
MariaJose: @chanceaclark I see you thumbs down this comment. Can you expand more about the reason why not?

chanceaclark: See https://github.com/bigcommerce/big-design/pull/933#discussion_r936046033

deini: @chanceaclark I don't think an updater function helps here. It's not being called multiple times and this is a click handler, which will get automatic batches even in React < 18.

BC-Turner (commented):
Couple comments:

1. When collapsing the component, it looks like the 2nd child overlaps the focused state of the first. Is there a way to bring that focused state forward so we don't lose part of the stroke?
Screen Shot 2022-08-03 at 9 13 53 AM

2. For the implementation example, let's use either text or radio buttons for both panels and have either icons on both or none. I think we don't want users mixing panels of different content and keeping consistent with either having an icon on all or no icons. Just to avoid confusion.

Otherwise it looks great!

packages/big-design/src/components/AccordionGroup/hooks/useAccordion.tsx
chanceaclark: 🤦‍♂️ I retract my 👎 

Either way should be fine 🤔

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
rtalvarez: Good comment, I think we should probably restrict the `style` as well

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
rtalvarez: Sorry kinda late to the party here, but should these icons be `<Button iconOnly={...} />` ? Or is that what you mean?

rtalvarez (commented):
Sorry kinda late to the party here, but this is looking great so far

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
chanceaclark: ^ Yeah, was trying to suggest using the built-in `icon*` props instead of a custom grid. We can style the `StylableButton` so if we need to move things around we can update it in the styled-component rather than markup.

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
bc-athorne: Wondering if I still need to do this if I added @chanceaclark's suggestion below about removing the ability to restyle 
```
  ({ children, header, iconLeft, defaultExpanded = false, className, style, ...props }) => {
  ```

Comment (MariaJose) [reply packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx:27]: I think this is good ^

File: packages/big-design/src/components/AccordionGroup/hooks/useAccordion.tsx (hunk 3 above)
Comment (jorgemoya) [inline packages/big-design/src/components/AccordionGroup/hooks/useAccordion.tsx:8]: Lets initialize the state to `defaultExpanded`, so that you don't need to use the `useEffect` below.

```suggestion
  const [isExpanded, setisExpanded] = useState(defaultExpanded);
```

packages/docs/pages/accordion-panel.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <>Do not use an accordions to hide a singular piece of content.</>,
+          ]}
+          recommended={[
+            <>Choose meaningful headers that describe the content revealed on expand.</>,
+            <>
+              Use the icon prop sparingly; an appropriate use would be the status of a task hidden
+              in the accordions.
+            </>,
+            <>Use accordions to present parallel content or like items.</>,
+            <>Limit the number of tasks hidden in each accordions to one.</>,
```
bc-andreadao, line 139:
```suggestion
            <>Limit the number of tasks hidden in each accordion to one.</>,
```

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { ExpandLessIcon, ExpandMoreIcon } from '@bigcommerce/big-design-icons';
+import React, { HTMLAttributes, memo } from 'react';
+
+import { useUniqueId } from '../../../hooks';
+import { BoxProps } from '../../Box/Box';
+import { useAccordion } from '../hooks/useAccordion';
+
+import { StyledAccordion, StyledAccordionButton, StyledBox, StyledGridItem } from './styled';
+
+export interface AccordionPanelProps extends HTMLAttributes<HTMLDivElement> {
```
jorgemoya, line 11: Should this actually extend from `PanelProps`? 🤔

packages/big-design/src/components/AccordionGroup/hooks/useAccordion.tsx
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from 'react';
+
+interface UseAccordionProps {
+  defaultExpanded?: boolean;
+}
+
+export const useAccordion = ({ defaultExpanded }: UseAccordionProps) => {
```
jorgemoya, line 7: No need to pass in as an object, instead do:

```suggestion
export const useAccordion = (defaultExpanded: boolean) => {
```

I removed the optional because you already set `defaultExpanded` to `false` in the AccordionPanel component.

jorgemoya (commented):
Looking good! Thanks for working on this Amanda. Left some comments, let me know what you think.

packages/big-design/src/components/AccordionGroup/AccordionGroup.tsx
bc-athorne: I didn't use `Panel` here because the padding was an issue since it's not styleable.

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
bc-athorne: I kept this as is since the parent is not a `Panel.`

bc-athorne:
Having a similar issue as mentioned before (https://github.com/bigcommerce/big-design/pull/933#issuecomment-1203192138) with one test failing on CircleCI but passes on my local. It shows the appropriate attributes for the component after it has been clicked.
Screen Shot 2022-08-10 at 12 23 20 PM

```
test('accordion panel collapses and expands on click', async () => {
  render(mockAccordionGroup);

  const buttonPanel = await screen.findByRole('button');

  userEvent.click(buttonPanel);

  const panelChild = await screen.findByRole('region');

  expect(panelChild).toBeVisible();

  userEvent.click(buttonPanel);

  expect(panelChild).not.toBeVisible();
});
```

In CircleCI:
Screen Shot 2022-08-10 at 12 29 59 PM

packages/big-design/src/components/AccordionGroup/spec.tsx
bc-athorne: Updated the tests but still seeing a failure for the second assertion - https://github.com/bigcommerce/big-design/pull/933#issuecomment-1211032183

BC-Turner (approved):
Looks awesome! Great job, Amanda!

packages/big-design/src/components/AccordionGroup/AccordionGroup.tsx
chanceaclark: Why do we need it to be styleable?

bc-athorne: The default panel has padding and I need the child panel component's border to extend to the full width of the parent container.

chanceaclark: Can we just add a wrapper component to do that using negative margins so we can just use our existing Panel component?

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
bc-athorne: Can you provide some clarification on why you suggest this?

packages/big-design/src/components/AccordionGroup/spec.tsx
bc-athorne: Added @MariaJose's suggestion (https://github.com/bigcommerce/big-design/pull/933#issuecomment-1203259014) to fix it

packages/big-design/src/components/AccordionGroup/AccordionGroup.tsx
bc-athorne: Added a wrapper around the panels to resolve this 👍

packages/docs/pages/accordion-panel.tsx
```diff
@@ -0,0 +1,122 @@
...
+                  </Text>
+                ),
+              },
+            ];
+
+            return (
+              <>
+                <AccordionGroup header="Accordion Group Panel Header" panels={panels} />
+              </>
+            );
```
chanceaclark, line 77: 🍹 Useless fragmetn
```suggestion
            return (
                <AccordionGroup header="Accordion Group Panel Header" panels={panels} />
            );
```

packages/docs/pages/accordion-panel.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+      <Panel header="Implementation" headerId="implementation">
+        <CodePreview>
+          {/* jsx-to-string:start */}
+          {function Example() {
+            const [isExpanded, setIsExpanded] = useState(false);
+
+            const onClick = () => {
+              setIsExpanded(!isExpanded);
+            };
```
chanceaclark, line 42: What is this for, when we already manage state inside the component? If this was actually working (by passing these props through the component), we'd see both panels open and close at the same time.

packages/big-design/src/components/AccordionGroup/AccordionGroup.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const AccordionGroup: React.FC<AccordionGroupProps> = memo(({ header, panels }) => {
+  const renderedPanels = useMemo(() => {
+    return panels.map((panel, index) => <AccordionPanel {...panel} key={index} />);
+  }, [panels]);
+
+  return (
+    <Panel header={header}>
+      <StyledAccordionPanelWrapper>{renderedPanels}</StyledAccordionPanelWrapper>
+    </Panel>
+  );
```
chanceaclark, line 22: 🍹 This might be a useless `useMemo` 🤔 Can we confirm that?

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,57 @@
...
+          isExpanded={isExpanded}
+          onClick={onClick}
+          variant="subtle"
+        >
+          <Text marginBottom="none" marginLeft={iconLeft ? 'xxxLarge' : 'none'}>
+            {header}
+          </Text>
+        </StyledAccordionButton>
+        <StyledAccordionContent
+          {...props}
```
chanceaclark, line 44: Why are we passing in the same props for all three components 🤔

chanceaclark (commented):
I have a few more comments for ya @bc-athorne but thanks for making those updates!

packages/big-design/src/components/AccordionGroup/AccordionPanel/styled.tsx
```diff
@@ -0,0 +1,60 @@
+import { theme as defaultTheme, remCalc } from '@bigcommerce/big-design-theme';
+import styled, { css } from 'styled-components';
+
+import { withTransition } from '../../../mixins/transitions';
+import { Box } from '../../Box';
+import { StyleableButton } from '../../Button/private';
+
+import { AccordionPanelProps } from './index';
+
+export const StyledAccordionButton = styled(StyleableButton)<AccordionPanelProps>`
```
MariaJose, line 10: Should we create an interface for the `StyledAccordionButton` to define the props that is gonna use instead of passing the `<AccordionPanelProps>` ? 

Something like this:

```tsx
interface StyledAccordionButtonProps {
  isExpanded: boolean;
}

StyledAccordionButton = styled(StyleableButton)<StyledAccordionButtonProps>`
```

packages/big-design/src/components/AccordionGroup/AccordionPanel/styled.tsx
```diff
@@ -0,0 +1,60 @@
...
+      `}
+  }
+
+  p {
+    grid-row: 1;
+    grid-column: 1;
+  }
+`;
+
+export const StyledAccordionContent = styled(Box)<AccordionPanelProps>`
```
MariaJose, line 54: Same here as well ?

packages/docs/PropTables/AccordionPanelPropTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+  },
+
+  {
+    name: 'onClick',
+    types: '() => void',
+    description: (
+      <>
+        Function that will be called when an <Code primary>Accordion Panel</Code> is clicked.
+      </>
+    ),
```
MariaJose, line 100: Probably we might not need the `onClick` prop since we handle the state internally, unless we want the user manage it? If that's the case we would have to think about an API

MariaJose (commented):
Looking good! 👏🏼   , just left a few comments

rtalvarez (approved):
🐑

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { StyledAccordionButton, StyledAccordionContent } from './styled';
+
+export interface AccordionPanelProps extends HTMLAttributes<HTMLDivElement> {
+  children?: React.ReactNode;
+  defaultExpanded?: boolean;
+  header: string;
+  iconLeft?: React.ReactNode;
+  isExpanded: boolean;
+  onClick?: () => void;
```
MariaJose, line 17: I think we could delete `onClick` ?

MariaJose (approved):
Lgtm 👍🏼 !

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
chanceaclark: And `isExpanded`

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
chanceaclark: Side bar from Jorges comment, do we even need consumers to pass in `HTMLDivElement` attributes? If we leave this out right now, we can always add back in when consumers need it. We can't remove it without a breaking change if we add it in now.

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,55 @@
...
+          aria-controls={accordionItemId}
+          aria-expanded={isExpanded}
+          iconLeft={iconLeft}
+          iconRight={<ExpandMoreIcon className="collapse-icon" color="secondary70" />}
+          id={accordionId}
+          isExpanded={isExpanded}
+          onClick={onClick}
+          variant="subtle"
+        >
+          <Text marginBottom="none" marginLeft={iconLeft ? 'xxxLarge' : 'none'}>
```
chanceaclark, line 38: Since this is using the `StylableButton`, couldn't we just apply some grid CSS to the `span` instead of passing a `p` tag into a `button`? Trying to see if we can cut unnecessary html elements from the DOM.

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
bc-athorne: Do we want to remove `isExpanded`? I thought this was needed if we wanted this to be uncontrolled / controlled. Although, I'm not 100% clear on that.

chanceaclark: Handling the state inside the component makes it a controlled component, which you are doing with this hook: https://github.com/bigcommerce/big-design/pull/933/files#diff-5fc0e597eae1361cf0fba2493f91154fba01f35b851baace00572488ec7ec010R22

If we want consumers to control the state, we'd elevate that outside of the component and let consumers pass it in.

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React, { HTMLAttributes, memo } from 'react';
+
+import { useUniqueId } from '../../../hooks';
+import { Box, BoxProps } from '../../Box/Box';
+import { useAccordion } from '../hooks/useAccordion';
+
+import { StyledAccordionButton, StyledAccordionContent } from './styled';
+
+export interface AccordionPanelProps extends HTMLAttributes<HTMLDivElement> {
+  children?: React.ReactNode;
```
bc-juanvasquez, line 12: have we tested nested children? Not sure if `memo` would interfere here with the diffing algo..

bc-apostoliuk (approved):

packages/big-design/src/components/AccordionGroup/AccordionPanel/AccordionPanel.tsx
rtalvarez: I don't think this is entirely accurate -- this component is purposefully meant to be a controlled component, hence the ability for consumers to provide `isExpanded` and `onClick`, and we are definitely elevating these as far as I can see (through the `panel` items).

We should, however, export the `useAccordion` hook (I now realize its not being exported, maybe this is the source of the disconnect?), so consumers can opt for controlled or uncontrolled accordions.

MariaJose: good call about the hook 👀 , didn't remember about that lol . If that's the case `onClick` and `isExpanded` will be necessary

bc-athorne: Sorry for the confusion 🤦‍♀️. I moved `useAccordion` outside of the component.

chanceaclark: The problem I see if that those props are not even being used, so we either need to use them or lose them.

rtalvarez: We _could_ drop `onClick`, but `isExpanded` is literally a required prop, it will get used. Not sure I follow

chanceaclark: What I mean is when you pull down the component, the passed in `isExpanded` and `onClick` are technically not being used/consumed within the component. `onClick` only works because we are spreading it into `<Box {...props}>` and the handler is being called when you expand/collapse the accordion. The passed `props.isExpanded` though is not being consumed anywhere, the state is being respected only through the state returned from `useAccordion`.

rtalvarez: Was chatting with Chance this morning in the office, and cleared up some of the misunderstandings in this thread. The issues, as they stand right now, are:

- The component is not honoring the provided `isExpanded` / `onClick` props. To correctly visualize this, we can think of a consumer that programatically mutates this prop. Since it's not going through `onClick`, it will not work
- Because of the above, the `useAccordion` hook kinda operates on the item level, but the exported element is the group level. This also means we only need to export 1 component (the group one)
- The Hooks folder is not being exported, so we need to move the `useAccordion` hook to the component folder and probably export it from there
- Some of the names could be tweaked to make more sense, stylistic changes, etc

We can sync offline and discuss these issues, IMO we'll need to address them before we move forward

packages/big-design/src/components/AccordionPanel/spec.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+let panels: AccordionProps[];
+
+const TestComponent: React.FC = () => {
+  ({ panels } = useAccordionPanel([
+    {
+      header: 'Panel Header',
+      children: <Text>This is a child component</Text>,
+    },
+  ]));
```
chanceaclark, line 21: Maybe we should just assign this to a local const instead of using this type of JavaScript sorcery.
```suggestion
  const { panels } = useAccordionPanel([
    {
      header: 'Panel Header',
      children: <Text>This is a child component</Text>,
    },
  ]);
```
Same below...

chanceaclark (approved):
One small little fix, otherwise looks good from my end.

rtalvarez (approved):
Thanks for all the changes 🙏

PR #941, opened 2022-08-09
chore: remove TableNext export

chanceaclark:
## What?/Why?

Removes the `TableNext` export. This ensures a few things:
1. Reduces our main bundle size.
2. Let's consumers know that you need to explicitly import it to use it. It'll be a breaking change when we switch `TableNext` -> `Table`.
3. Allows us to make this an alpha release to test it. **Further releases will not contain this export (see note above).**

To consume, you'll import it via:
```tsx
import { TableNext } from '@bigcommerce/big-design/dist/es/components/TableNext';
```

If you are using TypeScript, add this to your `global.d.ts` or wherever to get types:
```ts
declare module '@bigcommerce/big-design/dist/es/components/TableNext' {
  export * from '@bigcommerce/big-design/dist/components/TableNext';
}
```

Consumed in a undisclosed app:
Screen Shot 2022-08-09 at 16 53 45

## Testing/Proof

Replaced the node_module locally and consume it in an app.

### DIFF
--- a/packages/big-design/src/components/index.ts
+++ b/packages/big-design/src/components/index.ts
@@ -34,7 +34,6 @@ export * from './StatefulTree';
 export * from './Stepper';
 export * from './Switch';
 export * from './Table';
-export * from './TableNext';
 export * from './Tabs';
 export * from './Textarea';
 export * from './Timepicker';


MariaJose (approved):
Thanks for the update!

jorgemoya (approved):

bc-juanvasquez (approved):
👍💡

bc-athorne (approved):

PR #961, opened 2022-08-31
fix(component): render empty string when prop is undefined

jorgemoya:
## What?

Render empty cells in `Worksheet` when a prop is missing in the item, instead of `undefined`. Shouldn't change the values of onChange until values are actually edited.

## Why?

When I build it I was under the assumption that the items should always have all the props but this is not always true.

Before:
Screen Shot 2022-08-31 at 4 52 40 PM

After:
Screen Shot 2022-08-31 at 4 51 49 PM

## Testing/Proof

Updated unit tests to reflect this change.

### DIFF
--- a/packages/big-design/src/components/Worksheet/Row/Row.tsx
+++ b/packages/big-design/src/components/Worksheet/Row/Row.tsx
@@ -67,7 +67,7 @@ const InternalRow = <T extends WorksheetItem>({ columns, rowIndex }: RowProps<T>
           rowIndex={rowIndex}
           type={column.type ?? 'text'}
           validation={column.validation}
-          value={row[column.hash]}
+          value={row[column.hash] ?? ''}
         />
       ))}
     </StyledTableRow>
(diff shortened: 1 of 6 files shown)


bc-andreadao (approved):

bc-andreadao (approved):

packages/big-design/src/components/Worksheet/Row/Row.tsx
```diff
@@ -67,7 +67,7 @@ const InternalRow = <T extends WorksheetItem>({ columns, rowIndex }: RowProps<T>
           rowIndex={rowIndex}
           type={column.type ?? 'text'}
           validation={column.validation}
-          value={row[column.hash]}
+          value={row[column.hash] ?? ''}
```
bc-juanvasquez, line 70: Isn't this an issue with `renderedCell` in Cell`? I see references to `renderedValue` taking care of this issue but it's not being used for every cell type...

jorgemoya: Let me take a look

chanceaclark (approved):

packages/big-design/src/components/Worksheet/Row/Row.tsx
jorgemoya: `renderedValue` in `Cell` is only used for `Text/Number` columns when they're not being edited, you're correct, all the other cell types handle that on their own. It is way easier for me to just pass in a default value than to update all the "rendered values" for all cell types.

jorgemoya: Will make some changes, though, since we get a default value.

bc-juanvasquez (approved):

PR #984, opened 2022-09-23
feat(component): change updating cell strategy for text field

bc-apostoliuk:
## What?
Change `Worksheet` updating cell strategy for the `text` and `number` fields.

- Use cases:

1. When opening the Inventory Editor the top left cell gets the Focused state (so merchants can enter value right after accessing the page with no additional clicks on the cell.

2. When a cell is Focused (by keyboard or mouse) it can be edited straight away by:

- Inserting char or numeric values.

- Press Enter from the keyboard.

- Single mouse click.

3. When a cell is Edited all changes can be saved by the “lost focus” user’s action:

- pressing the Tab button - the cell to the right becomes Focused.

- pressing Enter button - the cell to the bottom becomes Focused.

- selecting any other cell by mouse - selected cell becomes Focused.

4. When a cell is Edited all changes can be canceled by pressing the Esc button. Edited cell changes its state to the Focused one.

5. When changing the value within the cell, the previous one should be replaced with the newly entered value (currently new value is added to the existing one).

## Why?

It needs to be changed to the new strategy according to the new guideline.

https://user-images.githubusercontent.com/84462142/192572546-3300a55b-8dab-4772-8764-366451240372.mov

## Testing/Proof

Explain how you tested your new feature or bugfix. Please refrain from one-liners like, `Added tests.`; expand upon how you tested your changes like, `Added local tests to the new component and ensured test coverage was met. Manually tested on Safari@x.x.x.`.


vdm-kr (approved):
lgtm

jorgemoya (changes requested):
Will need to figure out how we want to handle arrow navigation with this behavior. Do we just lose it? Do we need to fix it? Tests are breaking because of this.

chanceaclark:
IMO, we probably shouldn't introduce this behavior. Merchants/users already have the habit of pressing enter to edit a cell based on using Excel or Google Sheets. And like Jorge mentioned, there are other behaviors we need to account for including event handling (onChange, onError, etc...). We also need to add a few tests for this.

vdm-kr:
@chanceaclark actually you don't have to press enter to edit a cell in excel/google docs or our old balk edit. And that's what we wanted to achieve

chanceaclark:
> @chanceaclark actually you don't have to press enter to edit a cell in excel/google docs or our old balk edit. And that's what we wanted to achieve

Ah I see now... however I don't think that this change is the _right fix_ for the problem we are trying to solve. I know you can start typing when you are focused on a cell, but you need to initially be focused on that cell so this onblur/onfocus method is not the right approach.

https://user-images.githubusercontent.com/10539418/192005333-64ab5639-57fc-45b7-a52a-4e3e4563a2ed.mov

The fix I see is that if you are focused on a cell and use a keyboard to start typing, then you start editing – not click on a cell and now it's in editing mode.

DmytroSirenko:
@kristinapototska, FYI.

kristinapototska:
> @kristinapototska, FYI. There is a discussion about saving values in cells on the "lost focus" behavior.

Thanks Dmytro. I can see that @vdm-kr has already shared the expected behaviour here. We do want the experience to be intuitive and similar to worksheet editing experience available with other tools, like good sheets etc.
Also we can use bulk edit as an example.
When the focus is on cell (for text fields) you can start typing new value right at once with no need to do additional clicks. And we've previously discussed with Vadym that when inventory editor is opened, ideally would have the default focus on `stock` cell so that merchants can start typing the value right after the page is loaded which would save time and reduce the amount of clicks. 
If the `focus is lost` (e.g. merchant clicked outside the table, the editing will be unavailable till merchant click on the cell again for it to be in focus.

What's the open question that we still need to figure out here?

bc-apostoliuk:
@chanceaclark, @jorgemoya thanks for the comments, I have changed the implementation approach, please review again.

jorgemoya:
> @chanceaclark, @jorgemoya thanks for the comments, I have changed the implementation approach, please review again.

Can you share a recording of the new approach?

bc-apostoliuk:
@jorgemoya I`ve reattached the new recording to the PR. Please check it there.

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
```diff
@@ -158,18 +164,19 @@ const InternalCell = <T extends WorksheetItem>({
...
-    isEdited,
-    isEditing,
     options,
+    formatting,
     rowId,
+    disabled,
+    editWithValue,
+    isEdited,
+    handleKeyDown,
     renderedValue,
```
jorgemoya, line 179: We prefer to `abc`.

packages/big-design/src/components/Worksheet/hooks/useEditableCell/useEditableCell.ts
```diff
@@ -40,9 +41,22 @@ export const useEditableCell = <T extends WorksheetItem>(cell: Cell<T>) => {
...
+        updateItems(
+          [selectedCells[0]],
+          [isNumberCell ? Number(event?.target.value) : event?.target.value],
+        );
+      }
+
+      restoreFocus();
+    },
+    [restoreFocus, selectedCells, updateItems],
+  );
```
jorgemoya, line 59: Hmm if we had a multiple selection of cells (in the future), how would this be managed? Can we pass the the `cell` as a param instead?

bc-apostoliuk: @jorgemoya This behavior should work only for the first selected cell. What do you mean by passing the `cell` as a param?

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
bc-apostoliuk: @jorgemoya that was the linter auto-fix, I have returned it back))

jorgemoya (approved):
Looks good to me 👍

MariaJose (approved):

eugene-polev:
@bc-apostoliuk , @vdm-kr should we highlight the value inside the cell when it is focused as we do for inputs?

Screenshot 2022-09-28 at 12 26 06
Screenshot 2022-09-28 at 12 24 10

bc-apostoliuk:
@eugene-polev I guess not if we are trying to reproduce the Excel behavior, since over there the selection of the field does not highlight the text inside.

PR #975, opened 2022-09-19
feat(component): allow custom column widths and overflow worksheet

jorgemoya:
## What?

**Breaking changes (??)**

- Allow custom widths to Worksheet columns.
- Scrollable overflow when columns are wider than table view. 
- Columns wont ellipsis anymore, instead words will wrap.
- When all columns have a fixed width, set Worksheet to auto width (instead of 100%).

## Screenshots/Screen Recordings
**Before**:
Kapture 2022-09-19 at 16 32 16

**After**:
Screen Shot 2022-09-19 at 3 56 10 PM

Kapture 2022-09-19 at 16 20 34

All columns are fixed width:
Screen Shot 2022-09-19 at 4 22 27 PM

At least one column is auto:
Screen Shot 2022-09-19 at 4 23 19 PM

## Testing/Proof

Locally, added unit tests.

PR #991, opened 2022-09-28
fix(component): prevent column resizing when editing text field

jorgemoya:
## What?

Sets Worksheet to `table-layout: fixed` and add `minWidth` prop.

Fixes this issue introduced by #975:

Kapture 2022-09-28 at 12 01 48

## Why?

The bug is happening because I set `table-layout: auto`, which means the table will try to automatically fit the content in the table (this is cool if we want all columns to remain visible even when shrinking the window), and overflow what doesn't fit (so we get a scroll bar).

Problem is that the `input` element is set to `width: 100%` and unless the `td` has a fixed width (set with the new width prop in the column), it won't calculate the 100% as I expected.

What I'm doing is reverting the table to `table-layout: fixed`. The problem with fixed and having static width columns it that when the table shrinks in size it will shrink all auto columns to 0 to accommodate the fixed width columns.

Like so:

Kapture 2022-09-28 at 10 46 43

To solve this I expose a `minWidth` prop for the Worksheet, so that we can set a minimum width. However this needs to be controlled by the developer, otherwise the issue above will happen.

Kapture 2022-09-28 at 10 49 35

Table will still have a static width if all columns have static widths.

Fixed inputs:

Kapture 2022-09-28 at 12 08 21

Overall this feels like a good compromise. We added the ability to set fixed widths to columns but by doing so developer needs to be mindful of setting a min-width if viewport gets too shrunk for the table. I believe `minWidth` was long overdue because the Worksheet has been breaking this way since it was introduced.

## Testing/Proof

Locally.

### DIFF
--- a/packages/big-design/src/components/Worksheet/spec.tsx
+++ b/packages/big-design/src/components/Worksheet/spec.tsx
@@ -1,5 +1,6 @@
 import { theme } from '@bigcommerce/big-design-theme';
 import { fireEvent, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
 
 import { StatefulTree } from '../StatefulTree';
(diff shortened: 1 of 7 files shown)


packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -726,11 +727,11 @@ describe('SelectEditor', () => {
 
     const cells = await findAllByDisplayValue('Value 1');
 
-    fireEvent.click(cells[0]);
+    await userEvent.click(cells[0]);
```
MariaJose, line 730: 🙇🏼‍♀️

MariaJose (approved):

bc-athorne (approved):
lgmt 👍

chanceaclark (approved):

PR #1000, opened 2022-10-06
fix(component): fix table selection bug on individual rows

chanceaclark:
## What?

Fixes a bug where selected indexes where getting filtered out that started with the row index you deselected. For example, if you deselected row `2` (index `1`), it would filter out indexes that started with `1` i.e. 13, 1003, etc...

## Why?

This was causing us to receive an invalid dictionary of selected items when deselecting a table row.

https://user-images.githubusercontent.com/10539418/194344473-54e02f0d-f993-4042-a37f-4e5ac5d871f7.mov

## Testing/Proof

See recording, plus updated tests.


bc-juanvasquez (approved):
💭

MariaJose (approved):
👍🏼 !

PR #1008, opened 2022-10-12
feat(component): provide props for customising pagination button labels

BC-SEven:
## What?

Adding a `label`, `previousLabel` , `nextLabel` and `getLabel` prop to the Pagination component that customise the labels of the navigation buttons and range selector 

## Why?

Allow greater customisability of the component.

https://user-images.githubusercontent.com/109577044/195255292-47d6a00d-ea04-4765-858e-30ceebe9dcf5.mov

## Testing/Proof

Added jest spec tests to define behaviour of new pagination props.

### DIFF
--- a/packages/big-design/src/components/Pagination/Pagination.tsx
+++ b/packages/big-design/src/components/Pagination/Pagination.tsx
@@ -18,6 +18,9 @@ export interface PaginationProps extends MarginProps {
   totalItems: number;
   onPageChange(page: number): void;
   onItemsPerPageChange(range: number): void;
+  previousPageLabel?: string;
+  nextPageLabel?: string;
+  getRangeLabel?(start: number, end: number, totalItems: number): string;
 }
 
 export const Pagination: React.FC<PaginationProps> = memo(
@@ -28,6 +31,9 @@ export const Pagination: React.FC<PaginationProps> = memo(
     itemsPerPageOptions = [],
     onPageChange,
     onItemsPerPageChange,
+    previousPageLabel = 'Previous page',
+    nextPageLabel = 'Next page',
+    getRangeLabel = null,
   }) => {
     const [maxPages, setMaxPages] = useState(Math.max(1, Math.ceil(totalItems / itemsPerPage)));
     const [itemRange, setItemRange] = useState({ start: 0, end: 0 });
@@ -90,11 +96,10 @@ export const Pagination: React.FC<PaginationProps> = memo(
       return onItemsPerPageChange(Number(item.hash));
     };
 
-    const showRanges = () => {
-      return itemRange.start === itemRange.end
-        ? `${itemRange.start} of ${totalItems}`
-        : `${itemRange.start} - ${itemRange.end} of ${totalItems}`;
-    };
+    if (getRangeLabel === null) {
+      getRangeLabel = (start: number, end: number, totalItems: number): string =>
+        start === end ? `${start} of ${totalItems}` : `${start} - ${end} of ${totalItems}`;
+    }
 
     return (
       <Flex aria-label="pagination" flexDirection="row" role="navigation">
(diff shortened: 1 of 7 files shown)


packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -18,6 +18,9 @@ export interface PaginationProps extends MarginProps {
   totalItems: number;
   onPageChange(page: number): void;
   onItemsPerPageChange(range: number): void;
+  previousPageLabel?: string;
+  nextPageLabel?: string;
+  getRangeLabel?(start: number, end: number, totalItems: number): string;
```
chanceaclark, line 23: Overall I think this is great! I'd love to sort of condense the names just a bit and add one more if that cool? Generally, I'd like this to align with Atlaskit: https://atlassian.design/components/pagination/code
```suggestion
  label?: string;
  previousLabel?: string;
  nextLabel?: string;
  getRangeLabel?(start: number, end: number, totalItems: number): string;
```

The new addition `label` would be for `aria-label="pagination"`.

packages/big-design/src/components/Pagination/Pagination.tsx
```diff
@@ -90,11 +96,10 @@ export const Pagination: React.FC<PaginationProps> = memo(
...
 
-    const showRanges = () => {
-      return itemRange.start === itemRange.end
-        ? `${itemRange.start} of ${totalItems}`
-        : `${itemRange.start} - ${itemRange.end} of ${totalItems}`;
-    };
+    if (getRangeLabel === null) {
+      getRangeLabel = (start: number, end: number, totalItems: number): string =>
+        start === end ? `${start} of ${totalItems}` : `${start} - ${end} of ${totalItems}`;
+    }
```
chanceaclark, line 102: How about we extract this outside of the component (same file) and default prop the `getRangeLabel`? I would also love to get rid of the ternary statement too 😬 
```tsx
const DEFAULT_GET_RANGE_LABEL: PaginationProps['getRangeLabel']  = (start, end, totalItems) => {
  if (start === end) {
    return `${start} of ${totalItems}`;
  }

  return `${start} - ${end} of ${totalItems};
}
```

packages/big-design/src/components/Table/TablePagination/TablePagination.tsx
```diff
@@ -6,24 +6,10 @@ import { TablePaginationProps } from '../types';
...
 export const TablePagination: React.FC<TablePaginationProps> = memo(
-  ({
-    currentPage,
-    itemsPerPage,
-    itemsPerPageOptions,
-    onItemsPerPageChange,
-    onPageChange,
-    totalItems,
-  }) => {
+  (props: TablePaginationProps) => {
```
chanceaclark, line 9: Can you revert this back? We actually did that to exclude the margin props to enforce design consistency. (Maybe add a comment about that?

packages/big-design/src/components/TableNext/TablePagination/TablePagination.tsx
```diff
@@ -6,24 +6,10 @@ import { TablePaginationProps } from '../types';
...
 export const TablePagination: React.FC<TablePaginationProps> = memo(
-  ({
-    currentPage,
-    itemsPerPage,
-    itemsPerPageOptions,
-    onItemsPerPageChange,
-    onPageChange,
-    totalItems,
-  }) => {
+  (props: TablePaginationProps) => {
```
chanceaclark, line 9: Same here

chanceaclark (commented):
This is something we've been meaning to get around to but haven't yet! Thanks for taking the time and effort to do so! I have a few comments, otherwise looks really good! 🙇

chanceaclark (approved):

davelinke (commented):
In terms of visuals I assume the getRangeLabel prop is to allow diverse languages to be used. I'm worried of people abusing it, but we can enforce misuse in demo reviews.

I got one question; Since pagers are usually found at the bottom of pages, does the dropdown open upwards in case there is not enough space for the dropdown to open downwards?

chanceaclark:
>Since pagers are usually found at the bottom of pages, does the dropdown open upwards in case there is not enough space for the dropdown to open downwards?

Yes, our popper library takes care of that.

PR #1015, opened 2022-10-14
fix(component): keys presses fill the field in with the key code in worksheet

MariaJose:
## What?
Keys presses fill the field in with the key code in worksheet. 

## Why?
We are passing the key code as an argument in `editSelectedCell` fn when entering info in a cell.

Examples: 
- if we use the `command` key the cell is filled with its corresponding code: 'Meta'
- if we press key `a` the code its `a`

Current solution (patch):
- Checking the length of the key. And check when using Meta and Control to be able to copy content in a cell.

Note: would be good to refactor our current code but might take more time.

Before:

https://user-images.githubusercontent.com/39739180/195959199-bdb4b31d-0d87-4b8d-bb89-85feb8032436.mp4

After

https://user-images.githubusercontent.com/39739180/196284677-91f906ac-3f09-4b4e-8c10-2c67dc210a49.mp4

## Testing/Proof
Test pass

### DIFF
--- a/packages/big-design/src/components/Worksheet/hooks/useKeyEvents/useKeyEvents.ts
+++ b/packages/big-design/src/components/Worksheet/hooks/useKeyEvents/useKeyEvents.ts
@@ -86,6 +86,7 @@ export const useKeyEvents = () => {
           default:
             if (
               key !== 'Escape' &&
+              key.length === 1 &&
               (selectedCell.type === 'text' || selectedCell.type === 'number')
             ) {
               event.preventDefault();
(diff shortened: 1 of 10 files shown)


packages/big-design/src/components/Worksheet/hooks/useKeyEvents/useKeyEvents.ts
```diff
@@ -83,14 +83,25 @@ export const useKeyEvents = () => {
...
+            editSelectedCell();
+            break;
+
+          case 'Control':
+            editSelectedCell();
+            break;
+
           default:
             if (
               key !== 'Escape' &&
```
chanceaclark, line 96: Could we do something like this instead of adding more `case` statements?
```suggestion
              !['Meta', 'Control', 'Escape'].includes(key) &&
```
Seems like we don't want to call `editSelectedCell();` when those keys are pressed 🤔

MariaJose: The problem is that if we include Meta , and Control, the use is not gonna be able to copy and paste the new content in the cell 👀

chanceaclark (approved):
👏

jorgemoya (approved):

Issue #1022, opened 2022-10-20
Accordion doesn't have a event.preventDefault which causes a form submit on click


tomzweb:
Using version 0.34.1, when the AccordionPanel component is wrapped in a Form component and an Accordion header is clicked, the forms submit event is fired.

According to the docs https://developer.bigcommerce.com/big-design/accordion-panel?props=accordion - the panel should have an onClick method and isExpanded property. However, these are omitted in the InitialPanels type definition. Even when using // ts-ignore, the onClick method does not appear to be called, so I can't manage the functionality myself.

Steps to reproduce the behavior

1. create an accordion inside a form component.

```

<Form onSubmit={onSubmit}>
  <AccordionPanel header='Sign up form' panels={panels} />
</Form>
```

2. Click an accordion header
3. onSubmit is called

I would expect the behavior to not submit the form.

tomzweb:
Looks similar to the issue here: https://github.com/bigcommerce/big-design/issues/940#issuecomment-1209889097 - where buttons should have a `type=button` to prevent this.

It should be applied here: https://github.com/bigcommerce/big-design/blob/e22ffa2cc8b4321a656dd19c3f3d0785a24eab8d/packages/big-design/src/components/AccordionPanel/Accordion/Accordion.tsx#L24

I've copied the AccordionPanel (and children) into my own project and tested it. It fixes the issue.

chanceaclark:
`type=button` looks like the appropriate solution! PR's are welcomed if you want to open up one!

tomzweb:
@chanceaclark PR submitted: https://github.com/bigcommerce/big-design/pull/1023

chanceaclark:
Thanks @tomzweb! I'll go ahead and close this now! We can probably patch a release soon, just need to chat internally with the team first.

PR #1023, opened 2022-10-21
fix(component): add type button to StyledAccordionButton (1022)

tomzweb:
## What?

Fixes an issue that when accordion headers are clicked, they fire the 'enter' event, causing forms to submit. See issue https://github.com/bigcommerce/big-design/issues/1022

## Testing/Proof

Tested by adding type='button' to the StyledAccordionButton props and then clicking the attribute header.


rtalvarez (approved):
Hello, nice catch! Thanks for the PR

PR #1044, opened 2022-11-21
feat: convert to react@18 only

chanceaclark:
## What?

Converts BigDesign to be `react@18` only.

ℹ️ Reviewing by commit will be helpful!

BREAKING CHANGE:
Requires `react@18` or higher.

## Why?
So we can get the benefits of SSR react using the new `useId` hooks.

N/A

## Testing/Proof

- Flipped on strict mode to test if there were some concurrency issues, but didn't look like it. Had to flip back off because `react-beautiful-dnd` doesn't work with strict mode.
- Manual tested some components
- Updated snapshots

### DIFF
--- a/packages/docs/pages/_app.tsx
+++ b/packages/docs/pages/_app.tsx
@@ -10,7 +10,6 @@ import Head from 'next/head';
 import { useRouter } from 'next/router';
 import Script from 'next/script';
 import React, { useEffect } from 'react';
-import { UIDFork, UIDReset } from 'react-uid';
 import { ThemeProvider } from 'styled-components';
 
 import { BetaRibbon, SideNav, StoryWrapper } from '../components';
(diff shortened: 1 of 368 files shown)


MariaJose (approved):

jorgemoya (approved):
Has there been any discussion in replacing `react-beautiful-dnd`?

jorgemoya (approved):

packages/docs/pages/_app.tsx
```diff
@@ -90,44 +89,40 @@ const App = ({ Component, pageProps }) => {
           />
         </>
       )}
-      <UIDReset>
-        <UIDFork>
```
deini, line 94: 💭

deini (approved):

chanceaclark:
> Has there been any discussion in replacing `react-beautiful-dnd`?

We have had chats about every so often offline, but we should have a better discussion here on GitHub.

PR #934, opened 2022-08-08
fix(component): Fix space and alignment in TableNext CHP-7624

MariaJose:
## What?
Space between the arrow and the next column is too big.
The text in the first column in subrows is not vertically centered.

## Why?
Fix bug

## Screenshots/Screen Recordings\

Before:
Screen Shot 2022-08-08 at 10 20 05 AM

After:
Screen Shot 2022-08-08 at 10 21 25 AM

Screen Shot 2022-08-08 at 3 54 02 PM

Screen Shot 2022-08-08 at 10 21 44 AM

## Testing/Proof
Test locally pass

PR #1056, opened 2022-12-01
fix(component): fix table-next column alignment

golcinho:
## What?
Fix horizontal alignment for columns by adding missing props to the flex component.

**Note:** both examples have `align === right` for the `stock` column.

**Before fix:**
Screenshot 2022-12-01 at 16 28 50

**After fix:**
Screenshot 2022-12-01 at 16 29 39

## Testing/Proof
Locally.

### DIFF
--- a/packages/big-design/src/components/TableNext/Row/Row.tsx
+++ b/packages/big-design/src/components/TableNext/Row/Row.tsx
@@ -18,8 +18,8 @@ interface PrivateProps {
 }
 
 const ALIGN_MAP: Record<string, FlexedProps['justifyContent']> = {
-  right: 'flex-start',
-  left: 'flex-end',
+  left: 'flex-start',
+  right: 'flex-end',
   center: 'center',
 };
 
--- a/packages/big-design/src/components/TableNext/Row/Row.tsx
+++ b/packages/big-design/src/components/TableNext/Row/Row.tsx
@@ -5,6 +5,7 @@ import { typedMemo } from '../../../utils';
 import { MessagingButton } from '../../Button/private';
 import { Checkbox } from '../../Checkbox';
 import { Flex } from '../../Flex';
+import { FlexedProps } from '../../Flex/types';
 import { DataCell } from '../DataCell';
 import { OnItemSelectFn } from '../hooks';
 import { TableColumn, TableItem, TableSelectable } from '../types';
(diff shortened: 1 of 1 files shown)


vdm-kr (approved):
LGTM

vitalii-koshovyi (approved):

packages/big-design/src/components/TableNext/Row/Row.tsx
```diff
@@ -154,7 +161,11 @@ const InternalRow = <T extends TableItem>({
               verticalAlign={verticalAlign}
               width={isDragging ? cellWidth : width}
             >
-              <Flex alignItems="center" flexDirection="row">
+              <Flex
+                alignItems="center"
+                flexDirection="row"
+                justifyContent={align && ALIGN_MAP[align]}
+              >
```
jorgemoya, line 168: Does this mean that the `align` property in the `DataCell` is not used. Should we remove it altogether?

@MariaJose Do you remember what is the reasoning behind adding the `Flex` container?

jorgemoya (commented):
Should we add a test to make sure we don't miss this again?

packages/big-design/src/components/TableNext/Row/Row.tsx
golcinho: Let me check it. Also I will check thead cells

MariaJose: @jorgemoya  It was to fix alignment and space issues when the children rows use to have the checkbox component as well. But since we remove the checkbox for the children for now, we might not need the Flex component.

Reference: https://github.com/bigcommerce/big-design/pull/934/files

bc-juanvasquez: We will add the feature back, so let's persist the change if it resolved a styling issue.

MariaJose: cool with that ^ 👍🏼

golcinho: @jorgemoya thead cells working right.

> Does this mean that the align property in the DataCell is not used. Should we remove it altogether?

We need it for typography because the `align` property adds `text-align`.

bc-juanvasquez (approved):

jorgemoya (approved):

PR #1067, opened 2022-12-14
fix(component): combobox can have an accessible name when no label is provided to MultiSelect

BC-SEven:
## What?

The aria-label prop is used by both the `input` and descendent `combobox` elements when the label prop is missing. Currently only the `input` has aria props passed to it.

## Why?

In the promotions editor we have a MultiSelect without a label. The purpose of the dropdown is already clear from the surrounding context.
image
But currently there is no way to give the `combobox` an accessible name for screen readers etc.
image

image

## Testing/Proof

Added jest test for accessing combobox via role and name.
Confirm browser compatibility on doc screenshot shared above.

### DIFF
--- a/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/big-design/src/components/MultiSelect/MultiSelect.tsx
@@ -257,6 +257,15 @@ export const MultiSelect = typedMemo(
       stateReducer: handleStateReducer,
     });
 
+    let comboboxProps = getComboboxProps();
+
+    if (!label && props['aria-label']) {
+      comboboxProps = {
+        ...comboboxProps,
+        'aria-label': props['aria-label'],
+      };
+    }
+
     // Popper
     const referenceRef = useRef(null);
     const popperRef = useRef(null);
@@ -425,7 +434,7 @@ export const MultiSelect = typedMemo(
     return (
       <div>
         {renderLabel}
-        <div {...getComboboxProps()}>{renderInput}</div>
+        <div {...comboboxProps}>{renderInput}</div>
         <Box ref={popperRef} style={styles.popper} {...attributes.poppper} zIndex="popover">
           <List
             action={action}
--- a/packages/big-design/src/components/MultiSelect/spec.tsx
+++ b/packages/big-design/src/components/MultiSelect/spec.tsx
@@ -1,6 +1,6 @@
 import { ArrowBackIcon, ArrowForwardIcon, DeleteIcon } from '@bigcommerce/big-design-icons';
 import { remCalc } from '@bigcommerce/big-design-theme';
-import { render, screen } from '@testing-library/react';
+import { render, screen, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import 'jest-styled-components';
 import React, { createRef } from 'react';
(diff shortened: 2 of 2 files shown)


packages/big-design/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -257,6 +257,15 @@ export const MultiSelect = typedMemo(
       stateReducer: handleStateReducer,
     });
 
+    let comboboxProps = getComboboxProps();
+
+    if (!label && props['aria-label']) {
```
MariaJose, line 262: I think would be good to make `label` prop required and create a prop  called `hideLabel`, and if `hideLabel` is true we can internally assign 'aria-label' with the label value. Thoughts?

MariaJose (commented):
Thanks for the changes! left a suggestion, let me know what you think 👍🏼

chanceaclark:
According to W3, I think the [`combobox` role](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html#rps_label_combobox) actually needs an `aria-labelledby` 🤔 What do you think about implementing that instead? Am I missing something else?

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
BC-SEven: I would like to avoid modifying the props if I can avoid it. 😅  The `label` prop being optional and devs having the option to use `aria-label` is already the behaviour available for other controls like `<Input>` and `<Select>`.

BC-SEven:
> According to W3, I think the [combobox role](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html#rps_label_combobox) actually needs an aria-labelledby 🤔 What do you think about implementing that instead? Am I missing something else?

The combobox should use `aria-labelledby` when a label is visible and `aria-label` when there is no visible label. So both are valid.
https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#wai-aria-roles-states-and-properties-6
> If the combobox has a visible label and the combobox element is an HTML element that can be labelled using the HTML label element (e.g., the input element), it is labeled using the label element. Otherwise, if it has a visible label, the combobox element has [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) set to a value that refers to the labelling element. Otherwise, the combobox element has a label provided by [aria-label](https://w3c.github.io/aria/#aria-label).

This is also a bit messy because our multiselect is built on `downshift@6` which is based off ARIA 1.1 specifications. And W3 examples are using the ARIA 1.2 specs. I think once we migrate to `downshift@7` we will probably be accessible out of the box.

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
BC-SEven: >I see 👀 , I suggested this since I think would be good for us worrying about accessibility instead of the user, but up to us 👍🏼
I do like the idea of the library enforcing accessibility with required fields but I feel this is probably a pattern we would want consistently across all our components then right? 

> Would it be good to add a comment in the docs for MultiSelect mentioning that this component supports all its element attributes (or something you think might be helpful for the user to know)? 🍹 similar like the input component does: https://developer.bigcommerce.com/big-design/input

Yea agreed. Would be nice to document something like `Props of the native input component are also available`.
I might whip up a quick separate PR for this

PR #1069, opened 2022-12-16
fix(component): make iconOnly button square

Meztimuro:
## What?

Make iconOnly Button square.

## Why?

iconOnly Button is square in Designer examples.

image
image

## Testing/Proof

Checked visually and ran tests.


packages/big-design/src/components/Button/styled.tsx
```diff
@@ -58,13 +58,14 @@ export const StyledButton = styled.button<ButtonProps & MarginProps>`
...
-  ${({ iconOnly: icon, theme }) =>
-    icon &&
-    css`
-      padding: 0 ${theme.spacing.xSmall};
-    `};
+    ${({ iconOnly: icon, theme }) =>
+      icon &&
+      css`
+        padding: 0;
+        width: ${addValues(theme.spacing.xxLarge, theme.spacing.xxSmall)};
```
golcinho, line 66: can we use `min-width` here? `min-width` should help if an icon will be wider then expected.

eugene-polev (approved):

packages/big-design/src/components/Button/styled.tsx
Meztimuro: I guess, but the height is fixed, so it still will be able to overflow the button.

golcinho: oh, I check the existing code, we need to increase the horizontal padding instead of adding `width`.

Meztimuro: Setting vertical padding to 5px/0.3125rem is an option

golcinho (approved):

jorgemoya (approved):

chanceaclark:
Hey @Meztimuro, looks like this change visually broke the `Chip` component. Could you check that out?
<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>
	<td>
</table>

PR #1087, opened 2023-01-04
fix(component): only focus on TreeItem when focus within Tree component

chanceaclark:
## What?

Focuses on the `TreeItem` whenever the `document.activeElement` is within the `Tree`.

Also, updates `Tree` tests to reflect better testing patterns (see #635).

## Why?

This fixes the behavior we've seen for a while which causes the tree to be initially focused on initial render. This causes some accessibility concerns as the we place the user onto the tree instead of the initial page or, in the case of modals, launch order.

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td><video src="https://user-images.githubusercontent.com/10539418/210616392-446f8357-eecf-4f4d-a426-c61ea79b02a9.mov" />
	<td><video src="https://user-images.githubusercontent.com/10539418/210616149-bf553eac-49c6-487a-9016-b0cf461af9df.mov" />
</table>

## Testing/Proof

Added two additional tests, along with updating the `Tree` tests to reflect better testing patterns.

---

Closes #801

### DIFF
--- a/packages/big-design/src/components/StatefulTree/spec.tsx
+++ b/packages/big-design/src/components/StatefulTree/spec.tsx
@@ -1,7 +1,8 @@
 import { theme } from '@bigcommerce/big-design-theme';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
 
-import { fireEvent, render } from '@test/utils';
+import { fireEvent, render, screen } from '@test/utils';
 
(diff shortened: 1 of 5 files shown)


jorgemoya (approved):
lgtm!

MariaJose (approved):
👍🏼 !

PR #1095, opened 2023-01-11
feat(component): add header tooltip for worksheet component

bc-apostoliuk:
## What?

Added header tooltip for worksheet component.

Refactored rendering of the tooltip for the Table component.
(packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx)

## Why?

To be able to leave a short description of the column.

Screenshot 2023-01-11 at 12 16 27

## Testing/Proof

Locally + UT.

### DIFF
--- a/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
+++ b/packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
@@ -66,7 +66,6 @@ const InternalHeaderCell = <T extends TableItem>({
     );
   };
 
-  // TODO: Update this. Return early if we don't have tooltip.
   const renderTooltip = () => {
     if (typeof tooltip === 'string' && tooltip.length > 0) {
       return (
--- a/packages/big-design/src/components/Worksheet/Worksheet.tsx
+++ b/packages/big-design/src/components/Worksheet/Worksheet.tsx
@@ -1,7 +1,18 @@
-import React, { createContext, createRef, useEffect, useMemo, useRef } from 'react';
+import { BaselineHelpIcon } from '@bigcommerce/big-design-icons';
+import React, {
+  createContext,
+  createRef,
+  useCallback,
+  useEffect,
+  useId,
+  useMemo,
+  useRef,
+} from 'react';
 import { StoreApi } from 'zustand';
 
 import { typedMemo } from '../../utils';
+import { Box } from '../Box';
+import { Tooltip } from '../Tooltip';
 
 import { UpdateItemsProvider } from './context';
 import { BaseState, createWorksheetStore, useKeyEvents, useWorksheetStore } from './hooks';
@@ -31,6 +42,7 @@ const InternalWorksheet = typedMemo(
     const tableRef = createRef<HTMLTableElement>();
     const shouldBeTriggeredOnChange = useRef(false);
     const { store, useStore } = useWorksheetStore();
+    const tooltipId = useId();
 
     const setRows = useStore(store, (state) => state.setRows);
     const setColumns = useStore(store, (state) => state.setColumns);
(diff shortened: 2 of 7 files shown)


packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -1223,3 +1224,25 @@ describe('column widths', () => {
...
+
+    expect(getByTitle('Hover or focus for additional context.')).toBeTruthy();
+  });
+
+  test('renders tooltip when hovering on icon', async () => {
+    const { getByTitle } = render(
+      <Worksheet columns={columns} items={items} onChange={handleChange} />,
+    );
+
+    fireEvent.mouseOver(getByTitle('Hover or focus for additional context.'));
```
golcinho, line 1242:
```suggestion
    await userEvent.hover(getByTitle('Hover or focus for additional context.'));
```

packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -1223,3 +1224,25 @@ describe('column widths', () => {
...
+
+    const result = await screen.findByText('Info about number field');
```
golcinho, line 1244:
```suggestion
    const result = screen.getByText('Info about number field');
```

packages/big-design/src/components/Worksheet/spec.tsx
bc-apostoliuk: Done

packages/big-design/src/components/Worksheet/spec.tsx
bc-apostoliuk: Done

golcinho (approved):

eugene-polev (approved):

bc-andreadao (approved):

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -66,29 +66,28 @@ const InternalHeaderCell = <T extends TableItem>({
     );
   };
 
-  // TODO: Update this. Return early if we don't have tooltip.
   const renderTooltip = () => {
-    if (typeof tooltip === 'string' && tooltip.length > 0) {
```
jorgemoya, line 71: Should we keep this check?

jorgemoya (commented):
Looking good! One thing is missing is to add the `tooltip` prop to the column types in the docs.

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
bc-apostoliuk: @jorgemoya 
Thanks for the comment 😊.
I don't think so. I guess the checking like `if(!tooltip) { return null }` looks completely sufficient. Why do you think we need to keep the previous condition? `typeof tooltip === 'string' && tooltip.length > 0`

bc-apostoliuk:
> Looking good! One thing is missing is to add the `tooltip` prop to the column types in the docs.

@jorgemoya Thanks, done 👍

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
chanceaclark: It's specifically for library consumers that don't use TypeScript. It adds a layer of protection in the component and I recommend keeping it and using type-guards for other components.

bc-apostoliuk: @jorgemoya @chanceaclark 👍 It makes sense. I've returned to the previous check.
Thanks)

jorgemoya (approved):

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
```diff
@@ -66,29 +66,28 @@ const InternalHeaderCell = <T extends TableItem>({
...
-                size="medium"
-                title="Hover or focus for additional context."
-              />
-            </Box>
-          }
-        >
-          {tooltip}
-        </Tooltip>
-      );
+    if (!(typeof tooltip === 'string' && tooltip.length > 0)) {
```
chanceaclark, line 70: 🍹 💅 Can we spread the negation or return it back to the original `if { return () }; return null;` route?
```suggestion
    if (typeof tooltip !== 'string' || !tooltip.length)) {
```

packages/big-design/src/components/Worksheet/Worksheet.tsx
```diff
@@ -89,20 +101,47 @@ const InternalWorksheet = typedMemo(
       }
     }, [invalidCells, onErrors, rows]);
 
+    const getRenderedTooltip = useCallback(
+      (tooltip?: string) => {
+        if (!(typeof tooltip === 'string' && tooltip.length > 0)) {
```
chanceaclark, line 106: 🍹 💅 Same here

chanceaclark (approved):

packages/big-design/src/components/Table/HeaderCell/HeaderCell.tsx
bc-apostoliuk: Yes,  I have returned it back to its original state.

packages/big-design/src/components/Worksheet/Worksheet.tsx
bc-apostoliuk: Done

PR #1112, opened 2023-01-25
chore: update repo to node v18

chanceaclark:
## What?

Updates the repos support node version to `v18.3.0`.

## Why?

Lesser node version end of life are coming up this year and we want to stay on top of LTS node versions.

N/A

## Testing/Proof

See CI runs.


packages/big-design/src/components/Datepicker/spec.tsx
```diff
@@ -65,7 +65,7 @@ test('no error when input date value manually', async () => {
     });
   });
 
-  expect(changeFunction).not.toHaveBeenCalled();
```
chanceaclark, line 68: IMO, not sure why this wasn't firing initial, but it should.

deini (approved):

MariaJose (approved):

jorgemoya (approved):

PR #1150, opened 2023-03-03
feat(component): add tooltip for Input and Select components

bc-apostoliuk:
## What?

Add a tooltip for `Input` and `Select` components.

## Why?

To be able to provide an extended description of the field.

Screenshot 2023-03-06 at 11 26 18
Screenshot 2023-03-06 at 11 11 41

## Testing/Proof

Locally + UT.

### DIFF
--- a/packages/big-design/src/components/Input/Input.tsx
+++ b/packages/big-design/src/components/Input/Input.tsx
@@ -1,3 +1,4 @@
+import { BaselineHelpIcon } from '@bigcommerce/big-design-icons';
 import React, {
   cloneElement,
   forwardRef,
@@ -10,9 +11,11 @@ import React, {
 } from 'react';
 
 import { typedMemo, warning } from '../../utils';
+import { Box } from '../Box';
 import { Chip, ChipProps } from '../Chip';
 import { FormControlDescription, FormControlLabel } from '../Form';
 import { useInputErrors } from '../Form/useInputErrors';
+import { Tooltip } from '../Tooltip';
 
 import { StyledIconWrapper, StyledInput, StyledInputContent, StyledInputWrapper } from './styled';
 
@@ -24,6 +27,7 @@ export interface Props {
   iconRight?: React.ReactNode;
   label?: React.ReactChild;
   labelId?: string;
+  tooltip?: string;
 }
 
 interface PrivateProps {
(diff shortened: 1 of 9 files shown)


packages/big-design/src/components/Input/Input.tsx
```diff
@@ -24,6 +27,7 @@ export interface Props {
   iconRight?: React.ReactNode;
   label?: React.ReactChild;
   labelId?: string;
+  tooltip?: string;
```
golcinho, line 30: - Should we allow to set `placement` for the tooltip? 
- Should we allow to set `trigger` for the tooltip?

According to all the above questions, should it be `string` or `TooltipProps & { text: string }`? 😅

golcinho (commented):
LGTML 👍 Good job! but, I have one conceptual question...

packages/big-design/src/components/Input/Input.tsx
golcinho: @eugene-polev Should we allow the possibility to use any icon instead of a hardcoded `question mark icon`?

vdm-kr (approved):
LGTM

packages/big-design/src/components/Input/Input.tsx
jorgemoya: Agree with @golcinho, we need to know figure out how configurable we want this to be.

golcinho: Looks like here we had almost the same case with a tooltip inside `th`: https://bigcommerce.github.io/big-design/table?props=columns

@vdm-kr Could you please confirm that the tooltip placement should be `right`? or should it be configurable?

vdm-kr: Just keep the question mark for the icon. 
The default position should be on the right. If possible, it should be automatically adjusted to left/top/bottom when it doesn't fit the screen in the default position.

golcinho (approved):

chanceaclark:
Why do we need to add this when we can do this with a `FormControlLabel`?
```html
<Input
  label={(
    <FormControlLabel>
      Label <Tooltip placement="right"
trigger={<InfoIcon size="medium" color="secondary50" />}>Info</Tooltip>
    </FormControlLabel>
  )}
  {...props}
/>
```

PR #1167, opened 2023-03-20
fix(component): fix extra triggering onError when new rows are passed

bc-apostoliuk:
## What?  Why?

Fix extra triggering onError when new rows are passed.

### Before fix
https://user-images.githubusercontent.com/84462142/226363462-ebcd6637-2299-4f7f-9d49-3418b3f159b0.mov

### After fix

https://user-images.githubusercontent.com/84462142/226364148-58dec5ae-4196-4074-bb78-d712a7e435db.mov

## Testing/Proof

Locally.


rtalvarez (commented):
In my opinion, the dependency array should remain exhaustive, so this feels kinda off. I would advise to try to find another way to fix this, but I'm curious what other people think

MariaJose:
> In my opinion, the dependency array should remain exhaustive, so this feels kinda off. I would advise to try to find another way to fix this, but I'm curious what other people think

Hey! I agree with Rob, would be good to try find the root case issue

bc-apostoliuk:
@rtalvarez @MariaJose Please, check it out one more time. I have changed the approach for a fix.

rtalvarez (approved):
thanks for looking for an alternate fix 🙇

PR #1173, opened 2023-03-22
feat(component): add cell selection with a keyboard

bc-apostoliuk:
## What? Why?

1. Implement the selection of multiple cells in a row using Shift+arrow buttons.
2. Implement the selection of multiple cells in a row within a column by holding SHIFT and clicking on the first and the last cells in the selection.

https://user-images.githubusercontent.com/84462142/226999149-65a2b457-abb1-4577-86af-8a2b7bf4e1e8.mov

## Testing/Proof
Locally + UT

### DIFF
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -66,11 +66,16 @@ const InternalCell = <T extends WorksheetItem>({
     useMemo(() => (state) => state.editWithValue, []),
   );
 
+  const isShiftPressed = useStore(
+    store,
+    useMemo(() => (state) => state.isShiftPressed, []),
+  );
+
   const isMetaKey = useStore(store, (state) => state.isMetaKey);
 
   const isControlKey = useStore(store, (state) => state.isControlKey);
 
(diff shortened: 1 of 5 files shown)


packages/big-design/jest.config.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
       statements: 95.73,
       branches: 86.97,
       functions: 97,
-      lines: 95.77,
+      lines: 95.76,
```
jorgemoya, line 15: This should be going up, not down.

bc-apostoliuk:
@jorgemoya thanks for noticing, I missed pushing this file 😣. I have already pushed. Check it out, please.

jorgemoya:
One other question I have is if this works horizontally or diagonally?

packages/big-design/jest.config.js
bc-apostoliuk: Done)

MariaJose:
Hey Lgtm!  just a question, I tried to select the cells vertically from top to bottom, but noticed that if I want to select them from bottom to top the cells don't get selected. Is this something would be good to be added it? 🍹  or only is gonna be from top to bottom?

bc-apostoliuk:
@MariaJose Hey, for now, it's gonna be from top to bottom.

PR #1179, opened 2023-03-28
feat(component): add custom rendering for text/number  cells (worksheet)

bc-apostoliuk:
## What?

Add custom renderer for `text` and number fields. (worksheet)

## Why?
To be able to provide a custom template within the cell if needed.
Screenshot 2023-03-28 at 22 29 09

Screenshot 2023-03-28 at 22 30 55

Screenshot 2023-03-28 at 19 54 38

## Testing/Proof

Locally + UT

### DIFF
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -21,6 +21,7 @@ interface CellProps<Item> extends TCell<Item> {
   options?: WorksheetSelectableColumn<Item>['config']['options'];
   rowId: number | string;
   formatting?: WorksheetTextColumn<Item>['formatting'];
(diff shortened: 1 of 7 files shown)


jorgemoya (approved):
Code looks good to me! Not represented on the screenshots, but I'm assuming that editing the cells will render the input but once the value is set it will go back to the rendered component. Please correct me if I'm wrong.

Obviously this will make anything to be rendered a possibility (for the text and number cells). Is @bigcommerce/product-design good with this approach?

MariaJose (approved):

chanceaclark:
I'm a little cautious about introducing custom rendering since the core part of the library is enforcing design consistency. If this is the direction we want to go, why don't we change our current error border approach to this design instead of enabling custom rendering?

PR #1193, opened 2023-04-11
feat(component): add notation for worksheet cells

bc-apostoliuk:
## What?

Add notation for worksheet cells

- It should be applicable to a cell of any type.
- The indicator color should be customizable within the BigDesign color palette.
- The message should be customizable.
- The indicator position is always the same – the top right corner.
- The tooltip position by default should be on the right of the indicator and automatically adjusted if it doesn’t fit the screen.

## Why?

To be able to provide a way to show additional notes/comments/instructions on a particular cell level.

https://user-images.githubusercontent.com/84462142/231280748-fc82a7ce-c25c-4d72-8b73-2cb0a3cfb537.mov

## Testing/Proof

Locally + UT

### DIFF
--- a/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
+++ b/packages/big-design/src/components/Worksheet/Cell/Cell.tsx
@@ -1,6 +1,7 @@
-import React, { useCallback, useEffect, useMemo } from 'react';
+import React, { useCallback, useEffect, useId, useMemo } from 'react';
 
 import { typedMemo } from '../../../utils';
+import { Tooltip } from '../../Tooltip';
 import { Small } from '../../Typography';
 import { CheckboxEditor, ModalEditor, SelectEditor, TextEditor, ToggleEditor } from '../editors';
 import { useAutoFilling, useEditableCell, useWorksheetStore } from '../hooks';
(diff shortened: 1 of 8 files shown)


packages/big-design/src/components/Worksheet/Cell/Cell.tsx
```diff
@@ -248,6 +257,26 @@ const InternalCell = <T extends WorksheetItem>({
     renderedValue,
   ]);
 
+  const renderedNote = useMemo(() => {
+    if (!notation) {
+      return null;
+    }
+
+    const note = notation(value, row);
```
MariaJose, line 265: Should we also check if note is `undefined` and return `null`? 🍹 

```tsx
 if (!note) {
   return null
 }

return (
    <Tooltip>
        ....
    </Tooltip>
)
```

packages/big-design/src/components/Worksheet/spec.tsx
```diff
@@ -561,6 +567,25 @@ describe('formatting', () => {
   });
 });
 
+describe('nitation', () => {
```
MariaJose, line 570: `notation` ?

MariaJose (approved):
Lgtm, just few suggestions

packages/big-design/src/components/Worksheet/Cell/Cell.tsx
bc-apostoliuk: I think it will not be redundant to avoid returning undefined. So I`ve updated the code according to your suggestion.

packages/big-design/src/components/Worksheet/spec.tsx
bc-apostoliuk: Thanks, done

vdm-kr (approved):
LGTM

jorgemoya (approved):
Looks great!

PR #1241, opened 2023-06-26
feat(component): add enabled prop to within the columns entity (worksheet)

bc-apostoliuk:
## What?

Add enabled prop to within the columns entity (worksheet).

Fix the prop naming in the docs (`disable` -> `disabled`)

## Why?

To allow the enablement of the columns for the case when the rows are disabled.

Screenshot 2023-06-26 at 18 14 43

## Testing/Proof

Locally.


bc-apostoliuk:
@chanceaclark 
Currently, if the row is disabled all cells within this row are disabled. And if we pass `disabled: false for some column it will not enable this column, since the full row is disabled.
Actually, we can rework the current logic of the `disabled` field of the column and handle the `disabled: false` as the highest priority over the disabled row. I have created the additional property to avoid confusion.

What do you think, do we need to set the highest priority for the `disabled` prop within the column, and handle `disabled: false` like this column always should be available independently of disabled rows?

bc-andreadao (approved):

chanceaclark (approved):

PR #1284, opened 2023-08-08
fix(icons): downgrade 'camelcase' to 6.3.0

bc-apostoliuk:
## What? Why?

Fix icons build.

## Testing/Proof
Locally.

ping @bigcommerce/inventory-team


packages/big-design-icons/package.json
```diff
@@ -57,7 +57,7 @@
     "@types/react-dom": "^18.0.6",
     "@types/styled-components": "^5.1.26",
     "babel-plugin-styled-components": "^2.0.7",
-    "camelcase": "^7.0.1",
+    "camelcase": "^6.3.0",
```
chanceaclark, line 60: Could we also add this to the [dependabot ignore config](https://github.com/bigcommerce-labs/ng-cp2-control-panel/blob/master/.github/dependabot.yml#L22)? We will need to add a few others due to not being able to support ESM yet.

chanceaclark (approved):
One small non-blocking comment, but would love if you could add it 🙏 Thanks for doing this!

packages/big-design-icons/package.json
bc-apostoliuk: @chanceaclark  Yes, good point. I have added it, check it out, please 
Thanks.

bc-apostoliuk: Done

PR #1303, opened 2023-10-09
utility button variant and new icons

deini:
## What?
- Adds a new `utility` variant to buttons.
- Adds `WindowExpand` and `WindowMinimize` icons

Note: These are non-standard icons provided by design.

## Why?
Design is introducing this new button variant, we've been using it enough for it to become a proper pattern.

PXGL8WWH


chanceaclark (approved):
Looks good to me. Only callout is we should audit some of the components using `StyleableButton` or `MessagingButton` and see if we can replace it with this utility button.

PR #1343, opened 2024-02-13
feat(component): add icon prop for FileUploader drop-zone

bc-apostoliuk:
## What?

Add icon props for FileUploader drop-zone.

## Why?

To be able to set custom icon.

Screenshot 2024-02-13 at 12 46 48

## Testing/Proof

Locally.

### DIFF
--- a/packages/big-design/src/components/FileUploader/DropZone.tsx
+++ b/packages/big-design/src/components/FileUploader/DropZone.tsx
@@ -26,8 +26,9 @@ export interface DropZoneLocalization {
 }
 
 interface Props extends InputHTMLAttributes<HTMLInputElement> {
-  label?: string;
   description?: string;
+  icon?: React.ReactNode;
+  label?: string;
   localization?: DropZoneLocalization;
   onFilesChange(files: FileList | null): void;
 }
@@ -36,6 +37,7 @@ export const DropZone = ({
   accept,
   description,
   disabled,
+  icon = <DraftIcon />,
   id,
   label,
   localization = defaultLocalization,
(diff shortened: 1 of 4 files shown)


packages/big-design/src/components/FileUploader/DropZone.tsx
```diff
@@ -1,8 +1,5 @@
-import {
-  DraftIcon,
-  FileDownloadIcon,
-  RemoveCircleOutlineIcon,
-} from '@bigcommerce/big-design-icons';
+import { FileDownloadIcon, RemoveCircleOutlineIcon } from '@bigcommerce/big-design-icons';
+import * as icons from '@bigcommerce/big-design-icons';
```
chanceaclark, line 2: This ends up importing all the icons which prevents them from being tree-shaked. Since that increases the bundle size for this component, I recommend just having a prop that consumes an icon instead (see [`Button` component](https://github.com/bigcommerce/big-design/blob/main/packages/big-design/src/components/Button/Button.tsx#L15)).

bc-apostoliuk: Done

packages/docs/yarn-error.log
chanceaclark, line 1: Can we remove this file?

bc-apostoliuk: my bad)
removed ✅

golcinho (approved):

chanceaclark (approved):

jorgemoya (approved):

PR #1344, opened 2024-02-16
feat(component): extend FileUploader with new errors view, custom file actions

bc-apostoliuk:
## What?

- Update errors displaying view;
- Add custom action for files;
 
## Why?

According to the updated design.

Screenshot 2024-02-16 at 11 17 08

Screenshot 2024-02-16 at 11 19 05

## Testing/Proof

Locally.


packages/big-design/src/components/FileUploader/File.tsx
```diff
@@ -1,18 +1,22 @@
...
 
 import { Button } from '../Button';
+import { Dropdown } from '../Dropdown';
+import { DropdownItem } from '../Dropdown/types';
 
 import { FileStyled, ImageStyled, TextEllipsed } from './styled';
 
 interface Props extends HTMLAttributes<HTMLDivElement> {
-  name: string;
+  actions?: DropdownItem[];
```
chanceaclark, line 11: Do you think we would need other actions than removing the file? If we don't, can we just handle this for the user instead?

chanceaclark (commented):
I'm curious why we are using an `InlineMessage` to display errors in a form component. Can we reach back out to design to use our existing form error approach with [`FormGroup`'s](https://github.com/bigcommerce/big-design/blob/main/packages/big-design/src/components/Form/Group/Group.tsx) + an error prop? It would allow a more flexible validation for the consumer of this component.

packages/big-design/src/components/FileUploader/File.tsx
bc-apostoliuk: @vdm-kr, could you please explain, why we decided to use `InlineMessage` for form component, and do we actually need additional `actions` other than deletion?

chanceaclark: My comment on using `InlineMessage` still stands, @vdm-kr can we update the designs to go with our current pattern of displaying form error messages?

vdm-kr: Same for the error message since it could be used not only in form and files could be rendered separately we need to display errors in a more universal way. 
Errors

chanceaclark: @vdm-kr the `InlineMessage` component was only meant to be used within `Modal`'s. I remember we worked with design to come up with that guideline:
Screenshot 2024-03-05 at 09 36 25

vdm-kr: Screenshot 2024-03-11 at 08 36 26
In the guidelines below we do allow to place it inside panels. I also discussed it one more time with David, Eugene, and Turner, they confirmed

chanceaclark: We probably need to update that description then if they can be used in both. Now that I think about it, we have used it in both places.

@vdm-kr The big issue I see here is that now we have two different ways of showing errors within form controls. Since a `FileUploader` is a type of form control, those will need to work within forms. `InlineMessage` components were _originally_ meant to be used at the top of modals as a more condensed way of showing a message. Now that they have been expanded to be used within pages, there are now two different ways of showing errors within forms. As a pattern library we are trying to enforce a consistency for both our merchants and our developers (including 3rd party devs). So as a developer, trying to enforce a consistent design pattern, what should we use for showing form validation issues now? `InlineMessage` or the errors underneath the form fields? I know there's probably use-cases for one or the other, but this will make a confusing pattern for consumers of this.

Screenshot 2024-03-14 at 12 01 51

vdm-kr: Let’s update description to “Used for displaying alerts within Panels and Modals. Is a condensed version of the Message component.”

I think it’s okay to have both since fields have different behaviour. File uploader doesn’t look and work in the same way as a text field. Also, this error states are embedded in the component so I don’t really think there will be any issues because developers won’t need to decide how to display errors.

chanceaclark (approved):

jorgemoya (approved):

PR #1372, opened 2024-03-29
refactor: use ComponentPropsWithoutRef for types

chanceaclark:
## What?

This changes our hand rolled element types to React's built-in `ComponentTypesWithoutRef` type. I also aligned the imports from the type to come from the `react` module rather than consuming it from the `React.` namespace.

_Note: Also ran `npx yarn-deduplicate` on the lockfile (hence a lot of deleted lines)._

## Why?

There are some discrepancies between hand rolling types vs. using this built-in type. This ensures that the underlying element has the exact types it actually needs. 

N/A

## Testing/Proof

See CI for builds


packages/big-design/src/components/List/GroupHeader/styled.tsx
```diff
@@ -1,7 +1,7 @@
 import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import styled from 'styled-components';
 
-export const StyledGroupHeader = styled.li<React.HTMLProps<HTMLLIElement>>`
```
chanceaclark, line 4: This was duplicative for the type, so just omitted it.

packages/big-design/src/components/MultiSelect/types.ts
```diff
@@ -1,12 +1,12 @@
...
-import React, { RefObject } from 'react';
+import React, { ComponentPropsWithoutRef, RefObject } from 'react';
 
 import { InputProps } from '../Input';
 import { InputLocalization } from '../Input/Input';
 import { SelectAction, SelectOption } from '../Select';
 import { SelectOptionGroup } from '../Select/types';
 
-interface BaseSelect extends Omit<React.HTMLAttributes<HTMLInputElement>, 'children'> {
+interface BaseSelect extends Omit<ComponentPropsWithoutRef<'input'>, 'children' | 'value'> {
```
chanceaclark, line 9: This was the only type I actually had to change due to a type issue. We are using our own custom `value` prop rather than the intrinsic input element `value` attribute so we needed to omit it from the types.

jorgemoya (approved):
🏆

packages/big-design/src/components/Worksheet/__snapshots__/spec.tsx.snap
chanceaclark, line 1: This snapshot changed because we changed the file contents of https://github.com/bigcommerce/big-design/pull/1372/files#diff-f5a5f06a3778e6700a2aebb013af6c7089cefb1482f7c1a3170da13366271e2aR9

The babel plugin hashes the file content: https://github.com/styled-components/babel-plugin-styled-components/blob/4e2eb388d9c90f2921c306c760657d059d01a518/src/visitors/displayNameAndId.js#L182

PR #1398, opened 2024-04-09
fix: typescript build output with wrong imports

chanceaclark:
## What?

Updates typescript to the latest version to fix the build outputs.

Typescript Issue: https://github.com/microsoft/TypeScript/issues/57956

## Why?

Types we using absolute paths, which was causing the types to be `any`.

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>
	<td>
</table>

### DIFF
--- a/packages/configs/package.json
+++ b/packages/configs/package.json
@@ -7,6 +7,6 @@
     "@typescript-eslint/parser": "^7.4.0",
     "eslint": "^8.33.0",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "typescript": "^5.3.3"
+    "typescript": "^5.4.4"
   }
 }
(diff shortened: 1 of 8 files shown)


deini (approved):

packages/configs/package.json
```diff
@@ -7,6 +7,6 @@
     "@typescript-eslint/parser": "^7.4.0",
     "eslint": "^8.33.0",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "typescript": "^5.3.3"
+    "typescript": "^5.3.4"
```
deini, line 10: 5.4.4*

deini (approved):

PR #1401, opened 2024-04-11
feat: export types without reassignment

chanceaclark:
## What?

Exports the types without reassignment to new type definitions.

## Why?

Previously, we were reassigning the types due to tree-shaking issues. However, this is no longer with later versions of typescript, so this reassignment is not longer needed.

Screenshot 2024-04-10 at 15 33 20

### DIFF
--- a/packages/big-design/src/components/FileUploader/types.ts
+++ b/packages/big-design/src/components/FileUploader/types.ts
@@ -1,0 +1,13 @@
+export interface Localization {
+  upload: string;
+  optional: string;
+  showLess: string;
+  showMore: string;
+  canNotUploadTheseFiles: string;
+}
+
+export interface ValidatorConfig {
+  message?: string;
+  type: string;
+  validator: (file: File) => boolean;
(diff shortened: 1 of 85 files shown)


packages/big-design/src/components/FileUploader/types.ts
```diff
@@ -0,0 +1,13 @@
+export interface Localization {
```
chanceaclark, line 1: All the `FileUploader` changes where just shifting things around, simplifying types (one `Localization` type), and removing type exports that weren't needed. Confirmed with @bc-apostoliuk which types were needed via export.

packages/big-design/src/components/Form/__snapshots__/spec.tsx.snap
chanceaclark, line 1: This was changed due to the hash of `packages/big-design/src/components/Form/Label/styled.tsx` changing

jorgemoya (approved):

PR #1402, opened 2024-04-11
chore(ci): install package managers in ~/bin folder

chanceaclark:
## What?

Installs any corepack installed package managers in a folder with permissions.

## Why?

Running into a `Internal Error: EACCES: permission denied, unlink '/usr/local/bin/pnpm'` when activating pnpm.
https://app.circleci.com/pipelines/github/bigcommerce/big-design/6077/workflows/469edca2-53d5-41ec-a6d0-75e629711561/jobs/28248

N/A

## Testing/Proof

See CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ commands:
       - run:
           name: Install pnpm package manager
           command: |
-            corepack enable
+            corepack enable --install-directory ~/bin
             corepack prepare pnpm@latest-8 --activate
             pnpm config set store-dir .pnpm-store
       - run:


.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ commands:
       - run:
           name: Install pnpm package manager
           command: |
-            corepack enable
+            corepack enable --install-directory ~/bin
```
tekstrand, line 32: Will this work if that directory doesn't exist?

chanceaclark: Yeah, it'll create it if it doesn't exist. We just don't have access to `/<os>` folders, but we should have access to the home directory.

tekstrand (approved):

PR #1403, opened 2024-04-12
refactor(docs): update side-nav to use radix scrollbar

chanceaclark:
## What?

Updates the side nav on the docs side to use a radix-ui scroll area. Alongside that, I had to adjust a few layout bits in order to accommodate this new component. This helped clean up a few bits of legacy code that we should've removed a while ago. 

Some of the layout changes:
- Added more width to the content and side nav
- Removed helpful links and hoisted the Github link up since I know it's used a lot.

## Why?

The scroll bar was continuously present which was a bit annoying. Using radix allows it to auto-hide and customize the scroll bar with cross-browser support. Increasing the width of the content and side-nav also helps make the docs look nicer.

### Before
Screenshot 2024-04-12 at 16 25 25

### After

https://github.com/bigcommerce/big-design/assets/10539418/d9152fe1-9cab-4f64-8875-979ef581a6b9

### DIFF
--- a/packages/docs/components/SideNav/SideNavLogo/SideNavLogo.tsx
+++ b/packages/docs/components/SideNav/SideNavLogo/SideNavLogo.tsx
@@ -1,13 +1,16 @@
 import { Small } from '@bigcommerce/big-design';
+import Image from 'next/image';
 import Link from 'next/link';
 import React from 'react';
+
+import Logo from '../../../public/logo-with-text.svg';
 
 import { StyledLogo } from './styled';
 
 export const SideNavLogo: React.FC = () => (
-  <Link href="/">
+  <Link href="/" style={{ textDecoration: 'none' }}>
     <StyledLogo>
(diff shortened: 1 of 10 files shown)


packages/docs/components/SideNav/SideNavLogo/SideNavLogo.tsx
```diff
@@ -1,13 +1,16 @@
 import { Small } from '@bigcommerce/big-design';
+import Image from 'next/image';
 import Link from 'next/link';
 import React from 'react';
 
+import Logo from '../../../public/logo-with-text.svg';
```
deini, line 6: 🍹 can this use `~/`?

deini (approved):

packages/docs/components/SideNav/SideNavLogo/SideNavLogo.tsx
chanceaclark: it'll be a hassle. either have to combat eslint or do some tsconfig `paths` magic.

PR #1408, opened 2024-04-16
chore(deps): bump deps to latest pinned versions

chanceaclark:
## What?

Bumps deps to the latest pinned versions.

## Why?

Make sure everything is mostly up-to-date.

N/A

## Testing/Proof

See CI


packages/big-design-icons/package.json
```diff
@@ -42,35 +42,35 @@
...
     "@babel/preset-typescript": "^7.15.0",
-    "@bigcommerce/configs": "^0.20.1",
+    "@bigcommerce/configs": "workspace:^",
     "@styled/typescript-styled-plugin": "^1.0.1",
     "@svgr/core": "^8.1.0",
     "@svgr/plugin-jsx": "^8.1.0",
     "@svgr/plugin-prettier": "^8.1.0",
     "@svgr/plugin-svgo": "^8.1.0",
     "@types/react": "^18.2.73",
     "@types/react-dom": "^18.2.23",
```
chanceaclark, line 58: There's an issue with the latest `@types/react` + `@types/react-dom` that doesn't mesh well with `@types/styled-components`. For some reason `0n` is being passed into as a falsey value, which may be related with [this lodash type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/80f7e6bdd99074bf98ce4c00a25d2ffc949d71f7/types/lodash/common/array.d.ts#L37). I'm not entirely sure, but I avoided it for now.

packages/big-design/package.json
```diff
@@ -32,61 +32,61 @@
...
     "@popperjs/core": "^2.11.6",
     "@types/react-datepicker": "^6.2.0",
     "date-fns": "3.6.0",
     "downshift": "9.0.4",
-    "focus-trap": "^7.0.0",
-    "polished": "^4.0.0",
+    "focus-trap": "^7.5.4",
+    "polished": "^4.3.1",
     "react-beautiful-dnd": "^13.1.1",
     "react-datepicker": "^6.6.0",
```
chanceaclark, line 45: I also couldn't bump this to the latest as they added an [import from `date-fns`](https://github.com/Hacker0x01/react-datepicker/compare/v6.6.0...v6.8.0#diff-c5ae56d754e9b921eb7696fed16bdde4b1548e7786fb0edb59ac711e6c378041R16) that causes the next.js build to break. I think we need to look at getting rid of the custom babel build in our docs site.

chanceaclark (commented):
Dependencies updates caused the snapshots to regenerate new hashes.

matthewvolk (approved):
🥳

PR #1424, opened 2024-05-01
docs(release): document process for updating snapshots

PascalZajac:
## What?
Add instructions on how to update snapshots when needed.

## Why?
I forget literally every time and spend 30 minutes searching my shell history to remember what commands I need to run and where.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,14 @@ Run tests with:
 pnpm run test
 ```
 
+Update snapshots with:
+
+```
+pnpm build
+cd packages/big-design
+pnpm run test --update-snapshot
+```
+
 Run linter with:
 
 ```


README.md
```diff
@@ -92,6 +92,14 @@ Run tests with:
 pnpm run test
 ```
 
+Update snapshots with:
+
+```
+pnpm build
+cd packages/big-design
+npx jest --update-snapshot
```
chanceaclark, line 100: Since our docs reference pnpm, let's use that instead of npm (I believe the `--` is needed but not entirely confident).
```suggestion
pnpm run test -- --update-snapshot
```

PascalZajac: Tested (the middle `--` was unnecessary) and updated.

chanceaclark (approved):

PR #1429, opened 2024-05-12
fix(component): remove excess margin from contentless Panel

leeBigCommerce:
## What?

Currently a contentless (or `children`-less) `Panel` component renders with excessive margin under its header.
This PR removes that extra margin.

## Why?

I am not sure if a contentless use of `Panel` was originally envisioned, but the prop is optional and design are looking to use it for 'placeholders'. For example, the `Shipping` panel on Manual Orders remains empty until preceding mandatory sections are filled.

### Before (left) After (right) 

Note: only the final variant is affected by this change

Screenshot 2024-05-17 at 1 30 05 PM

## Testing/Proof

- Screenshots from documentation app above
- Passing existing tests, proving no other scenarios were affected
- New snapshot test to defend the new conditional styling

### DIFF
--- a/packages/big-design/src/components/Panel/Panel.tsx
+++ b/packages/big-design/src/components/Panel/Panel.tsx
@@ -4,6 +4,7 @@ import { MarginProps } from '../../helpers';
 import { excludePaddingProps } from '../../helpers/paddings/paddings';
 import { warning } from '../../utils';
 import { Badge, BadgeProps } from '../Badge/Badge';
+import { Box } from '../Box';
 import { Button, ButtonProps } from '../Button';
 import { Flex } from '../Flex';
 import { Text } from '../Typography';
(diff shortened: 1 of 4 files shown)


packages/big-design/src/components/Panel/Panel.tsx
```diff
@@ -40,10 +40,12 @@ export const RawPanel: React.FC<PanelProps & PrivateProps> = memo(({ forwardedRe
       return null;
     }
 
+    const headerBottomMargin = description ? 'xxSmall' : props.children ? 'medium' : 'none';
```
leeBigCommerce, line 43: I don't love a nested ternary, but this is just an A or B or C evaluation.

Happy to move it to a `let` and `if` combo (or a better recommendation) if people prefer.

icatalina: 🤔 I wonder if something like `gap` will work better in this scenario

jorgemoya: I support `gap` as well but this is fine, up to you 🍹 . In hindsight it is what I would use but `gap` wasn't readily available yet.

leeBigCommerce: `gap` might be over egging it 🤷 

We just have the one bit of spacing here, albeit with multiple possible values
Whereas a `gap` infers a more complex, multi row/column layout, where the value is intended to create multiple spaces

or it could just be that I'm old and don't like new CSS 😆

leeBigCommerce: @icatalina I've gone for the approach you mentioned here: https://github.com/bigcommerce/big-design/pull/1429/commits/58c3a28b615f316e4bfacbe26b4436204205d85d

icatalina (approved):

jorgemoya (approved):

davelinke:
Less javaScript more CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/:empty

PR #1456, opened 2024-06-25
feat(component):  implement `select all` for multi-select component

bc-apostoliuk:
## What?
Implement `select all` for multi-select component;

## Why?

To be able to select/unselect all options by one click.

https://github.com/bigcommerce/big-design/assets/84462142/a8df07b7-0bd7-447a-a369-0c9aa48fcb99

## Testing/Proof

Locally + UT.

### DIFF
--- a/packages/big-design/src/components/List/List.tsx
+++ b/packages/big-design/src/components/List/List.tsx
@@ -15,6 +15,7 @@ import { useIsomorphicLayoutEffect, useWindowSize } from '../../hooks';
 import { typedMemo } from '../../utils';
 import { Box } from '../Box';
 import { DropdownItem, DropdownItemGroup, DropdownLinkItem, DropdownProps } from '../Dropdown';
+import { MultiSelectLocalization } from '../MultiSelect/types';
 import { SelectAction, SelectOption, SelectOptionGroup, SelectProps } from '../Select';
 
 import { ListGroupHeader } from './GroupHeader';
@@ -43,6 +44,7 @@ export interface ListProps<T> extends ComponentPropsWithoutRef<'ul'> {
     | UseSelectPropGetters<any>['getMenuProps']
     | UseComboboxPropGetters<any>['getMenuProps'];
   update: (() => Promise<Partial<State>>) | null;
+  localization?: { selectAll: MultiSelectLocalization['selectAll'] };
   removeItem?(item: SelectOption<T>): void;
 }
 
@@ -69,6 +71,7 @@ const StyleableList = typedMemo(
     selectedItems,
     selectAll,
     update,
+    localization = { selectAll: 'Select All' },
     removeItem,
     ...props
   }: ListProps<T> & PrivateProps): ReturnType<React.FC<ListProps<T> & PrivateProps>> => {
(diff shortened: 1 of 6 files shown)


MichalDobry (approved):

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -280,7 +286,13 @@ export const MultiSelect = typedMemo(
...
       inputValue,
       itemToString: (option) => (option ? option.content : ''),
-      items: filteredOptions,
+      items: [
+        ...(selectAll
+          ? // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+            ([{ content: localization?.selectAll, value: 'select-all' }] as Array<SelectOption<T>>)
+          : []),
+        ...filteredOptions,
+      ],
```
icatalina, line 295: 🤔 I see why you need the `as` here, but I wonder if we should actually keep `value` as a generic and rather just type it as `string`.

I'm not sure it makes sense for value to be anything other than `string`, someone with more expertise in BD might correct me.

cc: @jorgemoya @chanceaclark

packages/big-design/src/components/List/List.tsx
```diff
@@ -83,6 +90,61 @@ const StyleableList = typedMemo(
       scheduleUpdate();
     }, [isOpen, height, width, selectedItems?.length]);
 
+    const handleSelectAll = useCallback(
+      (listItems: Array<SelectOption<T>>) => {
+        const enabledItems = listItems.filter((item) => !item.disabled);
+
+        updateItems && updateItems(enabledItems);
```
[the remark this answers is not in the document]
bc-apostoliuk: Make sense, Done 👍

packages/big-design/src/components/List/List.tsx
```diff
@@ -83,6 +90,63 @@ const StyleableList = typedMemo(
...
+          const enabledItems = listItems.filter((item) => !item.disabled);
+
+          updateItems(enabledItems);
+        }
+      },
+      [updateItems],
+    );
+
+    const handleUnselectAll = useCallback(() => {
+      updateItems && updateItems([]);
```
icatalina, line 105: the comment applied to all cases where we had an expression that wasn't really used. eg `a && a()` not just the one I pointed out

bc-apostoliuk: My bad, missed that

Already fixed. Thanks.

packages/big-design/src/components/MultiSelect/MultiSelect.tsx
jorgemoya: I agree we should change this to string, I believe the original idea of having value as a generic is flawed, but maybe something we fix in a different PR.

packages/big-design/src/components/MultiSelect/spec.tsx
```diff
@@ -1085,6 +1085,41 @@ test('renders localized labels', async () => {
   expect(label).toHaveStyleRule('content', "' (opcional)'", { modifier: '::after' });
 });
 
+test('selects all triggers onOptionsChange with available options', async () => {
```
jorgemoya, line 1088: Can you add a test for indeterminate and unselect all?

bc-apostoliuk: Yes, added two more tests to cover these cases

jorgemoya (approved):

PR #1490, opened 2024-08-15
chore: use turborepo instead of lerna + nx

chanceaclark:
## What?

Switches the repo over to using `turbo` instead of `lerna`'s built-in `nx` caching functionality. Also updates CI to use the caching benefit using remote caching.

## Why?

This is part of the path to removing lerna from the repo.

N/A

## Testing/Proof
Latest test run in CI is using Turbo Remote Caching.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,7 +68,7 @@ jobs:
           keys:
             - big-design-icons-<<parameters.cache-version>>-{{ .Branch }}
             - big-design-icons-<<parameters.cache-version>>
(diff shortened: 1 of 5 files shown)


PascalZajac:
Does this change only concern the CI build, or also local development?

golcinho (approved):

chanceaclark:
>Does this change only concern the CI build, or also local development?

Both. It'll cache local scripts that are ran, remote cache is completely optional locally.

PR #1523, opened 2024-08-28
docs: add header pattern documentation

davelinke:
## What?

Adding the header pattern component documentation

## Why?

To guide developers on how to use the header component

header-docs

### DIFF
--- a/packages/docs/PropTables/HeaderPropTable.tsx
+++ b/packages/docs/PropTables/HeaderPropTable.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { Prop, PropTable, PropTableWrapper } from '../components';
+import { NextLink, Prop, PropTable, PropTableWrapper } from '../components';
 
 const headerProps: Prop[] = [
   {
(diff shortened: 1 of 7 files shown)


packages/docs/PropTables/HeaderPropTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+    required: true,
+  },
+  {
+    name: 'actions',
+    types: 'Array<ActionButtonProps | ActionDropdownProps>',
+    description: 'An array of actions to display in the header. Can include buttons or dropdowns.',
+  },
+  {
+    name: 'backLink',
+    types: 'BackLinkProps',
```
chanceaclark, line 19: I think we need to add `BackLinkProps` to the tabs

packages/docs/pages/header.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <Panel header="Implementation" headerId="implementation">
+        <Fragment key="basic">
+          <Text>
+            To use <Code primary>Header</Code> import the component from the package:
+          </Text>
+          <CodeSnippet>{`import { Header } from '@bigcommerce/big-design-patterns';`}</CodeSnippet>
+        </Fragment>
+        <CodePreview>
+          {/* jsx-to-string:start */}
+          {function Example() {
```
chanceaclark, line 47: We don't really need this function. We can just let it be `<Header ... />`

golcinho (approved):

packages/docs/PropTables/HeaderPropTable.tsx
```diff
@@ -11,34 +11,59 @@ const headerProps: Prop[] = [
...
+        <NextLink href={{ hash: 'action-button-prop-table', query: { props: 'action-button' } }}>
+          ActionButton
+        </NextLink>
+        {' | '}
+        <NextLink
+          href={{ hash: 'action-dropdown-prop-table', query: { props: 'action-dropdown' } }}
+        >
+          ActionDropdown
+        </NextLink>
+        {'>'}
```
chanceaclark, line 26: We actually haven't done this before, but I figured it was a better approach than our previous iterations.

packages/docs/PropTables/HeaderPropTable.tsx
```diff
@@ -11,34 +11,59 @@ const headerProps: Prop[] = [
...
     description: 'An array of actions to display in the header. Can include buttons or dropdowns.',
   },
   {
     name: 'backLink',
-    types: 'BackLinkProps',
+    types: (
+      <NextLink href={{ hash: 'backlink-prop-table', query: { props: 'back-link' } }}>
+        BackLink
+      </NextLink>
+    ),
```
chanceaclark, line 37: If we are referring to another prop, we need to add links to the prop table.

packages/docs/PropTables/HeaderPropTable.tsx
```diff
@@ -11,34 +11,59 @@ const headerProps: Prop[] = [
...
   {
     name: 'icon',
     types: 'ReactNode',
     description: 'An icon to display to the left of the title.',
   },
-  {
-    name: 'children',
-    types: 'ReactNode',
-    description: 'Any additional content to render inside the header.',
-  },
```
chanceaclark: `children` is not actually passed anywhere in the component so I think we'll need to remove the `PropsWithChildren` on the type.

packages/docs/pages/header.tsx
```diff
@@ -85,24 +86,34 @@ const HeaderPage = () => {
...
             },
             {
-              id: 'action-button-props',
+              id: 'action-button',
               title: 'ActionButton',
-              render: () => <ActionButtonPropsTable />,
+              render: () => (
+                <ActionButtonPropsTable
+                  id="action-button-prop-table"
+                  inheritedProps={<ButtonPropTable collapsible />}
```
chanceaclark, line 99: We like to use `inheritedProps` to showcase they are extended. The only issue is that we can't omit things from these tables. How we do props in the documentation needs to be revamped which can be for the future.

packages/docs/pages/header.tsx
```diff
@@ -111,14 +122,14 @@ const HeaderPage = () => {
       <Panel header="Do's and Don'ts" headerId="guidelines">
         <GuidelinesTable
           discouraged={[
-            <>Don’t overload the header with too many actions or content.</>,
-            <>Don’t use more than one primary action button.</>,
-            <>Avoid unrelated or non-contextual content in the header.</>,
+            'Don’t overload the header with too many actions or content.',
```
chanceaclark, line 125: Use strings instead of fragments where possible.

bc-apostoliuk (approved):

PR #1525, opened 2024-08-30
refactor(patterns): update page + actionbar to work outside iframes

chanceaclark:
## What/Why?

Refactors the Page and ActionBar components to be able to work within iframeless contexts.

Shout out to @deini for the CSS approach here 🙏 

## Screenshots/Screen Recordings

Screenshot 2024-08-30 at 14 09 51

## Testing/Proof

Reran the tests and cover and nothing is off.

PR #1535, opened 2024-09-12
fix(patterns): adjust flex layout for header

chanceaclark:
## What/Why?

When adding support for iframes #1525, we introduced a bug where the header doesn't wrap correctly. This fixes that issue.

### Before

Screenshot 2024-09-12 at 17 00 56

### After

Screenshot 2024-09-12 at 16 59 46

## Testing/Proof

Local testing proof.


davelinke (approved):
Tested positively. Thanks

deini (approved):

PR #1566, opened 2024-10-16
feat(component): Add action props to FileUploader

bc-apostoliuk:
## What?

Add `action` prop to FileUploader.

## Why?

To be able to pass the custom action for `FileUploader` component.

Screenshot 2024-10-16 at 15 57 29

Screenshot 2024-10-16 at 15 57 57

## Testing/Proof

Locally + UT.

### DIFF
--- a/packages/big-design/src/components/FileUploader/DropZone.tsx
+++ b/packages/big-design/src/components/FileUploader/DropZone.tsx
@@ -13,15 +13,17 @@ import React, {
   useState,
 } from 'react';
 
+import { excludeMarginProps } from '../../helpers';
 import { Box } from '../Box';
 import { Flex } from '../Flex';
 import { Small, Text } from '../Typography';
 
 import { StyledButton, StyledDropzone } from './styled';
-import { Localization } from './types';
+import { Action, Localization } from './types';
 import { validateFileFormat } from './utils';
 
 export interface Props extends ComponentPropsWithoutRef<'input'> {
+  action?: Action;
   description?: string;
   viewType?: 'row' | 'block';
   icon?: React.ReactNode;
@@ -32,6 +34,7 @@ export interface Props extends ComponentPropsWithoutRef<'input'> {
 
 export const DropZone = ({
   accept,
+  action,
   description,
   disabled,
   icon = <DraftIcon />,
@@ -149,6 +152,34 @@ export const DropZone = ({
     );
   }, [description]);
 
+  const renderedAction = useMemo(() => {
+    if (!action) {
+      return null;
+    }
+
+    const { label, onClick, style, ...actionProps } = action;
+
+    return (
+      <StyledButton
+        {...excludeMarginProps(actionProps)}
+        color="secondary"
+        marginTop={isRowView ? 'none' : 'medium'}
+        onClick={(event) => {
+          event.preventDefault();
+          event.stopPropagation();
+
+          if (onClick) {
+            onClick(event);
+          }
+        }}
+        type="button"
+        variant="subtle"
+      >
+        {label}
+      </StyledButton>
+    );
+  }, [action, isRowView]);
+
   /* istanbul ignore next */
   return (
     <StyledDropzone
@@ -191,15 +222,19 @@ export const DropZone = ({
               {renderedDescription}
             </Flex>
           </Flex>
-          <StyledButton
-            color="secondary"
-            disabled={disabled}
-            marginTop={isRowView ? 'none' : 'medium'}
-            onClick={(e) => e.preventDefault()}
-            variant="subtle"
-          >
-            {localization.upload}
-          </StyledButton>
+          <Flex flexDirection="row">
+            {renderedAction}
+            <StyledButton
+              color="secondary"
+              disabled={disabled}
+              marginTop={isRowView ? 'none' : 'medium'}
+              onClick={(e) => e.preventDefault()}
+              type="button"
+              variant="subtle"
+            >
+              {localization.upload}
+            </StyledButton>
+          </Flex>
         </>
       )}
     </StyledDropzone>
--- a/packages/big-design/src/components/FileUploader/FileUploader.tsx
+++ b/packages/big-design/src/components/FileUploader/FileUploader.tsx
@@ -20,7 +20,7 @@ import { InlineMessage } from '../InlineMessage';
 import { defaultLocalization } from './constants';
 import { File } from './File';
 import { StyledDropZoneWrapper, StyledFileUploaderWrapper, StyledList } from './styled';
-import type { Localization, ValidatorConfig } from './types';
+import { Action, Localization, ValidatorConfig } from './types';
 import { getImagesPreview, validateFiles } from './utils';
 
 export interface FileAction extends Omit<DropdownItem, 'onItemClick'> {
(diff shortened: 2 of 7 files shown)


MichalDobry (approved):
LGTM

chanceaclark (changes requested):
Just to prevent the component from getting through, I am requesting changes here. We are renaming and changing parts of the API here, which is considered a breaking change. Can we think of other ways to extend the component here instead?

packages/big-design/src/components/FileUploader/FileUploader.tsx
```diff
@@ -42,11 +42,12 @@ interface FileValidationError {
 }
 
 interface Props {
-  actions?: FileAction[];
```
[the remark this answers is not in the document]
bc-apostoliuk: Yes, but `actions` field was not even described in the documentation, so it does not have consumers.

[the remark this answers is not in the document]
chanceaclark: We can't assume that as people (3rd party devs) may have found that prop by other means.

packages/big-design/src/components/FileUploader/DropZone.tsx
```diff
@@ -149,6 +151,33 @@ export const DropZone = ({
...
 
+  const renderedAction = useMemo(() => {
+    if (!action) {
+      return null;
+    }
+
+    const { label, onClick, ...actionProps } = action;
+
+    return (
+      <StyledButton
```
chanceaclark, line 162: Let's make sure to put `type="button"` on each of these `StyledButton`'s so it doesn't submit a form.

packages/big-design/src/components/FileUploader/FileUploader.tsx
```diff
@@ -60,7 +61,7 @@ export type FileUploaderProps = Props & ComponentPropsWithoutRef<'input'>;
 
 export const FileUploader: React.FC<FileUploaderProps> = ({
   accept,
-  actions,
+  actions: fileActions,
```
chanceaclark, line 64: Can we avoid this rename?

packages/big-design/src/components/FileUploader/spec.tsx
```diff
@@ -372,6 +372,29 @@ describe('FileUploader', () => {
 
     expect(screen.getAllByText('File name is too long')).toHaveLength(15);
   });
+
+  it('renders additional action', async () => {
```
chanceaclark, line 376: Would be nice to have some additional tests around not being able to pass styles or different button variants into `dropzoneConfig.action`

packages/docs/PropTables/FileUploaderPropTable.tsx
```diff
@@ -33,8 +54,14 @@ const fileUploaderProps: Prop[] = [
   },
   {
     name: 'dropzoneConfig',
-    types: '{ label?: string; description?: string; icon?: ReactNode }',
-    description: 'Adds a label and a description to the drop-zone box.',
+    types: '{ label?: string; description?: string; icon?: ReactNode, action: object }',
```
chanceaclark, line 57: At this point we probably should split this into its own tab on the prop table.

packages/docs/PropTables/FileUploaderPropTable.tsx
```diff
@@ -157,6 +184,70 @@ const fileUploaderValidatorProps: Prop[] = [
...
+  {
+    name: 'onItemClick',
+    types: '(name: File, idx: number) => void',
+    required: true,
+    description: 'Triggers when clicking on action item.',
+  },
+  {
+    name: 'tooltip',
+    types: 'string',
+    description: <>Adds tooltip for disabled item.</>,
```
chanceaclark, line 242:
```diff
- description: <>Adds tooltip for disabled item.</>,
+ description: 'Adds tooltip for disabled item.',
```

chanceaclark (commented):
Thanks for refactoring! Got a few comments for you but looking good.

packages/big-design/src/components/FileUploader/DropZone.tsx
bc-apostoliuk: Done

packages/big-design/src/components/FileUploader/FileUploader.tsx
bc-apostoliuk: yes, done

packages/big-design/src/components/FileUploader/spec.tsx
bc-apostoliuk: I have already covered.

packages/docs/PropTables/FileUploaderPropTable.tsx
bc-apostoliuk: I have extracted it to a separate table.

packages/docs/PropTables/FileUploaderPropTable.tsx
bc-apostoliuk: Done

packages/big-design/src/components/FileUploader/types.ts
```diff
@@ -11,3 +13,7 @@ export interface ValidatorConfig {
   type: string;
   validator: (file: File) => boolean;
 }
+
+export interface Action extends Omit<ButtonProps, 'variant'> {
```
golcinho, line 17: should we `omit` `icons` and `margin` props?

bc-apostoliuk: Updated

```
export interface Action
  extends Omit<
    ButtonProps,
    | 'variant'
    | 'children'
    | 'iconOnly'
    | 'iconRight'
    | 'iconLeft'
    | 'margin'
    | 'marginTop'
    | 'marginRight'
    | 'marginBottom'
    | 'marginLeft'
    | 'marginVertical'
    | 'marginHorizontal'
  > {
  label: string;
}
```

packages/big-design/src/components/FileUploader/DropZone.tsx
```diff
@@ -149,6 +151,34 @@ export const DropZone = ({
     );
   }, [description]);
 
+  const renderedAction = useMemo(() => {
+    if (!action) {
+      return null;
+    }
+
+    const { label, onClick, style, ...actionProps } = action;
```
golcinho, line 159: According to updated `interface` we need to user `excludeMarginProps` to exclude `margin` props here.

https://github.com/bigcommerce/big-design/blob/main/packages/big-design/src/helpers/margins/margins.tsx#L33C17-L33C35

bc-apostoliuk: Done

golcinho (approved):

packages/big-design/src/components/FileUploader/DropZone.tsx
```diff
@@ -191,15 +222,19 @@ export const DropZone = ({
...
-          <StyledButton
-            color="secondary"
-            disabled={disabled}
-            marginTop={isRowView ? 'none' : 'medium'}
-            onClick={(e) => e.preventDefault()}
-            variant="subtle"
-          >
-            {localization.upload}
-          </StyledButton>
+          <Flex>
```
chanceaclark, line 225: 🍹 Might want to enforce the flex direction for mobile devices
```suggestion
          <Flex flexDirection="row">
```

chanceaclark (approved):

jorgemoya (approved):

packages/big-design/src/components/FileUploader/DropZone.tsx
bc-apostoliuk: Done

PR #1567, opened 2024-10-16
feat(component): implement FeatureTag component

davelinke:
## What?

Introducing the FeatureTag component

## Why?

The feature tag component is a stylized badge interface element that describes categories of a particular application or feature within the BC properties.

### Figma
Screenshot 2024-12-13 at 3 09 31 PM

Screenshot 2024-12-27 at 2 40 19 PM

### Prototype
Screenshot 2024-12-27 at 2 39 24 PM

## Testing/Proof

<details>

```tsx
import { FeatureTag, Flex, Grid, FlexItem, FeatureSet, H2 } from '@bigcommerce/big-design';
import { AutoAwesomeIcon } from '@bigcommerce/big-design-icons';

const PageFeatureTag: FunctionComponent = () => {
  return (
    <Flex flexDirection="column" flexGap="1rem" padding="large">
      <H2>Feature set</H2>
      <FlexItem>
        <FeatureSet
          features={[
            { label: 'Feature 1', icon: <AutoAwesomeIcon /> },
            { label: 'Feature 2', icon: <AutoAwesomeIcon /> },
            { label: 'Feature 3', icon: <AutoAwesomeIcon /> },
          ]}
        ></FeatureSet>
      </FlexItem>
      <FlexItem>
        <Grid gridColumns={'100px'}>
        <FeatureSet
          features={[
            { label: 'A really long text', icon: <AutoAwesomeIcon /> },
            { label: 'A really long text', icon: <AutoAwesomeIcon /> },
            { label: 'A really long text', icon: <AutoAwesomeIcon /> },
          ]}
        ></FeatureSet>
        </Grid>
      </FlexItem>
      <H2>Feature tag</H2>
      <FeatureTag icon={<AutoAwesomeIcon />} label="Woot" />
      <FeatureTag icon={<AutoAwesomeIcon />} label="BigCommerce" />
      <FeatureTag icon={<AutoAwesomeIcon />} label="BigCommerce" marginBottom={'xLarge'} />
      <Grid gridColumns="100px">
        <FeatureTag
          icon={<AutoAwesomeIcon />}
          label="A really long text"
          title="A really long text"
        />
      </Grid>
    </Flex>
  );
};

export default PageFeatureTag;


```

</details>


packages/big-design/src/components/FeatureTag/FeatureTag.tsx
```diff
@@ -0,0 +1,32 @@
...
+      target={target}
+      title={label}
+    >
+      {icon ? <StyledFeatureTagIcon>{icon}</StyledFeatureTagIcon> : null}
+      <StyledFeatureTagLabel>{label}</StyledFeatureTagLabel>
+    </StyledFeatureTag>
+  ) : null;
+};
+
+FeatureTag.displayName = 'FeatureTag';
```
chanceaclark, line 32: Ton's of things here:
1. For icon props, we just use `ReactNode` as the type. Allows consumers to pass in high-order components/icons if needed.
2. I know this might be a contention point, but let's remove the `isActive` prop. It feels like it's a link state, rather than prop state here.
3. Use `ComponentPropsWithoutRef` instead of `HTML<Element>Attributes` to extend from.
4. Avoid using `FunctionComponent`, instead use `({ some, prop, ...props }: Props) => {...}` as React now implies it's a functional component.
5. You'll see in `styled.tsx` some simplification there, which simplifies the HTML and styles in the component.
6. Spread the rest of the props as we are extending a link component.

```suggestion
import React, { ComponentPropsWithoutRef, forwardRef, ReactNode } from 'react';

import { StyledFeatureTag, StyledFeatureTagLabel } from './styled';

export interface FeatureTagProps extends ComponentPropsWithoutRef<'a'> {
  icon?: ReactNode;
  label: string;
}

export const FeatureTag = forwardRef<HTMLAnchorElement, FeatureTagProps>(
  ({ label, icon, ...props }, ref) => {
    return label ? (
      <StyledFeatureTag ref={ref} {...props}>
        {icon}
        <StyledFeatureTagLabel>{label}</StyledFeatureTagLabel>
      </StyledFeatureTag>
    ) : null;
  },
);

FeatureTag.displayName = 'FeatureTag';
```

packages/big-design/src/components/FeatureTag/styled.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+`;
+
+StyledFeatureTagIcon.defaultProps = { theme: defaultTheme };
+
+export const StyledFeatureTagLabel = styled.span`
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+`;
```
chanceaclark, line 79: 1. We don't need `fill` set as our icons use `color: currentColor` for the stroke/fill it needs.
2. Switch over to the `theme.typography.fontSize` + `theme.lineHeight` settings, instead of using `remCalc` (Use `remCalc` when we don't have the value on the theme).
3. Use the `& > svg` to size the SVG instead of needing another HTML element.
4. Use `ellipsis` from `polished`.
5. Cleaned up link state styles.

```suggestion
import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
import { ellipsis } from 'polished';
import styled from 'styled-components';

export const StyledFeatureTag = styled.a`
  align-items: center;
  background-color: ${({ theme }) => theme.colors.secondary20};
  border-radius: ${({ theme }) => theme.borderRadius.normal};
  display: inline-flex;
  color: ${({ theme }) => theme.colors.secondary60};
  cursor: pointer;
  flex-wrap: nowrap;
  font-size: ${({ theme }) => theme.typography.fontSize.small};
  gap: ${({ theme }) => theme.spacing.xxSmall};
  justify-content: center;
  line-height: ${({ theme }) => theme.lineHeight.small};
  max-width: fit-content;
  outline: none;
  padding-block: ${({ theme }) => theme.helpers.remCalc(2)};
  padding-inline: ${({ theme }) => theme.spacing.xSmall};
  text-decoration: none;
  user-select: none;

  &:hover {
    background-color: ${({ theme }) => theme.colors.primary20};
    color: ${({ theme }) => theme.colors.primary60};
  }

  &:focus-visible {
    outline: 4px solid ${({ theme }) => theme.colors.primary20};
  }

  &:active {
    background-color: ${({ theme }) => theme.colors.primary10};
    color: ${({ theme }) => theme.colors.primary50};
  }

  &:focus-visible:active {
    background-color: ${({ theme }) => theme.colors.primary10};
    color: ${({ theme }) => theme.colors.primary60};
  }

  & > svg {
    height: ${({ theme }) => theme.spacing.medium};
    flex-shrink: 0;
    width: ${({ theme }) => theme.spacing.medium};
  }
`;

StyledFeatureTag.defaultProps = { theme: defaultTheme };

export const StyledFeatureTagLabel = styled.span`
  ${ellipsis()}
`;
```

chanceaclark:
Before you convert this from draft to regular PR make sure to add documentation.

packages/big-design/src/components/FeatureTag/FeatureTag.tsx
davelinke: Hi @chanceaclark .

Good pointers, but regarding the isActive prop, it denotes that it is the current category that you are browsing, just like a nav menu would highlight the current page. Would it be better to change the prop name to `isCurrent`?

chanceaclark: Can you provide an example on how app developers are going to use this? To me it seems like `FeatureTag`s list the capabilities of the app without linking anywhere. It's not clear to me how an app developer should think about using this component, specifically linking to what/where?

If we are trying to build a component that has active click states, then we need to redesign the concept of this component to act more like a tabable navigation instead as there are other accessibility roles we should handle as a library.

davelinke: Sure, the iadea now is the following. When seeing a particular integration within the control panel, there will be feature tags representing the type and category of this particular app or integration. This could be things like, headless, payments, checkout and so on.

If you were consuming that particular content within the context of say "headless", then the "headless" feature tag would be marked as "current" or "active", kind of like filter chips on the sidebar of our storefronts when a filtering occurs.

The feature tag has always been thought as a navigational item, that's why the current/active state was set in place.

davelinke: Following Figma asset as source of truth:
https://www.figma.com/design/ig91vdN7fcn1sCPgj68jA0/BigDesign-Component-Library?node-id=1446-29&node-type=canvas&t=ExzdQvs66DfnrxnX-0

davelinke: After discussion with both @BC-Turner and @chanceaclark , the direction of this component is set to be a tag and not a link.

becomevocal (approved):
This follows all our best practices so I'm a 👍

jorgemoya (approved):

PR #1580, opened 2024-11-01
chore(common): APPSEC-705 Add SECURITY.md

bc-donfran:
## What?
Add SECURITY.md.

## Why?
Helps researchers/Github users report security bugs to us.

## Testing / Proof
NA

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,11 @@
+# Reporting security issues
+BigCommerce is dedicated to the responsible disclosure of security vulnerabilities.
+If you have found a security vulnerability in an active open-source repository created and owned by BigCommerce, please report it to our [public bug bounty program](https://bugcrowd.com/bigcommerce). If you would prefer to submit via email, please send your report to [security@bigcommerce.com](mailto:security@bigcommerce.com).
+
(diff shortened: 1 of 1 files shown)


PascalZajac:
You'll need to amend your commit message to use one of the defined scopes, plus fix the casing.

I checked some other open-source repositories like https://github.com/bigcommerce/checkout-js and https://github.com/bigcommerce/cornerstone and can't find another `SECURITY.md` to compare to, is this our new standard?

bc-donfran:
Hey Pascal :D Fixed up the linting issue and PRs have been made for checkout-js and cornerstone. The closest you will see would be for catalyst found here: https://github.com/bigcommerce/catalyst/blob/main/SECURITY.md. This will be the standard for our commonly used repos so that we can reliably get researchers to submit any security bugs to us for BugCrowd and this came off a discussion from this issue: https://github.com/bigcommerce/bigcommerce-for-wordpress/issues/338 which was left public and open and unknown.

PascalZajac (approved):
Thanks for the explanation.

The text in this version seems slightly more polished than the Catalyst repository, I take it that will be updated to match?

bc-donfran:
Yes it will be.

PR #1598, opened 2024-11-25
Add new StatelessPagination component and documentation

leeBigCommerce:
## What?

- Add new `StatelessPagination` component
- Add new documentation detailing `StatelessPagination`

### Upcoming

- `StatelessPagination` will be added as an option for the `Table` component in a seperate, upcoming PR
- `OffsetPagination` will be changed to use `StatelessPagination` under the hood in a seperate, upcoming PR

## Why?

- The current `OffsetPagination` component is highly stateful and requires knowable page numbers, these concepts are incompatible with cursor-based pagination
  - cursor-based pagination has very fluid concepts of 'pages' and how many there might be
  - cursor-based pagination `next` and `previous` actions are knowable for the current page only, they cannot be deterministically calculated for _all_ pages ahead of time

### Why not name the component `CursorPagination`?

- The component's props and internals are completely agnostic to cursors being used or not
  - any attempt to do so may unintentionally couple it to a specific, idiosyncratic cursor API 
- Expands the component's potential usage to other forms of pagination or more complicated logic
  - e.g. a paginated form where field states are interrogated before acting on a next or previous navigation 

### Why is the example in the documentation closer to a true cursor implementation?

- Faking a somewhat realistic cursor API takes a _lot_ of boiler plate code which would dominate and confuse the example code
- You can see a full example here: https://github.com/leeBigCommerce/big-design/blob/7568a4f73bbf76dcff5b0e45fd49ba5ebec6ceeb/packages/docs/pages/stateless-pagination.tsx
  - Not a merge candidate, purely for reference
  - Extra ugly as everything is in one file, lodash and other helpful libraries  

https://github.com/user-attachments/assets/c995cec7-e9a8-461a-8a3b-7667094ec89e

## Testing/Proof

- New tests for `StatelessPagination`

### DIFF
--- a/packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
+++ b/packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
@@ -1,0 +1,102 @@
+import {
+  ArrowDropDownIcon,
+  ChevronLeftIcon,
+  ChevronRightIcon,
+} from '@bigcommerce/big-design-icons';
+import React, { memo } from 'react';
+
+import { MarginProps } from '../../helpers';
+import { Dropdown, DropdownItem } from '../Dropdown';
+import { Flex, FlexItem } from '../Flex';
+
+import { StyledButton } from './styled';
+
+export interface StatelessPaginationLocalization {
+  previousPage: string;
+  nextPage: string;
+}
+
+const defaultLocalization: StatelessPaginationLocalization = {
+  previousPage: 'Previous page',
+  nextPage: 'Next page',
+};
+
+export interface StatelessPaginationProps extends MarginProps {
+  itemsPerPage: number;
+  itemsPerPageOptions: number[];
+  onItemsPerPageChange(range: number): void;
+  disableNext?: boolean;
+  onNext(): void;
+  disablePrevious?: boolean;
+  onPrevious(): void;
+  label?: string;
+  rangeLabel?: string;
+  localization?: StatelessPaginationLocalization;
+}
+
+export const StatelessPagination: React.FC<StatelessPaginationProps> = memo(
+  ({
+    itemsPerPage,
+    itemsPerPageOptions,
+    disableNext = false,
+    onNext,
+    disablePrevious = false,
+    onPrevious,
+    onItemsPerPageChange,
+    label = 'pagination',
+    localization = defaultLocalization,
+    rangeLabel,
+  }) => {
+    const handleRangeChange = (item: DropdownItem) => {
+      onItemsPerPageChange(Number(item.hash));
+    };
+
+    return (
+      <Flex aria-label={label} flexDirection="row" role="navigation">
+        <FlexItem>
+          <Dropdown
+            items={itemsPerPageOptions.map((range) => ({
+              content: `${range}`,
+              hash: `${range}`,
+              onItemClick: handleRangeChange,
+            }))}
+            positionFixed={true}
+            selectedItem={{
+              content: `${itemsPerPage}`,
+              hash: `${itemsPerPage}`,
+              onItemClick: handleRangeChange,
+            }}
+            toggle={
+              <StyledButton
+                iconRight={<ArrowDropDownIcon size="xxLarge" />}
+                type="button"
+                variant="subtle"
+              >
+                {rangeLabel || `Show ${itemsPerPage} items`}
+              </StyledButton>
+            }
+          />
+        </FlexItem>
+        <FlexItem>
+          <StyledButton
+            disabled={disablePrevious}
+            iconOnly={<ChevronLeftIcon title={localization.previousPage} />}
+            onClick={() => onPrevious()}
+            type="button"
+            variant="subtle"
+          />
+
+          <StyledButton
+            disabled={disableNext}
+            iconOnly={<ChevronRightIcon title={localization.nextPage} />}
+            onClick={() => onNext()}
+            type="button"
+            variant="subtle"
+          />
+        </FlexItem>
+      </Flex>
+    );
+  },
+);
+
+StatelessPagination.displayName = 'StatelessPagination';
--- a/packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
+++ b/packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
@@ -3,7 +3,7 @@ import {
   ChevronLeftIcon,
   ChevronRightIcon,
 } from '@bigcommerce/big-design-icons';
-import React, { memo } from 'react';
+import React, { memo, useMemo } from 'react';
 
 import { MarginProps } from '../../helpers';
 import { Dropdown, DropdownItem } from '../Dropdown';
@@ -12,14 +12,20 @@ import { Flex, FlexItem } from '../Flex';
 import { StyledButton } from './styled';
 
 export interface StatelessPaginationLocalization {
-  previousPage: string;
-  nextPage: string;
+  previousPage?: string;
+  nextPage?: string;
+  rangeLabel?: string;
+  label?: string;
 }
 
-const defaultLocalization: StatelessPaginationLocalization = {
+const getDefaultLocalization = (
+  itemsPerPage: number,
+): Required<StatelessPaginationLocalization> => ({
   previousPage: 'Previous page',
   nextPage: 'Next page',
-};
+  label: 'pagination',
+  rangeLabel: `Show ${itemsPerPage} items`,
+});
 
 export interface StatelessPaginationProps extends MarginProps {
   itemsPerPage: number;
@@ -29,8 +35,6 @@ export interface StatelessPaginationProps extends MarginProps {
   onNext(): void;
   disablePrevious?: boolean;
   onPrevious(): void;
-  label?: string;
-  rangeLabel?: string;
   localization?: StatelessPaginationLocalization;
 }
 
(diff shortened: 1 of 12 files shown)


packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
```diff
@@ -0,0 +1,102 @@
...
+            type="button"
+            variant="subtle"
+          />
+        </FlexItem>
+      </Flex>
+    );
+  },
+);
+
+StatelessPagination.displayName = 'StatelessPagination';
```
matt-evangelidis, line 102: I'm assuming it's a library thing, but what's the `displayName` field used for?

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
```diff
@@ -0,0 +1,102 @@
...
+              hash: `${itemsPerPage}`,
+              onItemClick: handleRangeChange,
+            }}
+            toggle={
+              <StyledButton
+                iconRight={<ArrowDropDownIcon size="xxLarge" />}
+                type="button"
+                variant="subtle"
+              >
+                {rangeLabel || `Show ${itemsPerPage} items`}
```
matt-evangelidis, line 75: Is it worth pulling out `Show ${itemsPerPage} items` into a `defaultLabel` variable to make it a tad more explicit? I'd imagine we'd set the `rangeLabel` prop to `rangeLabel = "Show ${itemsPerPage} items" ` if we could.

packages/big-design/src/components/StatelessPagination/spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+  );
+
+  const option = await findByText('Show 2 items');
+
+  await userEvent.click(option);
+  await userEvent.keyboard('{ArrowDown}{Enter}');
+
+  expect(changeRange).toHaveBeenCalled();
+  expect(option).toBeInTheDocument();
+});
```
matt-evangelidis, line 69: Not quite sure on the exact action being tested. Is this testing for selecting `2` in the dropdown or selecting `3`? 

Not sure if the initial click puts the current user selection at the parent element or at the first child element in the dropdown.

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
leeBigCommerce: turns out, we can: https://github.com/bigcommerce/big-design/pull/1598/commits/34b20eeaa9cdd0b63944fe41fd1e4b6ace5e8219

I've never thought to try using other variables within a defaulting before 🤔

packages/big-design/src/components/StatelessPagination/spec.tsx
leeBigCommerce: that is what's happening, by default the currently highlighted option will be focussed on first interaction. Pressing down goes to the next one (3) in the list

the `Dropdown` component (which this uses under the hood) already has tests proving keyboard usability of this component: https://github.com/bigcommerce/big-design/blob/main/packages/big-design/src/components/Dropdown/spec.tsx#L326-L342

so I've updated this _test_ to use the more explicit form of selecting and clicking on items: https://github.com/bigcommerce/big-design/pull/1598/commits/8884d2bf38ebd818ee2fd7d13e56f7741d23bee8

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
leeBigCommerce: it helps provide more specific debug messages when things go wrong

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
matt-evangelidis: 😮

packages/big-design/src/components/StatelessPagination/spec.tsx
matt-evangelidis: Yep, thanks kindly. 

The clicking is definitely more explicit, but the most helpful bit here is the `toHaveBeenCalledWith(3)`. If we went back to the keyboard interaction, you'd be much more easily able to puzzle out that if the second option is being tested for, the selection must start at the first option.

matt-evangelidis (approved):
Very nice, @leeBigCommerce 👍

Thanks kindly for clarifying some things and making tests clearer.

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
```diff
@@ -0,0 +1,102 @@
...
+  itemsPerPage: number;
+  itemsPerPageOptions: number[];
+  onItemsPerPageChange(range: number): void;
+  disableNext?: boolean;
+  onNext(): void;
+  disablePrevious?: boolean;
+  onPrevious(): void;
+  label?: string;
+  rangeLabel?: string;
+  localization?: StatelessPaginationLocalization;
```
icatalina, line 34: 🤔 What makes `label` and `rangeLabel` special in comparison to `previousPage`/`nextPage`?

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const defaultLocalization: StatelessPaginationLocalization = {
+  previousPage: 'Previous page',
+  nextPage: 'Next page',
+};
+
+export interface StatelessPaginationProps extends MarginProps {
+  itemsPerPage: number;
+  itemsPerPageOptions: number[];
+  onItemsPerPageChange(range: number): void;
```
icatalina, line 27: 🤔 is this a `range?` or more like a `count`?

icatalina (approved):

icatalina:
Just some nitpicking... overall looks pretty good

packages/big-design/src/components/StatelessPagination/spec.tsx
```diff
@@ -0,0 +1,181 @@
+import { render, screen } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
+import React from 'react';
+
+import 'jest-styled-components';
+
+import { StatelessPagination } from './index';
+
+test('render Stateless pagination component', async () => {
+  const { findByRole } = render(
```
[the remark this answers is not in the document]
leeBigCommerce: guilty, fixed here: https://github.com/bigcommerce/big-design/pull/1598/commits/706155c971b67f87612923f0d9dea7244b799c2f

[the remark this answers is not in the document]
matt-evangelidis: Out of curiosity, is there a functional difference between using `screen.findByRole` and the function pulled from `render`? Examines a different component scope or something?

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
leeBigCommerce: other than `rangeLabel` being a touch more volatile (given the limit can changed), not a lot

I've moved them all into a consolidated `locaization` object: https://github.com/bigcommerce/big-design/pull/1598/commits/7e2fe591649c9c0eea33fa87816fd92b5059c7f5

it add a _tiny_ bit more complexity to the implementation, now that it's fully optional and one default is dynamic

packages/big-design/src/components/StatelessPagination/spec.tsx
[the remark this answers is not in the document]
leeBigCommerce: functionally, very little

the deconstructing of `render` can start to look messy and change a little frequently
using `screen` also prevents people from changing the `baseElement` (the searchable context), which can catch people out

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
leeBigCommerce: I guess not really, I'd think of a range as a `1 -10`, `50-100`, two value concept

to me, this'd be `newItemsPerPage` or `chosenItemsPerPage` (given it has been picked by the user, but not yet added to state)

but they're long and we already have precedent of `range` on the docs for regular pagination: https://bigcommerce.github.io/big-design/pagination#props

packages/big-design/src/components/StatelessPagination/StatelessPagination.tsx
```diff
@@ -0,0 +1,105 @@
...
+    itemsPerPageOptions,
+    onNext,
+    onPrevious,
+    onItemsPerPageChange,
+    localization: customLocalization,
+  }) => {
+    const localization: Required<StatelessPaginationLocalization> = useMemo(
+      () => ({ ...getDefaultLocalization(itemsPerPage), ...customLocalization }),
+      [customLocalization, itemsPerPage],
+    );
```
matt-evangelidis, line 51: Lot of hoop-jumping to handle default and custom localisation 😂

I imagine there aren't too many ways of simplifying this without compromises. The one thing I can think of is to go back to treating the default `rangeLabel` field as its own special case which would reduce the number of dependencies involved for memoisation, but now we're just backtracking. This is probably the best solution.

matt-evangelidis (approved):
👏

packages/big-design/src/components/StatelessPagination/styled.tsx
```diff
@@ -0,0 +1,11 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled from 'styled-components';
+
+import { StyleableButton } from '../Button/private';
+
+export const StyledButton = styled(StyleableButton)`
```
chanceaclark, line 6: Since `defaultProps` is deprecated in React, we've been trying to migrate over to `.attrs({ theme: defaultTheme })` here instead.

chanceaclark (approved):
Just a reminder to add a changeset for this PR and the other one (`pnpm changeset`).

packages/big-design/src/components/StatelessPagination/styled.tsx
leeBigCommerce: done: https://github.com/bigcommerce/big-design/pull/1598/commits/df1613855107fb6cff8e0a4014e8e206ece956c7

chanceaclark (approved):

PR #1602, opened 2024-12-03
Incorporate StatelessPagination to Table and TableNext

leeBigCommerce:
**N.B.**

- best read with `Hide whitespace` enabled as some nesting/spacing has changed
- the duplication across `Table` and `TableNext` is intentional and required
  - the long term plan is to eventually move everyone to `TableNext`, duplication prevents entanglement

## What?

- Enable our `Table` component to make use of `StatelessPagination`
- Additive change where a `Table`'s pagination prop can be either those of `OffsetPagination` _or_ `StatelessPagination`
  - the developer adheres to either signature, which is then discriminated within the `Table` component

## Why?

- Allows `Table` consumers to make use of cursor-based, or more opinionated, versions of pagination 

### New `Stateless` example with the `Table` docs

https://github.com/user-attachments/assets/2aed49d0-1783-483e-b73c-344f10583f3a

### Selection behaviour differences between the pagination types
(the examples used in the video were _not_ added to this PR for brevity within the docs)

https://github.com/user-attachments/assets/6d76758e-6ba2-4c16-9e95-6b2ca5acc786

## Testing/Proof

- New tests within `Table` and `TableNext`
- Unchanged, passing tests within `Table` and `TableNext` proving current features are unaffected

### DIFF
--- a/packages/big-design/src/components/Table/Actions/Actions.tsx
+++ b/packages/big-design/src/components/Table/Actions/Actions.tsx
@@ -34,15 +34,15 @@ const InternalActions = <T extends TableItem>({
   ...props
 }: ActionsProps<T>) => {
   const isSelectable = typeof onSelectionChange === 'function';
-  const statelessPagination = pagination?.type === 'stateless';
-  const totalItems = pagination && !statelessPagination ? pagination.totalItems : items.length;
+  const paginationWithoutTotal = pagination && pagination.totalItems === undefined;
+  const totalItems = pagination?.totalItems ?? items.length;
 
   const renderItemName = () => {
     if (typeof itemName !== 'string') {
       return null;
     }
 
-    const text = isSelectable || statelessPagination ? itemName : `${totalItems} ${itemName}`;
+    const text = isSelectable || paginationWithoutTotal ? itemName : `${totalItems} ${itemName}`;
 
     return (
       <FlexItem flexShrink={0} marginRight="medium">
@@ -71,7 +71,7 @@ const InternalActions = <T extends TableItem>({
         items={items}
         onChange={onSelectionChange}
         selectedItems={selectedItems}
-        totalItems={!statelessPagination ? totalItems : undefined}
+        totalItems={!paginationWithoutTotal ? totalItems : undefined}
       />
       {renderItemName()}
       {renderActions()}
--- a/packages/big-design/src/components/Table/Actions/Actions.tsx
+++ b/packages/big-design/src/components/Table/Actions/Actions.tsx
@@ -59,10 +59,11 @@ const InternalActions = <T extends TableItem>({
     <StyledFlex
       alignItems="center"
       aria-controls={tableId}
-      data-testid="table-controls"
+      aria-label="Table Controls"
       flexDirection="row"
       justifyContent="stretch"
       ref={forwardedRef}
+      role="toolbar"
       stickyHeader={stickyHeader}
       {...props}
     >
--- a/packages/big-design/src/components/Table/Actions/Actions.tsx
+++ b/packages/big-design/src/components/Table/Actions/Actions.tsx
@@ -5,7 +5,7 @@ import { FlexItem } from '../../Flex';
 import { Text } from '../../Typography';
 import { SelectAll } from '../SelectAll';
 import { TablePagination } from '../TablePagination';
-import { TableItem, TablePaginationProps, TableSelectable } from '../types';
+import { DiscriminatedTablePaginationProps, TableItem, TableSelectable } from '../types';
 
 import { StyledFlex } from './styled';
 
@@ -14,7 +14,7 @@ export interface ActionsProps<T> {
   forwardedRef: RefObject<HTMLDivElement>;
   itemName?: string;
   items: T[];
-  pagination?: TablePaginationProps;
+  pagination?: DiscriminatedTablePaginationProps;
   onSelectionChange?: TableSelectable<T>['onSelectionChange'];
   selectedItems: Set<T>;
   stickyHeader?: boolean;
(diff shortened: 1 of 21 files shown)


packages/big-design/src/components/Table/spec.tsx
```diff
@@ -197,90 +197,219 @@ test('renders the total number of items + item name', () => {
...
+            { header: 'Stock', hash: 'stock', render: ({ stock }) => stock },
+          ]}
+          itemName="Product"
+          items={[
+            { sku: 'SM13', name: '[Sample] Smith Journal 13', stock: 25 },
+            { sku: 'DPB', name: '[Sample] Dustpan & Brush', stock: 34 },
+            { sku: 'OFSUC', name: '[Sample] Utility Caddy', stock: 45 },
+            { sku: 'CLC', name: '[Sample] Canvas Laundry Cart', stock: 2 },
+            ...selectedItems,
+          ]}
```
matt-evangelidis, line 391: Just to clarify, what does providing an overlapping product between `items` and `selectedItems` prove about the table's behaviour?

leeBigCommerce: it's more a requirement of `Table`'s implementation

inside it does an strict equality check, so they literally need to be the same objects to class as 'selected'

matt-evangelidis: Ah, yes, the good old internal `Set`...

matt-evangelidis (approved):
🙌

packages/big-design/src/components/TableNext/Actions/Actions.tsx
```diff
@@ -74,6 +75,7 @@ const InternalActions = <T extends TableItem>({
     <StyledFlex
       alignItems="center"
       aria-controls={tableId}
+      data-testid="table-controls"
```
chanceaclark, line 78: I think we want to avoid adding a test id wherever possible. I know we have some floating around the repo so we should try and clean those up eventually.

But this brings in the question, is this `Actions` component actually accessible? Do we need to add either a `toolbar` or `menubar` role to this element?

packages/big-design/src/components/TableNext/Actions/Actions.tsx
```diff
@@ -50,14 +50,15 @@ const InternalActions = <T extends TableItem>({
...
+  const statelessPagination = pagination?.type === 'stateless';
+  const totalItems = pagination && !statelessPagination ? pagination.totalItems : items.length;
 
   const renderItemName = () => {
     if (typeof itemName !== 'string') {
       return null;
     }
 
-    const text = isSelectable ? itemName : `${totalItems} ${itemName}`;
+    const text = isSelectable || statelessPagination ? itemName : `${totalItems} ${itemName}`;
```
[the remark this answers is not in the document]
leeBigCommerce: I've gone back and forth on this one

`StatelessPagination` does not take a `totalItems` props for a couple of reasons

1. not all cursor pagination implementations offer a total count (it runs a little counter to the "endless feed of lots of stuff" angle)
2. if an implementation did offer a total count, we were only using it to build a range label. Users can set the range label directly by optionally setting directly `localization.rangeLabel`
3. we could add an optional `totalItems` prop and change `localization.rangeLabel` back into a function `(totalItems?: number) => string` - but then we're just giving the user back a number they just handed to us

in short, it seemed kinda redundant to have it

_however_

should a user be using a `Table` with a cursor based pagination that _does_ provide a total count, it would be nice to offer them a `x/y` summary

to achieve this, I've added an optional `totalItems` not to `StatelessPagination` itself, but to an augmented version of its props imported by `Table`: https://github.com/bigcommerce/big-design/pull/1602/commits/aba906a689200952494b2cae6a220e5741345afb

on the plus side it works and does not make `StatelessPagination`'s props less concise, the downside is that it's kinda 'secret' and not easy to document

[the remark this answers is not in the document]
leeBigCommerce: still on the fence, but see what you think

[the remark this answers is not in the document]
chanceaclark: I think if an API does offer `totalItems` we should provide a way to show it rather than gating it under a specific set of props. This specific part I was thinking about applies to the number of items selected, which would be valuable to show a total. It does beg the question if we should enable a consumer to localize the label though, but that can be for another PR.

packages/big-design/src/components/Table/Actions/Actions.tsx
```diff
@@ -58,17 +59,19 @@ const InternalActions = <T extends TableItem>({
     <StyledFlex
       alignItems="center"
       aria-controls={tableId}
+      aria-label="Table Controls"
```
chanceaclark, line 62: 🍹 We should localize this before we release, but you can tackle that in another PR if you want.

packages/big-design/src/components/Table/SelectAll/SelectAll.tsx
```diff
@@ -54,6 +54,9 @@ export const SelectAll = <T extends TableItem>({
 
   const label = allInPageSelected ? 'Deselect All' : 'Select All';
 
+  const nonSelectionSummary = totalItems || '';
```
chanceaclark, line 57: If `totalItems` is 0, wouldn't this be `false`, as in it'll show up in the UI as `false`? Should we use null coalescing instead?
```suggestion
  const nonSelectionSummary = totalItems ?? '';
```

I think there is another place that does this too.

packages/big-design/src/components/Table/Actions/Actions.tsx
leeBigCommerce: I think a seperate PR would be best, as we'll need to update the props of all `Table` variants in one go

packages/big-design/src/components/Table/SelectAll/SelectAll.tsx
leeBigCommerce: fixed: https://github.com/bigcommerce/big-design/pull/1602/commits/81e592f6d0fc61ff3311a757aa44723d4c9c16c2

packages/big-design/src/components/TableNext/Actions/Actions.tsx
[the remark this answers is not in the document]
leeBigCommerce: I forgot to attach the video (see below), but the idea was that https://github.com/bigcommerce/big-design/commit/aba906a689200952494b2cae6a220e5741345afb has enabled that situation, where an optional `totalItems` can be added and it'll be used in the summary


https://github.com/user-attachments/assets/030c2b4c-dc67-4724-82c5-7c9b829d7f39

the `itemName` prop can be used to localize the summary, right now we just don't hand over control of the `x/y` or if that goes before or after `itemName` - so it's probably not in high demand to localization

chanceaclark (approved):

PR #1610, opened 2024-12-13
chore(icons): add changeset to release trigger release

RobinOng:
## What? / Why?
Add `changeset` to trigger release. It will update minor version of `BigDesign` to include the new Pin icon


chanceaclark:
I don't think we need this as you already created a changeset in your PR. You should be able to just approve and merge #1609

chanceaclark (approved):

PR #1615, opened 2024-12-27
fix(component): remove vertical margin for collapse button

RobinOng:
## What?
Remove vertical margin for collapse button

## Why?
With the margin built in, the component is less flexible to use. Margin can be easily added back with a wrapping container

Before margin is removed:
Screenshot 2024-12-27 at 11 57 33 AM

Screenshot 2024-12-27 at 12 20 39 PM

With margin removed:
Screenshot 2024-12-27 at 12 06 56 PM

## Testing/Proof
Added above

### DIFF
--- a/.changeset/beige-lemons-sin.md
+++ b/.changeset/beige-lemons-sin.md
@@ -1,0 +1,6 @@
+---
+'@bigcommerce/big-design': patch
+'@bigcommerce/docs': patch
+---
+
+Remove vertical margin for collapse button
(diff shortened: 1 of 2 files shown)


bc-akshatkaushik (approved):

.changeset/beige-lemons-sin.md
```diff
@@ -0,0 +1,6 @@
+---
+'@bigcommerce/big-design': patch
+'@bigcommerce/docs': patch
```
chanceaclark, line 3: This is technically a visual breaking change, as it would change what a 3rd party app or view in the control panel would display. Can we chat outside of GitHub about the reason for this change?

As an aside, I'm thinking you can get away with a negative margin container to combat this margin if need be. Just thinking about a way to have a fix for this without adding a BC.

PR #1637, opened 2025-02-05
feat(component)!: change required behavior on form elements

davelinke:
## What?

Changing the behavior on form elements to render an asterisk when required instead of an optional label when optional

## Why?

This pattern follows the most encountered pattern on the web. 
The most common scenario with forms is that there will be little required fields to complete a form and many optionals.
Product Add/Edit follows the asterisk pattern due to this reason, and many other teams, B2B being the one whose user experience is looks flawed due to this factor have requested the change.

In terms of accessibility, screen readers announce to the user when a form field is required by making use of the native HTML required attribute. 

Screenshot 2025-02-05 at 3 09 41 PM

## Testing/Proof

<details>

```tsx
import {
  FormControlLabel,
  Counter,
  FileUploader,
  Input,
  Select,
  Textarea,
  MultiSelect,
  H1,
  Grid,
  GridItem,
} from '@bigcommerce/big-design';
import React, { FunctionComponent, useState } from 'react';

const RequiredFormItems: FunctionComponent = () => {
  return (
    <Grid gridColumns="repeat(2,1fr)" gridGap="2rem" padding={'xxLarge'}>
      <GridItem gridColumn={'span 2'}>
        <H1>Form Items</H1>
      </GridItem>

      <FormControlLabel htmlFor="email" renderRequired={true}>
        Sole label required
      </FormControlLabel>
      <FormControlLabel htmlFor="email">Sole label not required</FormControlLabel>

      <Input required label={'Input required'} />
      <Input label={'Input not required'} />

      <Counter label="Counter required" required onCountChange={() => {}} value={0}></Counter>
      <Counter label="Counter not required" onCountChange={() => {}} value={0}></Counter>

      <FileUploader label="File uploader required" required files={[]} onFilesChange={() => {}} />
      <FileUploader label="File uploader not required" files={[]} onFilesChange={() => {}} />

      <Select
        label="Select required"
        required
        options={[{ value: '1', content: 'Option 1' }]}
        onOptionChange={() => {}}
        value={'1'}
      />
      <Select
        label="Select not required"
        options={[{ value: '1', content: 'Option 1' }]}
        onOptionChange={() => {}}
        value={'1'}
      />

      <Textarea label="Textarea required" required />
      <Textarea label="Textarea not required" />
      
      <MultiSelect
        label="MultiSelect required"
        required
        options={[
          { value: '1', content: 'Option 1' },
          { value: '2', content: 'Option 2' },
        ]}
        value={['1']}
        onOptionsChange={() => {}}
      />
      <MultiSelect
        label="MultiSelect not required"
        options={[
          { value: '1', content: 'Option 1' },
          { value: '2', content: 'Option 2' },
        ]}
        onOptionsChange={() => {}}
        value={['1']}
      />
    </Grid>
  );
};

export default RequiredFormItems;
```

</details>

### DIFF
--- a/packages/big-design/src/components/Form/Label/Label.tsx
+++ b/packages/big-design/src/components/Form/Label/Label.tsx
@@ -5,19 +5,20 @@ import { Text } from '../../Typography/';
 import { StyledLabel } from './styled';
 
 export interface FormControlLabelProps extends ComponentPropsWithoutRef<'label'> {
-  renderRequired?: boolean;
+  required?: boolean;
 }
 
 export const FormControlLabel: React.FC<FormControlLabelProps> = ({
   className,
   style,
-  renderRequired,
+  required,
   ...props
 }) => (
   <StyledLabel {...props}>
     {props.children}
-    {renderRequired && (
-      <Text aria-hidden="true" as="span" color="danger" marginLeft="xxSmall">
+    {required && (
+      <Text aria-hidden="true" as="span" color="danger">
+        {' '}
         *
       </Text>
     )}
--- a/packages/big-design/src/components/Form/Label/Label.tsx
+++ b/packages/big-design/src/components/Form/Label/Label.tsx
@@ -1,23 +1,25 @@
 import React, { ComponentPropsWithoutRef } from 'react';
 
+import { Text } from '../../Typography/';
+
 import { StyledLabel } from './styled';
 
-export interface LabelLocalization {
-  optional: string;
-}
-
-const defaultLocalization: LabelLocalization = {
-  optional: 'optional',
-};
-
 export interface FormControlLabelProps extends ComponentPropsWithoutRef<'label'> {
-  renderOptional?: boolean;
-  localization?: LabelLocalization;
+  renderRequired?: boolean;
 }
 
 export const FormControlLabel: React.FC<FormControlLabelProps> = ({
   className,
-  localization = defaultLocalization,
   style,
+  renderRequired,
   ...props
-}) => <StyledLabel localization={localization} {...props} />;
+}) => (
+  <StyledLabel {...props}>
+    {props.children}
+    {renderRequired && (
+      <Text aria-hidden="true" as="span" color="danger" marginLeft="xxSmall">
+        *
+      </Text>
+    )}
+  </StyledLabel>
+);
--- a/packages/big-design/src/components/Form/Label/spec.tsx
+++ b/packages/big-design/src/components/Form/Label/spec.tsx
@@ -19,9 +19,9 @@ test('does not forward styles', () => {
   expect(container.firstChild).not.toHaveStyle('background: red');
 });
 
-test('appends * text when renderRequired', () => {
-  const { container } = render(<FormControlLabel renderRequired>This is a label</FormControlLabel>);
+test('appends * text when required', () => {
+  const { container } = render(<FormControlLabel required>This is a label</FormControlLabel>);
   const label = container.querySelector('label');
 
-  expect(label?.lastChild).toHaveTextContent('*');
+  expect(label).toHaveTextContent('This is a label *');
 });
--- a/packages/big-design/src/components/Form/Label/spec.tsx
+++ b/packages/big-design/src/components/Form/Label/spec.tsx
@@ -19,20 +19,9 @@ test('does not forward styles', () => {
   expect(container.firstChild).not.toHaveStyle('background: red');
 });
 
-test('appends (optional) text when renderOptional', () => {
-  const { container } = render(<FormControlLabel renderOptional>This is a label</FormControlLabel>);
+test('appends * text when renderRequired', () => {
+  const { container } = render(<FormControlLabel renderRequired>This is a label</FormControlLabel>);
   const label = container.querySelector('label');
 
-  expect(label).toHaveStyleRule('content', "' (optional)'", { modifier: '::after' });
-});
-
-test('appends custom (optional) text when passed', () => {
-  const { container } = render(
-    <FormControlLabel localization={{ optional: 'Custom optional' }} renderOptional>
-      This is a label
-    </FormControlLabel>,
-  );
-  const label = container.querySelector('label');
-
-  expect(label).toHaveStyleRule('content', "' (Custom optional)'", { modifier: '::after' });
+  expect(label?.lastChild).toHaveTextContent('*');
 });
--- a/packages/big-design/src/components/Input/spec.tsx
+++ b/packages/big-design/src/components/Input/spec.tsx
@@ -350,18 +350,9 @@ describe('error does not show when invalid type', () => {
   });
 });
 
-test('appends (optional) text to label if input is not required', () => {
-  const { container } = render(<Input label="Test Label" />);
+test('appends * text to label if input is required', () => {
+  const { container } = render(<Input label="Test Label" required />);
   const label = container.querySelector('label');
 
-  expect(label).toHaveStyleRule('content', "' (optional)'", { modifier: '::after' });
-});
-
-test('renders localized labels', () => {
-  const { container } = render(
-    <Input label="Test Label" localization={{ optional: 'opcional' }} />,
-  );
-  const label = container.querySelector('label');
-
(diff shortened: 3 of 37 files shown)


packages/big-design/src/components/Form/Label/Label.tsx
```diff
@@ -1,23 +1,25 @@
...
-  localization = defaultLocalization,
   style,
+  renderRequired,
   ...props
-}) => <StyledLabel localization={localization} {...props} />;
+}) => (
+  <StyledLabel {...props}>
+    {props.children}
+    {renderRequired && (
+      <Text aria-hidden="true" as="span" color="danger" marginLeft="xxSmall">
```
leeBigCommerce, line 20: rather than the marginLeft standing in for a space, we could do this to have an actual space between the label content and star:

```tsx
      <Text aria-hidden="true" as="span" color="danger">
        {' '}
        *
      </Text>
```

🍹

packages/big-design/src/components/Input/spec.tsx
```diff
@@ -350,18 +350,9 @@ describe('error does not show when invalid type', () => {
...
-});
-
-test('renders localized labels', () => {
-  const { container } = render(
-    <Input label="Test Label" localization={{ optional: 'opcional' }} />,
-  );
-  const label = container.querySelector('label');
-
-  expect(label).toHaveStyleRule('content', "' (opcional)'", { modifier: '::after' });
+  expect(label?.lastChild).toHaveTextContent('*');
```
leeBigCommerce, line 357: most of these can be simplified to just

```tsx
  render(<Input label="Test Label" required />);

  expect(screen.getByLabelText('Test Label*')).toBeInTheDocument();
  ```
  
That is if we stick with real element for the star, else we're back to the `toHaveStyleRule` method

File: packages/big-design/src/components/Form/Label/Label.tsx [hunk 3]
```diff
@@ -1,23 +1,25 @@
...
 
-const defaultLocalization: LabelLocalization = {
-  optional: 'optional',
-};
+import { StyledLabel } from './styled';
 
 export interface FormControlLabelProps extends ComponentPropsWithoutRef<'label'> {
-  renderOptional?: boolean;
-  localization?: LabelLocalization;
+  renderRequired?: boolean;
```
leeBigCommerce, line 8: we can probably live without the `render` prefix, it's kinda implied by this being a visual component

and it's not as if a field can be `required` without us rendering the star 🍹

packages/big-design/src/components/Form/Label/Label.tsx
leeBigCommerce: Also, it looks like the star is top aligned by virtue of the font family just having stars smaller and higher than its other  glyphs. A different font, i.e. 'monospace' would not preserve this styling.

Not that we're every likely to change the font, but we might want to consider using the `sup` tag to explicitly (if we're sticking with the physical element): https://developer.mozilla.org/en-US/docs/Web/HTML/Element/sup

packages/big-design/src/components/Form/Label/spec.tsx
```diff
@@ -19,20 +19,9 @@ test('does not forward styles', () => {
...
-test('appends custom (optional) text when passed', () => {
-  const { container } = render(
-    <FormControlLabel localization={{ optional: 'Custom optional' }} renderOptional>
-      This is a label
-    </FormControlLabel>,
-  );
-  const label = container.querySelector('label');
-
-  expect(label).toHaveStyleRule('content', "' (Custom optional)'", { modifier: '::after' });
+  expect(label?.lastChild).toHaveTextContent('*');
```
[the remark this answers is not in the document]
davelinke: A pseudo does not have the ability to be Aria-hidden, thus a screen reader would pick the asterisk

packages/big-design/src/components/Input/spec.tsx
davelinke: Thanks, will do!

packages/big-design/src/components/Form/Label/Label.tsx
davelinke: Yes, thanks, will do

packages/big-design/src/components/Form/Label/Label.tsx
davelinke: I will add the actual space as you suggested tho, even if it's aria hidden, it makes sense for the star to be visually separated in the case we would change how to implement margins and such. Good insight.

packages/big-design/src/components/Form/Label/spec.tsx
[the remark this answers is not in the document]
leeBigCommerce: Gotcha,

this definitely will be a breaking change then.

Along with the aesthetics and props changes, a lot of clients' tests will break by changing to use a real element:

```tsx
render(<Input label="Test Label" required />);

// this will PASS, getByRole honours "aria-hidden" and ignores it while selecting
expect(screen.getByRole('textbox', { name: 'Test Label' })).toBeInTheDocument();

// this will FAIL, getByLabelText does not honour "aria-hidden" and expects its content to be included
expect(screen.getByLabelText('Test Label')).toBeInTheDocument();

// this will PASS, existing uses of getByLabelText will need to include the star, or change to using getByRole
expect(screen.getByLabelText('Test Label *')).toBeInTheDocument();
```

any client who has made use of `getByLabelText` will need to change their tests upon moving to this new version of our inputs

it's not a lot of work, but we'll need to carefully inform them as it'll seem very confusing to see a bunch of test failures reporting that most of their form fields have gone missing

packages/big-design/src/components/Input/spec.tsx
leeBigCommerce: it seems like:

```tsx
expect(screen.getByRole('textbox', { name: 'Test Label' })).toBeInTheDocument();
```

might be more robust/future proof. `getByRole` is the only selector that will accurately ignore elements marked with "aria-hidden"

---

`getByLabelText` will work _if_ you add the star, but you'll want to check if you want to couple a specific test to the field being required or not

```tsx
expect(screen.getByLabelText('Test Label *')).toBeInTheDocument();
```

davelinke: I fixed many of the tests as per the suggestion, there were others that required the structure present.

packages/big-design/src/components/Form/Label/Label.tsx
```diff
@@ -1,23 +1,26 @@
...
   ...props
-}) => <StyledLabel localization={localization} {...props} />;
+}) => (
+  <StyledLabel {...props}>
+    {props.children}
+    {required && (
+      <Text aria-hidden="true" as="span" color="danger">
+        {' '}
+        *
+      </Text>
```
chanceaclark, line 23: 🍹 For my curiosity, do we intend this to be `sup` element or more inline with the text? Example of me changing it to the `<sup>` element:
Screenshot 2025-03-05 at 15 17 30

I would check the alignment in other browsers to ensure it's consistent.

chanceaclark (commented):
Found one more place we need to remove and `optional` reference:
https://github.com/bigcommerce/big-design/blob/61f4f09d1bdf77d1641a173febbbb73cf70d2858/packages/docs/PropTables/CounterPropTable.tsx#L90

chanceaclark:
@davelinke for reference in commit messages, if there is a breaking change, we usually add a `!` before the `:` to indicate such. If you can, make sure the title of the PR has it so when we squash and merge your commits that'll be the commit message:
>feat(component)!: change required behavior on form elements

packages/big-design/src/components/Form/Label/Label.tsx
davelinke: No intention of making that, it still is an aria hidden element, it has no value using a proper tag whatsoever

chanceaclark: More asking about the visual aspect of it, rather a semantic tag here.

davelinke: Oh, yes, visual is OK as is right now.

chanceaclark (approved):

PR #1638, opened 2025-02-06
Update codeowners

PascalZajac:
## What?

Update the `CODEOWNERS` file to reference our newly created team.

## Why?

The hope is that having a narrower group who are responsible for reviewing BD contributions will improve ownership / responsiveness thus facilitating better engagement and adoption.

## Testing/Proof
Screenshot 2025-02-06 at 9 58 14 pm

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,1 @@
-* @bigcommerce/frontend @bigcommerce/product-design
-packages/docs @bigcommerce/dev-docs-team @bigcommerce/frontend @bigcommerce/product-design
+* @bigcommerce/big-design-stewards @bigcommerce/team-platform-design


.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
-* @bigcommerce/frontend @bigcommerce/product-design
-packages/docs @bigcommerce/dev-docs-team @bigcommerce/frontend @bigcommerce/product-design
```
PascalZajac, line 2: I'm not sure if this line should actually change. It might be fine for the entire frontend group to be able to approve documentation changes?

golcinho: @PascalZajac It looks like we don’t actively contribute to the docs package, as updates mostly happen alongside changes to the big-design package. Given that, these changes seem valid.

Since the documentation is public and directly tied to BigDesign components, it makes sense for it to have the same focus and ownership as the big-design package.

golcinho (approved):

jorgemoya (approved):

PR #1656, opened 2025-03-13
feat(icons): add LightbulbIcon

davelinke:
## What?

Adding lightbulb icon.

## Why?

Requested by Patrycja Mielewczyk for New catalog feature


chanceaclark (commented):
Let's add a changeset so we can release this.

chanceaclark (approved):

PR #1661, opened 2025-03-19
feat(component): add StatusMessage component

davelinke:
## What?

Adding the Status message component for empty states and error pages

## Why?

This unifies the way in which we show empty states in features as well as full error pages

Screenshot 2025-03-19 at 11 06 42 AM
Screenshot 2025-03-19 at 11 06 26 AM
Screenshot 2025-03-19 at 11 06 09 AM

## Testing/Proof

<details>

```tsx
import { Box, Button, Flex, Grid, Panel, StatusMessage, StatusMessageProps } from '@bigcommerce/big-design';

const statusMessages: StatusMessageProps[] = [
  {
    heading: 'Page not found',
    message:
      'The page you are looking for might have been removed, had its name changed, or is temporarily unavailable.',
    variant: '404',
    // actions: [{ text: 'Go back to main page', variant: 'subtle' }],
    actions: <Button variant="subtle">Go back to main page</Button>,
  },
  {
    heading: 'You have no products yet',
    message: 'Start adding products to your store to see them here.',
    variant: 'info',
    // actions: [{ text: 'Add product', variant: 'secondary' }],
    actions: <Button variant="secondary">Add product</Button>,
  },
  {
    heading: 'There was an error connecting to the server',
    message: 'Please check your internet connection and try again.',
    variant: 'error',
    // actions: [{ text: 'Retry', variant: 'secondary' }],
    actions: <Button variant="secondary">Retry</Button>,
  },
  {
    heading: 'No results for search "Woot"',
    message: 'Try adjusting your search or filter to find what you are looking for.',
    variant: 'search',
    // actions: [{ text: 'Clear search', variant: 'secondary' }],
    actions: <Button variant="secondary">Clear search</Button>,
  },
  {
    heading: 'Internal server error',
    message:
      'The server encountered an internal error and was unable to complete your request. Please try again later.',
    variant: 'server-error',
    // actions: [{ text: 'Retry', variant: 'secondary' }],
    actions: <Button variant="secondary">Retry</Button>,
  },
  {
    heading: 'Operation successful',
    message: 'Your operation was completed successfully.',
    variant: 'success',
    // actions: [{ text: 'View details', variant: 'subtle' }],
    actions: <Button variant="subtle">View details</Button>,
  },
  {
    heading: 'Unauthorized access',
    message: 'You do not have the necessary permissions to access this resource.',
    variant: 'unauthorized',
    // actions: [{ text: 'Request access', variant: 'secondary' }],
    actions: <Button variant="secondary">Request access</Button>,
  },
  {
    heading: 'Quota limit exceeded',
    message:
      'You have exceeded your quota limit for this resource. Please try again in 32 minutes.',
    variant: 'warning',
    // actions: [{ text: 'Try again', variant: 'subtle' }],
    actions: <Button variant="subtle">Try again</Button>,
  },
];

const PageIllustrations: FunctionComponent = () => {
  return (
    <Grid gridColumns={{ mobile: '1fr', tablet: '1fr 1fr' }} gridGap="2rem" padding="xxLarge">
      <Panel>
        <Flex
          alignItems="center"
          flexDirection="column"
          flexGap="2rem"
          justifyContent="center"
          paddingVertical="xxLarge"
        >
          {statusMessages.map((statusMessage, index) => (
            <Box borderBottom="box" key={index} style={{ width: '100%' }}>
              <StatusMessage
                actions={statusMessage.actions}
                heading={statusMessage.heading}
                message={statusMessage.message}
                size="panel"
                variant={statusMessage.variant}
              />
            </Box>
          ))}
        </Flex>
      </Panel>

      <Flex
        alignItems="center"
        flexDirection="column"
        flexGap="2rem"
        justifyContent="center"
        padding="xxLarge"
      >
        {statusMessages.map((statusMessage, index) => (
          <Box borderBottom="box" key={index} style={{ width: '100%' }}>
            <StatusMessage
              actions={statusMessage.actions}
              heading={statusMessage.heading}
              message={statusMessage.message}
              size="page"
              variant={statusMessage.variant}
            />
          </Box>
        ))}
      </Flex>
    </Grid>
  );
};

export default PageIllustrations;
```

</details>


packages/big-design/src/components/StatusMessage/StatusMessage.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { StatusMessageIcons } from './StatusMessageIcons';
+import { StatusMessagePatterns } from './StatusMessagePatterns';
+import { StatusMessageIllustration, StyledStatusMessage } from './styled';
+
+export interface StatusMessageButtonProps extends Omit<ButtonProps, 'children'> {
+  text: string;
+}
+
+export interface StatusMessageProps {
+  actions?: StatusMessageButtonProps[];
```
leeBigCommerce, line 16: this `actions` prop is a bit open ended

- we're opinionated that they must be buttons, but not how many
- the type enforces only that `text` is required, but the click handler is optional
- there's no way to pass in a natural link (the client would need to make a button, which pretended to be a link)

perhaps we should either go fully agnostic, like our `Table`'s `actions` where the content is completely up to the client (we just position it): https://github.com/bigcommerce/big-design/blob/main/packages/big-design/src/components/Table/types.ts#L60

or become more opinionated and lock things down like we do in `Dropdown`: https://github.com/bigcommerce/big-design/blob/main/packages/big-design/src/components/Dropdown/types.ts#L39

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
```diff
@@ -0,0 +1,74 @@
...
+    | 'info'
+    | 'error'
+    | 'search'
+    | 'server-error'
+    | 'success'
+    | 'unauthorized'
+    | 'warning';
+}
+
+export const StatusMessage: React.FC<StatusMessageProps> = memo(
```
leeBigCommerce, line 31: unless we have a clear reason, best not to use `memo` 🍹

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
```diff
@@ -0,0 +1,74 @@
...
+  ({ actions, heading, message, size = 'panel', variant = 'info' }) => {
+    return (
+      <StyledStatusMessage size={size} variant={variant}>
+        <StatusMessageIllustration aria-hidden="true" size={size} variant={variant}>
+          <StatusMessagePatterns variant={variant} />
+          <StatusMessageIcons variant={variant} />
+        </StatusMessageIllustration>
```
leeBigCommerce, line 38: Given these all want the same props, there's probably a missing abstraction here

especially since you don't expect them to be ever used in a different order/configuration

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
```diff
@@ -0,0 +1,74 @@
...
+            size === 'page' ? (
+              <H1 margin="none" marginBottom="small">
+                {heading}
+              </H1>
+            ) : (
+              <H4 margin="none" marginBottom="xSmall">
+                {heading}
+              </H4>
+            )
+          ) : null}
```
leeBigCommerce, line 50:
```tsx
const Header = size === 'page' ? H1 : H4;
// etc etc
{heading !== undefined && (
  <Header margin="none" marginBottom="small">
    {heading}
  </Header>
)}
```

easier to read without the nested conditionals🍹

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
```diff
@@ -0,0 +1,74 @@
...
+          <Text margin="none">{message}</Text>
+        </Flex>
+
+        {actions ? (
```
leeBigCommerce, line 54: we can use a short-circuit here 🍹

packages/big-design/src/components/StatusMessage/StatusMessageIcons.tsx
```diff
@@ -0,0 +1,86 @@
...
+    | 'server-error'
+    | 'success'
+    | 'unauthorized'
+    | 'warning';
+}
+
+export const StatusMessageIcons: React.FC<StatusMessageIconProps> = ({ variant = 'info' }) => {
+  return (
+    <StyledStatusMessageIcons>
+      {variant === '404' && (
```
leeBigCommerce, line 20: worth considering a pattern like: https://github.com/bigcommerce/big-design/pull/1662#discussion_r2006580094 here too?

packages/big-design/src/components/StatusMessage/StatusMessageIcons.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <svg fill="none" height="120" role="img" width="120" xmlns="http://www.w3.org/2000/svg">
```
leeBigCommerce, line 21: seems `<svg fill="none" height="120" role="img" width="120" xmlns="http://www.w3.org/2000/svg">` is the same across all of them, might be worth moving the variant logic inward so it can be shared?

leeBigCommerce, line 21: a few magic numbers here (`120`), should we getting these/calculating these from the theme?

packages/big-design/src/components/StatusMessage/StatusMessagePatterns.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export const StatusMessagePatterns: React.FC<StatusMessagePatternProps> = ({
+  variant = 'info',
+}) => {
+  return (
+    <StyledStatusMessagePatterns>
+      {variant === 'error' ||
+      variant === 'server-error' ||
+      variant === 'warning' ||
+      variant === 'unauthorized' ? (
```
leeBigCommerce, line 25: seems like there's an implicit `positive/negative` grouping here?

in the absence of making that explicit, we can make things clearer with a map:

```tsx
type Variant =
  | '404'
  | 'info'
  | 'error'
  | 'search'
  | 'server-error'
  | 'success'
  | 'unauthorized'
  | 'warning';

export interface StatusMessagePatternProps {
  variant?: Variant;
}

const backgrounds: Record<Variant, FC> = {
  '404': DiamondBackground,
  error: CrossBackground,
  info: DiamondBackground,
  search: DiamondBackground,
  'server-error': CrossBackground,
  success: DiamondBackground,
  unauthorized: CrossBackground,
  warning: CrossBackground,
};

export const StatusMessagePatterns: React.FC<StatusMessagePatternProps> = ({
  variant = 'info',
}) => {
  const Background = backgrounds[variant];

  return (
    <StyledStatusMessagePatterns>
      <Background />
    </StyledStatusMessagePatterns>
  );
};
```

it's also much clearer if we give the different SVG component's their own name, given the code doesn't give anyone a hint as to what they'll look like

packages/big-design/src/components/StatusMessage/StatusMessagePatterns.tsx
```diff
@@ -0,0 +1,115 @@
...
+        <svg fill="none" height="120" role="img" width="120" xmlns="http://www.w3.org/2000/svg">
```
leeBigCommerce, line 26: more magic numbers, would we need to keep this `120` and the `120` within packages/big-design/src/components/StatusMessage/StatusMessageIcons.tsx in sync with each other? Do they represent the same thing?

packages/big-design/src/components/StatusMessage/styled.tsx
```diff
@@ -0,0 +1,105 @@
...
+    css`
+      background-image: radial-gradient(circle at center top, #ebb2ca 0%, #eee8fa 100%);
+    `}
+
+  ${({ variant }) =>
+    variant === 'success' &&
+    css`
+      background-image: radial-gradient(circle at center top, #78e4a3 0%, #f3fdec 100%);
+    `}
+`;
```
leeBigCommerce, line 72: similar suggestion regarding a map here too 🍹

packages/big-design/src/components/StatusMessage/styled.tsx
```diff
@@ -0,0 +1,105 @@
...
+    css`
+      transform: scale(1.5);
+      transform-origin: bottom;
+      margin-block-start: ${theme.helpers.remCalc(60)};
+    `}
+
+  ${({ variant }) =>
+    (variant === '404' || variant === 'info' || variant === 'search') &&
+    css`
+      background-image: radial-gradient(circle at center top, #72d8db 0%, #d9f9fa 100%);
```
leeBigCommerce, line 52: lots of hardcoded colours, should these come from the theme?

packages/big-design/src/components/StatusMessage/styled.tsx
```diff
@@ -0,0 +1,105 @@
...
+StatusMessageIllustration.defaultProps = {
+  theme: defaultTheme,
+  variant: 'info',
+  size: 'panel',
+};
+
+export const StyledStatusMessageIcons = styled.div`
+  position: relative;
+  width: 120px;
+  height: 120px;
```
leeBigCommerce, line 83: `120` is popping up a lot 😄

packages/big-design/src/components/StatusMessage/spec.tsx
```diff
@@ -0,0 +1,222 @@
...
+    />,
+  );
+
+  expect(screen.getByRole('button', { name: 'Action 1' })).toBeInTheDocument();
+  expect(screen.getByRole('button', { name: 'Action 2' })).toBeInTheDocument();
+});
+
+// variants
+
+// renders default variant
```
leeBigCommerce, line 52: you can use a `describe` block if you wan to specifically group tests 🍹

packages/big-design/src/components/StatusMessage/spec.tsx
```diff
@@ -0,0 +1,222 @@
...
+test('renders with the default variant', () => {
+  render(<StatusMessage message="Basic status message" />);
+
+  const figure = screen.getByRole('figure', { hidden: true });
+
+  expect(figure).toHaveStyle({
+    backgroundImage: `radial-gradient(circle at center top, #72d8db 0%, #d9f9fa 100%)`,
+  });
```
leeBigCommerce, line 60: do we want to use snapshots here? I'm not sure this bit of svg code means much to a test reader

if we're just wanting to protect ourselves from an accidental change, a snapshot should do it (goes for other similar tests too)

packages/big-design/src/components/StatusMessage/spec.tsx
```diff
@@ -0,0 +1,222 @@
...
+});
+
+test('renders with the 404 variant', () => {
```
leeBigCommerce, line 63: if we are just wanting to protect from accidental styling/aesthetic changes we could consider replacing these with a singular:

```tsx
const variants = [
  'info',
  'error',
  'success',
  'warning',
  'unauthorized',
  'search',
  'server-error',
  '404',
] as const;

test.each(variants)('renders with the %p variant', (variant) => {
  const { container } = render(<StatusMessage message="Basic status message" variant={variant} />);

});
```

leeBigCommerce (commented):
could do with someone more up on their CSS best practices to help with this one

I've done my best to forget the all 😛

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
```diff
@@ -0,0 +1,74 @@
...
+  heading?: string;
+  message: string;
+  size?: 'page' | 'panel';
+  variant?:
+    | '404'
+    | 'info'
+    | 'error'
+    | 'search'
+    | 'server-error'
+    | 'success'
```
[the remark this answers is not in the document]
davelinke: Indeed, a dead end success. Still we may give many options with the array of CTAs.

packages/big-design/src/components/StatusMessage/styled.tsx
davelinke: They were at some point, the recommendation for @chanceaclark was to hardcode them since this is the only place we'll use them so far.

packages/big-design/src/components/StatusMessage/styled.tsx
chanceaclark: `theme.helpers.remCalc` here please

packages/big-design/src/components/StatusMessage/styled.tsx
davelinke: Second the idea, it could also have a theme function to modify saturation and lightness based on a base color, which would cover other scenarios (hovers, actives, and so on)

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
davelinke: I second the first option to enable the usage of links and such.

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
davelinke: Removed

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
davelinke: Created a version that does not need the illustration component and it's children, but is tied to the main component.

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
davelinke: Added this suggestion

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
davelinke: When adding actions as a react node i had to keep the ternary

packages/big-design/src/components/StatusMessage/StatusMessageIcons.tsx
davelinke: With the new implementation this is now similar to what you proposed in that comment.

packages/big-design/src/components/StatusMessage/StatusMessageIcons.tsx
davelinke: New implementation holds the icons in the styled file. I did take the decision that for maintainability it is better to use the full SVG code in each one of the options to make it easier to add new icons for future developers.

davelinke: The recommendation was to keep these values local to this particular component, as it has slightly different value types than the ones held in the theme.

packages/big-design/src/components/StatusMessage/StatusMessagePatterns.tsx
davelinke: A similar execution has now been implemented.

packages/big-design/src/components/StatusMessage/StatusMessagePatterns.tsx
davelinke: The patterns incredibly have some other dimensions other than 120

packages/big-design/src/components/StatusMessage/styled.tsx
davelinke: New execution follows recommendations of https://github.com/bigcommerce/big-design/pull/1662#discussion_r2006580094

packages/big-design/src/components/StatusMessage/styled.tsx
davelinke: used `theme.helpers.remCalc` in the required spots

packages/big-design/src/components/StatusMessage/spec.tsx
davelinke: Removed Snapshots

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const StatusMessage: React.FC<StatusMessageProps> = ({
+  actions,
+  heading,
+  message,
+  size = 'panel',
+  variant = 'info',
+}) => {
+  const Header = size === 'page' ? H1 : H4;
+  const hasHeading = heading !== undefined && heading.trim().length > 0;
```
leeBigCommerce, line 35: the `heading.trim()` is a little redundant

either we trust clients to only add spacing if they want them (so we shouldn't check)

_or_ we're so opinionated that we trim any input we get before using it, e.g.

```tsx
const formattedHeading = heading?.trim();
const hasHeading = formattedHeading && formattedHeading.length > 0;
// etc etc
{formattedHeading && (
  <Header margin="none" marginBottom="small">
    {formattedHeading}
  </Header>
)}
```

the latter seems a bit much and I'm not sure we do this kind of thing elsewhere, so I'd drop the `trim` 🍹

packages/big-design/src/components/StatusMessage/spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+  expect(screen.getByText('Basic status message')).toBeInTheDocument();
+});
+
+test('renders status message illustration', () => {
+  render(<StatusMessage message="Basic status message" />);
+
+  expect(screen.getByRole('figure', { hidden: true })).toBeInTheDocument();
+});
+
+test('renders panel status message', () => {
```
leeBigCommerce, line 33: look like we're testing the heading rather the message here

packages/big-design/src/components/StatusMessage/spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+  render(<StatusMessage heading="Panel heading" message="Basic status message" />);
+
+  expect(screen.getByRole('heading', { level: 4, name: 'Panel heading' })).toBeInTheDocument();
+});
+
+test('renders page status message', () => {
```
leeBigCommerce, line 39: as above

also, if there difference is the expected heading level, we can call that more clearly in the test description,
we can also use a describe block to make the `page` size context really explicit, e.g.

```tsx
describe('when at "page" size', () => {
  test('renders the heading as a H1', () => {
    render(<StatusMessage heading="Page heading" message="Basic status message" size="page" />);

    expect(screen.getByRole('heading', { level: 1, name: 'Page heading' })).toBeInTheDocument();
  });
});
```

packages/big-design/src/components/StatusMessage/spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+  render(<StatusMessage heading="Page heading" message="Basic status message" size="page" />);
+
+  expect(screen.getByRole('heading', { level: 1, name: 'Page heading' })).toBeInTheDocument();
+});
+
+test('renders with button actions', () => {
```
leeBigCommerce, line 45: this is a little misleading, given it could be any component

the `Table` did a similar thing, but more component agnostic:

```tsx
  test('renders custom actions', () => {
    const { getByTestId } = render(
      <Table
        actions={<div data-testid="customAction">Test Action</div>}
        columns={columns}
        items={items}
      />,
    );

    const customAction = getByTestId('customAction');

    expect(customAction).toBeInTheDocument();
    expect(customAction).toBeVisible();
  });
```

packages/big-design/src/components/StatusMessage/spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+  expect(screen.getByRole('button', { name: 'Action 1' })).toBeInTheDocument();
+});
+
+test.each(variants)('renders with the %p variant', (variant) => {
+  const { container } = render(<StatusMessage message="Basic status message" variant={variant} />);
+
+  expect(container.firstChild).toMatchSnapshot();
+});
+
+// renders the panel size
```
leeBigCommerce, line 57: another hint we might want a `describe('when at "page" size', () => {` wrapper to group some of there

packages/big-design/src/components/StatusMessage/styled.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  // Encode special characters
+  const encodedSvg = encodeURIComponent(cleanedSvg).replace(/'/g, '%27').replace(/"/g, '%22');
+
+  // Return CSS-compatible background image string
+  return `url("data:image/svg+xml,${encodedSvg}")`;
+}
+
+const icons = {
+  '404': `<svg fill="none" height="120" role="img" width="120" xmlns="http://www.w3.org/2000/svg">
```
leeBigCommerce, line 22: @chanceaclark what would you recommend for dealing with all these SVGs?

we could go down the route of generating TSX from SVGs like we do with icons, then import them to the map here

the downside being we've then got that build step to work in

or we could just have these as hand-cranked TSX files (ideally each in their own file for legibility, `404.tsx` etc) and just map them here

the extra render-time leg work of the `svgToCssBackground` function and the large blobs of co-ordinates here are pretty cumbersome 🤔

leeBigCommerce: or could we just move them to the `icons` package? 

Though I guess we don't really want them used as regular icons by clients

davelinke: Exactly, these shouldn't be part of the icon set.

davelinke: In the prior review, we talked about generating an abstraction to wrap up the background pattern and the icon, who share same props. This was my attempt to make it as clean as possible by also considering the guidance from @icatalina to use the full SVG code to ease maintenance. CSS background covers this particular aspect to avoid absolute positioning of SVG elements which create a deeper DOM structure and extra CSS code.

chanceaclark: We could use something like https://www.npmjs.com/package/babel-plugin-inline-svg if we wanted to keep the SVGs out of `.ts` files. Worth exploring, otherwise we can just keep it simple and have them inline.

Using the babel plugin, we'd get `svgo` OOTB which means we could optimize the svgs during build – just a thought 🤔

packages/big-design/src/components/StatusMessage/StatusMessage.tsx
davelinke: dropped

packages/big-design/src/components/StatusMessage/spec.tsx
davelinke: applied example

packages/big-design/src/components/StatusMessage/spec.tsx
davelinke: yes, plus rendering basic status message is covered by the first test. Created other tests to test the difference in headings with the size prop

packages/big-design/src/components/StatusMessage/spec.tsx
davelinke: implemented

packages/big-design/src/components/StatusMessage/styled.tsx
davelinke: let's leave as is for the moment.

chanceaclark (approved):

PR #1662, opened 2025-03-20
Feat/lozenge

davelinke:
## What?

Add lozenge component to library

## Why?

The lozenge component aims to be the sole way to indicate a feature lifecycle, separating this function from the badge component which can be use sparingly to denote many status types.

Screenshot 2025-03-20 at 2 51 16 PM

## Testing/Proof

<details>

```tsx
import { Box, H1, H2, Lozenge } from '@bigcommerce/big-design';

const PageLozenge: FunctionComponent = () => {
  return (
    <Box padding="xxLarge">
      <H1>Lozenge</H1>
      <H2>Basic</H2>
      <Lozenge label="Legacy" variant="legacy" /> <br />
      <br />
      <Lozenge label="New" variant="new" />
      <br />
      <br />
      <Lozenge label="Early Access" variant="new" />
      <br />
      <br />
      <Lozenge label="Alpha" variant="alpha" />
      <br />
      <br />
      <Lozenge label="Beta" variant="beta" />
      <br />
      <br />
      <Lozenge label="Deprecated" variant="deprecated" />
      <br />
      <br />
      <H2>With tooltip</H2>
      <Lozenge label="Beta" tooltipContent="This feature is in open beta" variant="beta" />
    </Box>
  );
};

export default PageLozenge;
```

</details>

### DIFF
--- a/packages/big-design-patterns/src/components/Header/Header.tsx
+++ b/packages/big-design-patterns/src/components/Header/Header.tsx
@@ -120,11 +120,12 @@ export interface HeaderProps {
   badge?: BadgeProps;
   description?: Description;
   icon?: ReactNode;
+  lozenge?: ReactNode;
   title: string;
 }
 
 export const Header = memo(
-  ({ actions, backLink, badge, description, icon = null, title }: HeaderProps) => {
+  ({ actions, backLink, badge, description, icon = null, lozenge, title }: HeaderProps) => {
     return (
       <Flex as="header" flexDirection="row" flexWrap="wrap">
         {backLink && (
@@ -144,6 +145,7 @@ export const Header = memo(
             <Flex alignItems="center" flexDirection="row" flexGap="1rem" marginBottom="xxSmall">
               {icon}
               <H1 marginBottom="none">{title}</H1>
+              {lozenge}
               {badge ? <Badge {...excludeMarginProps(badge)} /> : null}
             </Flex>
             <Description description={description} />
--- a/packages/big-design-patterns/src/components/Header/Header.tsx
+++ b/packages/big-design-patterns/src/components/Header/Header.tsx
@@ -9,6 +9,8 @@ import {
   Flex,
   FlexItem,
   H1,
+  Lozenge,
+  LozengeProps,
   Text,
 } from '@bigcommerce/big-design';
 import { ArrowBackIcon } from '@bigcommerce/big-design-icons';
@@ -120,7 +122,7 @@ export interface HeaderProps {
   badge?: BadgeProps;
   description?: Description;
   icon?: ReactNode;
-  lozenge?: ReactNode;
+  lozenge?: LozengeProps;
   title: string;
 }
 
@@ -145,7 +147,7 @@ export const Header = memo(
             <Flex alignItems="center" flexDirection="row" flexGap="1rem" marginBottom="xxSmall">
               {icon}
               <H1 marginBottom="none">{title}</H1>
-              {lozenge}
+              {lozenge ? <Lozenge {...lozenge} /> : null}
               {badge ? <Badge {...excludeMarginProps(badge)} /> : null}
             </Flex>
             <Description description={description} />
--- a/packages/big-design-patterns/src/components/Header/Header.tsx
+++ b/packages/big-design-patterns/src/components/Header/Header.tsx
@@ -147,8 +147,8 @@ export const Header = memo(
             <Flex alignItems="center" flexDirection="row" flexGap="1rem" marginBottom="xxSmall">
               {icon}
               <H1 marginBottom="none">{title}</H1>
-              {lozenge ? <Lozenge {...lozenge} /> : null}
-              {badge ? <Badge {...excludeMarginProps(badge)} /> : null}
+              {lozenge && <Lozenge {...lozenge} />}
+              {badge && <Badge {...excludeMarginProps(badge)} />}
             </Flex>
             <Description description={description} />
           </FlexItem>
--- a/packages/big-design/src/components/Lozenge/Lozenge.tsx
+++ b/packages/big-design/src/components/Lozenge/Lozenge.tsx
@@ -1,28 +1,32 @@
 import { InfoIcon } from '@bigcommerce/big-design-icons';
-import React, { ComponentPropsWithoutRef, memo } from 'react';
+import React from 'react';
 
 import { MarginProps } from '../../helpers';
+import { Tooltip } from '../Tooltip';
 
 import { StyledLozenge } from './styled';
 
-export interface LozengeProps extends ComponentPropsWithoutRef<'span'>, MarginProps {
+export interface LozengeProps extends MarginProps {
   label: string;
-  tooltipIcon?: boolean;
+  tooltipContent?: string;
   variant?: 'alpha' | 'beta' | 'deprecated' | 'legacy' | 'new';
 }
 
-export const Lozenge: React.FC<LozengeProps> = memo(({ className, style, label, ...props }) =>
-  typeof label === 'string' ? (
+export const Lozenge: React.FC<LozengeProps> = (props) => {
+  const LozengeElement = (
     <StyledLozenge {...props}>
-      {label}
-      {props.tooltipIcon ? (
-        <>
-          {' '}
-          <InfoIcon aria-hidden="true" size="large" />
-        </>
-      ) : null}
+      {props.label}
+      {props.tooltipContent ? <InfoIcon aria-hidden="true" size="large" /> : null}
     </StyledLozenge>
-  ) : null,
-);
+  );
+
+  return props.tooltipContent ? (
+    <Tooltip placement="auto" trigger={LozengeElement}>
+      {props.tooltipContent}
+    </Tooltip>
+  ) : (
+    LozengeElement
+  );
+};
 
 Lozenge.displayName = 'Lozenge';
--- a/packages/big-design/src/components/Lozenge/Lozenge.tsx
+++ b/packages/big-design/src/components/Lozenge/Lozenge.tsx
@@ -1,0 +1,29 @@
+import React, { ComponentPropsWithoutRef, memo } from 'react';
+import { InfoIcon } from '@bigcommerce/big-design-icons';
+
+import { MarginProps } from '../../helpers';
+
+import { StyledLozenge } from './styled';
+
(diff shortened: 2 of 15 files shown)


packages/big-design/src/components/Lozenge/styled.tsx
```diff
@@ -0,0 +1,64 @@
...
+  line-height: ${({ theme }) => theme.lineHeight.small};
+  text-align: center;
+  padding-block: ${({ theme }) => theme.spacing.xxSmall};
+  user-select: none;
+  vertical-align: middle;
+  padding-inline: ${({ theme, tooltipIcon }) =>
+    tooltipIcon ? `${theme.spacing.small} ${theme.spacing.xxSmall}` : theme.spacing.small};
+  white-space: nowrap;
+
+  ${({ theme, variant }) =>
```
leeBigCommerce, line 24: rather than repeating all these functions, we could either use a single function with a `switch` statement or we could create a little map of the styles like this:

```tsx
const variantStyles = {
  alpha: css`
    background-color: ${({ theme }) => theme.colors.warning20};
    color: ${({ theme }) => theme.colors.secondary70};
  `,
  beta: css`
    background-color: ${({ theme }) => theme.colors.primary20};
    color: ${({ theme }) => theme.colors.primary50};
  `,
  deprecated: css`
    color: ${({ theme }) => theme.colors.danger70};
    background-color: ${({ theme }) => theme.colors.danger20};
  `,
  legacy: css`
    background-color: ${({ theme }) => theme.colors.secondary30};
    color: ${({ theme }) => theme.colors.secondary70};
  `,
  new: css`
    background-color: ${({ theme }) => theme.colors.success20};
    color: ${({ theme }) => theme.colors.success70};
  `,
};
```

```tsx
// used within the main css definition like this
${({ variant = 'new' }) => variantStyles[variant]};
```

packages/big-design-patterns/src/components/Header/Header.tsx
```diff
@@ -120,11 +120,12 @@ export interface HeaderProps {
   badge?: BadgeProps;
   description?: Description;
   icon?: ReactNode;
+  lozenge?: ReactNode;
```
leeBigCommerce, line 123: this would be open to abuse, client's could put just about another component in here and it would render

if you want to be truly restrictive, take a look at https://github.com/bigcommerce/big-design/blob/main/packages/big-design/src/components/Input/Input.tsx#L94-L97

leeBigCommerce (commented):
This seems like a component for a very specific, narrow use case - i.e. it'll only ever be useful to clients who are describing software (unless the ignore our variant names and appropriate the colours for their own needs)

Which puts it as an outlier amongst our very use case agnostic components (`Table`, `Button`, `Input` etc etc)

packages/big-design/src/components/Lozenge/Lozenge.tsx
```diff
@@ -0,0 +1,28 @@
...
+import React, { ComponentPropsWithoutRef, memo } from 'react';
+
+import { MarginProps } from '../../helpers';
+
+import { StyledLozenge } from './styled';
+
+export interface LozengeProps extends ComponentPropsWithoutRef<'span'>, MarginProps {
+  label: string;
+  tooltipIcon?: boolean;
+  variant?: 'alpha' | 'beta' | 'deprecated' | 'legacy' | 'new';
```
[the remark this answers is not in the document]
davelinke: This is the one that's going to be most used when a new feature is launched and you have the option not to upgrade immediately or roll back. We also have defined the use of this terminology in our UX Writing standards.

packages/big-design/src/components/Lozenge/spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+});
+
+test('render new Lozenge', () => {
+  const { container } = render(<Lozenge label="New" variant="new" />);
+
+  expect(container.firstChild).toMatchSnapshot();
+  expect(container.firstChild).toHaveStyle(`background-color: ${theme.colors.success20}`);
+});
+
+test("doesn't render if label prop is invalid", () => {
```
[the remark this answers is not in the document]
davelinke: Fair, just mimicking the badge component.

packages/big-design/src/components/Lozenge/styled.tsx
davelinke: Great insight, will do.

packages/big-design/src/components/Lozenge/Lozenge.tsx
```diff
@@ -0,0 +1,28 @@
+import { InfoIcon } from '@bigcommerce/big-design-icons';
+import React, { ComponentPropsWithoutRef, memo } from 'react';
+
+import { MarginProps } from '../../helpers';
+
+import { StyledLozenge } from './styled';
+
+export interface LozengeProps extends ComponentPropsWithoutRef<'span'>, MarginProps {
```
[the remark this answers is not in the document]
davelinke: Same here, mimicking the label component.

packages/big-design/src/components/Lozenge/Lozenge.tsx
```diff
@@ -0,0 +1,28 @@
...
+  label: string;
+  tooltipIcon?: boolean;
```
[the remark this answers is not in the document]
davelinke: This is true, and will solve other issues.

packages/big-design/src/components/Lozenge/Lozenge.tsx
```diff
@@ -0,0 +1,28 @@
...
+  variant?: 'alpha' | 'beta' | 'deprecated' | 'legacy' | 'new';
+}
+
+export const Lozenge: React.FC<LozengeProps> = memo(({ className, style, label, ...props }) =>
+  typeof label === 'string' ? (
+    <StyledLozenge {...props}>
+      {label}
+      {props.tooltipIcon ? (
```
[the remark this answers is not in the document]
davelinke: Short circuiting is not allowed in our lint definition.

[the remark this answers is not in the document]
leeBigCommerce: it's not allowed for a `string` or a `number`, you're hinging off of a `boolean`

you can try it and run `pnpm run lint` locally, that snipped from above should pass

packages/big-design/src/components/Lozenge/styled.tsx
leeBigCommerce: we can define `variantStyles` outside of `StyledLozenge` as just a const within the file 🍹

packages/big-design/src/components/Lozenge/spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+});
+
+// add a test to check if the tooltipIcon prop is working
+test('render Lozenge with tooltip', async () => {
+  const { container } = render(<Lozenge label="Lozenge" tooltipContent="Tooltip content" />);
+
+  expect(container.firstChild).toMatchSnapshot();
+  expect(container.firstChild).toHaveStyle(
+    `padding-inline: ${theme.spacing.small} ${theme.spacing.xxSmall}`,
+  );
```
leeBigCommerce, line 72: this styling should already be defended by the `Tooltip` component's tests

repeating it here could lead to a scenario where `Tooltip` changes how its creates spacing and this test fails here

I'd recommend leaving that responsibility tested in `Tooltip`

instead, we can just focus on the behaviour that `Lozenge` is responsible for wiring up:

```tsx
test('render Lozenge with tooltip', async () => {
  render(<Lozenge label="Lozenge" tooltipContent="Tooltip content" />);

  const lozenge = screen.getByText('Lozenge');

  userEvent.hover(lozenge);

  const tooltip = await screen.findByText('Tooltip content');

  expect(tooltip).toBeVisible();

  userEvent.unhover(lozenge);

  await waitFor(() => expect(tooltip).not.toBeVisible());
});
```

packages/big-design/src/components/Lozenge/Lozenge.tsx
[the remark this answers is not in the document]
davelinke: It is now a string tho.

packages/big-design/src/components/Lozenge/styled.tsx
davelinke: this happens when you write it like that, is there any way around it?

Screenshot 2025-03-31 at 4 41 02 PM

leeBigCommerce: looks like we've got a couple of things there, mostly caused by us trying to re-use the same props for both `StyledLozenge` and `Lozenge`

1. `StyledLozenge` requires `tooltipContent`, but we're not passing it in (probably from when I suggested de-structuring props)

However, what it really wants is boolean to know if or not if there is a tooltip (it does not care about any tooltips actual text content), so maybe giving `StyledLozenge` it's own props with a `hasTooltip: boolean` would be more explicit

2. `StyledLozenge.defaultProps`

By trying to share the props definition and setting defaults, we're effectively making it impossible for `variant` to every be `undefined`. However, the props definition says it _could_ be - which is why I suggested the defaulting to `new` here:

`${({ variant = 'new' }) => variantStyles[variant]};`

We're basically adding a default value, after we've already added a default value another way.

I'd recommend giving `StyledLozenge` its own props and removing the `StyledLozenge.defaultProps` for any optional values.

davelinke: Thanks, this was really helpful.

packages/big-design/src/components/Lozenge/styled.tsx
```diff
@@ -0,0 +1,54 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
+import styled, { css } from 'styled-components';
+
+import { Box } from '../Box';
+
+import { LozengeProps } from './Lozenge';
+
+interface StyledLozengeProps extends Omit<LozengeProps, 'label' | 'tooltipContent'> {
+  hasTooltip?: boolean;
```
leeBigCommerce, line 9: might as well be mandatory, given your parent component is always providing it

optional props are more useful when the client is external (e.g. someone using BigDesign in their own app) as it saves them entering values when there's safe/expected defaults

given this component API isn't directly public facing we can have it more clearly state what it needs

then we don't the need the defaulting (line 53) below either

leeBigCommerce (approved):

PR #1729, opened 2025-08-22
feat(theme): add negative spacing utilities to spacing system

davelinke:
## What?

Adding negative values to spacing definitions

## Why?

We have had a disconnect for very long on some components in code Vs their appearance in patterns, which is that many things, especially tables can be bound to the edges of elements such as panels, modals, popovers, etc.

Negative spacing gives us the possibility to overcome this issue by inserting negative margin values on inner elements.

FireShot Capture 642 - BigDesign -  localhost

## Testing/Proof

<details>

```tsx
import { Box, H1, H2, Button, Popover, Text, Modal, Panel, Table } from '@bigcommerce/big-design';
import React, { FunctionComponent, useRef, useState } from 'react';

const PageLozenge: FunctionComponent = () => {
  const [isOpen, setIsOpen] = useState(false);

  const [modalOpen, setModalOpen] = useState(false);
  const buttonRef = useRef(null);

  return (
    <Box padding="xxLarge">
      <H1>Negative Spacing</H1>

      <H2>On panel</H2>
      <Panel
        action={{
          variant: 'secondary',
          text: 'Button',
          onClick: () => {
            // Do some action
          },
        }}
        badge={{
          label: 'primary',
          variant: 'primary',
        }}
        description="This is the panel's optional description."
        header="Panel header"
      >
        <Box marginHorizontal={{ mobile: 'mediumN', tablet: 'xLargeN' }}>
          <Table
            columns={[
              {
                header: 'Sku',
                hash: 'sku',
                tooltip: 'Header tooltip',
                render: ({ sku }) => sku,
              },
              { header: 'Name', hash: 'name', render: ({ name }) => name },
              { header: 'Stock', hash: 'stock', render: ({ stock }) => stock },
            ]}
            items={[
              { sku: 'SM13', name: '[Sample] Smith Journal 13', stock: 25 },
              { sku: 'DPB', name: '[Sample] Dustpan & Brush', stock: 34 },
              { sku: 'OFSUC', name: '[Sample] Utility Caddy', stock: 45 },
              { sku: 'CLC', name: '[Sample] Canvas Laundry Cart', stock: 2 },
              { sku: 'CGLD', name: '[Sample] Laundry Detergent', stock: 29 },
            ]}
          />
        </Box>
      </Panel>

      <H2>On Popover</H2>
      <>
        <Button onClick={() => setIsOpen((prev) => !prev)} ref={buttonRef}>
          Open Popover
        </Button>
        <Popover
          anchorElement={buttonRef.current as HTMLElement | null}
          isOpen={isOpen}
          label="Example Popover"
          onClose={() => setIsOpen(false)}
          placement="bottom-start"
        >
          <Box marginHorizontal="mediumN">
            <Text>This is the popover content bound to the edges of the popover</Text>
          </Box>
        </Popover>
      </>
      <br />
      <br />
      <H2>On Modal</H2>
      <>
        <Button onClick={() => setModalOpen(true)}>Open Modal</Button>

        <Modal
          actions={[
            {
              text: 'Cancel',
              variant: 'subtle',
              onClick: () => setModalOpen(false),
            },
            { text: 'Apply', onClick: () => setModalOpen(false) },
          ]}
          closeOnClickOutside={false}
          closeOnEscKey={true}
          header="Modal title"
          isOpen={modalOpen}
          onClose={() => setModalOpen(false)}
        >
          <Box marginHorizontal={{ tablet: 'xLargeN', mobile: 'mediumN' }}>
            <Text>
              Ea tempor sunt amet labore proident dolor proident commodo in exercitation ea nulla
              sunt pariatur. Nulla sunt ipsum do eu consectetur exercitation occaecat labore aliqua.
              Aute elit occaecat esse ea fugiat esse. Reprehenderit sunt ea ea mollit commodo tempor
              amet fugiat.
            </Text>
          </Box>
        </Modal>
      </>
    </Box>
  );
};

export default PageLozenge;

```

</details>


packages/big-design-theme/src/system/spacing.ts
chanceaclark, line 1: I still think we should go back to a more programmatic approach of generating the negative values based on the existing positive ones. If we need to add or change spacing values, it'll make it easier. I left some comments as I disabled some of the eslint rules covering type assertions. The more I think about it, it's okay to use type assertions here as it's a more controlled environment.

 I am a little rusty with TypeScript tbh, so if there is a better way someone else can offer, I am all ears.

```ts
import { remCalc } from '../helpers';

export interface BaseSpacing {
  none: 0;
  xxSmall: string;
  xSmall: string;
  small: string;
  medium: string;
  large: string;
  xLarge: string;
  xxLarge: string;
  xxxLarge: string;
}

type NegativeSpacing = {
  [K in `${Exclude<keyof BaseSpacing, 'none'>}N`]: string;
};

export type Spacing = BaseSpacing & NegativeSpacing;

export const createSpacing = (): Spacing => {
  const spacing: BaseSpacing = {
    none: 0,
    xxSmall: remCalc(4),
    xSmall: remCalc(8),
    small: remCalc(12),
    medium: remCalc(16),
    large: remCalc(20),
    xLarge: remCalc(24),
    xxLarge: remCalc(32),
    xxxLarge: remCalc(48),
  };

  const { none: _none, ...positives } = spacing;

  // We are disabling a few rules here around type assertions because we need to
  // assert the type of the keys and the accumulator in the reduce function.
  // We know these assertions are safe because we are deriving the negative keys
  // directly from the positive keys. The program shouldn't be able to modify
  // the existing objects at runtime in a way that would make these assertions unsafe.
  // If it does, we have bigger problems.
  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
  const positiveKeys = Object.keys(positives) as Array<keyof typeof positives>;

  const negativeSpacing = positiveKeys.reduce((acc, key) => {
    acc[`${key}N`] = `-${positives[key]}`;

    return acc;
    // We have to assert the type here because passing it through the generic parameter
    // causes TypeScript to complain that this type (`{}`) is not the same as NegativeSpacing.
    // eslint-disable-next-line @typescript-eslint/prefer-reduce-type-parameter, @typescript-eslint/consistent-type-assertions
  }, {} as NegativeSpacing);

  return { ...spacing, ...negativeSpacing };
};
```

packages/docs/pages/spacing.tsx
```diff
@@ -20,17 +36,28 @@ const SpacingPage = () => {
 
       <Panel header="Overview" headerId="overview">
         <Text>
-          BigDesign core theme provides a pre-defined set of spacing values that are primarily used
-          with any of the <Code primary>padding</Code> and <Code primary>margin</Code> props. There
-          are a few ways we can consume these values.
+          BigDesign’s theme provides a pre-defined spacing scale, primarily used with{' '}
+          <Code primary>padding</Code> and <Code primary>margin</Code> props. In addition to the
+          positive tokens (e.g. <Code>small</Code>, <Code>medium</Code>), the theme now exposes
```
chanceaclark, line 41:
```suggestion
          <Code primary>padding</Code> and <Code primary>margin</Code> props. In addition to the
          positive tokens (e.g. <Code>small</Code>, <Code>medium</Code>), the theme exposes
```

packages/docs/pages/spacing.tsx
```diff
@@ -1,9 +1,25 @@
...
+const POSITIVE_KEYS = [
+  'xxSmall',
+  'xSmall',
+  'small',
+  'medium',
+  'large',
+  'xLarge',
+  'xxLarge',
+  'xxxLarge',
+];
```
chanceaclark, line 21: Can we try an just filter this out via the object so if we add new spacing values it automatically reflects?
```ts
Object.keys(theme.spacing).filter(key => !key.endsWith("N"))...
```

.changeset/twelve-radios-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'@bigcommerce/big-design-theme': minor
```
chanceaclark, line 2: You can remove the other changeset and just modify the existing one.
```suggestion
'@bigcommerce/big-design-theme': minor
'@bigcommerce/docs': minor
```

packages/big-design-theme/src/system/spacing.ts
davelinke: take a look again to see if the current state is ok

icatalina:
could you share some examples of those disconnects? I worry negative margins could be abused and used to implement things that go against the standard...

davelinke:
> could you share some examples of those disconnects? I worry negative margins could be abused and used to implement things that go against the standard...

Sure! mostly on all panel features where you either need a table or a list, All components and patterns are designed for tables to touch the boundaries of panels horizontally, modals and other card-like components.

This just enables developers not to have to directly code that css themselves. as it is the case nowadays, where they just skip this styling or create styled components to make it happen, driving inconsistency further.

chanceaclark (approved):

PR #1736, opened 2025-09-03
fix(component): add portal to dropdown

MichalDobry:
## What?

- updated Dropdown component to render its popover menu inside a React Portal using react-popper
- replaced useRef with setReferenceElement/setPopperElement for stable positioning with Popper

## Why?

Previously, the Dropdown menu was rendered inline, which could cause issues with clipping (overflow: hidden, z-index conflicts, or positioning within scrollable containers). 
By rendering the menu into a portal (document.body or a custom container), we ensure consistent layering and positioning of the dropdown regardless of DOM hierarchy.

https://github.com/user-attachments/assets/84a16c95-7623-49dc-ab0e-b10f956fbf41

Result of suggestions: 
Modal before:
Screenshot 2025-09-08 at 11 50 59

Modal indexed higher after:
Screenshot 2025-09-08 at 12 15 29

## Testing/Proof

- updated and ran unit tests for the Dropdown component

### DIFF
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -129,9 +129,17 @@ export const Dropdown = memo(
       modifiers: [
         {
           name: 'eventListeners',
-          options: { scroll: isOpen, resize: isOpen },
+          options: {
+            scroll: isOpen,
+            resize: isOpen,
+          },
         },
-        { name: 'offset', options: { offset: [0, 4] } },
+        {
+          name: 'offset',
+          options: {
+            offset: [0, 4],
+          },
+        },
       ],
       placement,
       strategy: positionFixed ? 'fixed' : 'absolute',
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -4,10 +4,12 @@ import React, {
   isValidElement,
   memo,
   useCallback,
+  useEffect,
   useId,
   useMemo,
-  useRef,
+  useState,
 } from 'react';
+import { createPortal } from 'react-dom';
 import { usePopper } from 'react-popper';
 
 import { Box } from '../Box';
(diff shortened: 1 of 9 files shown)


chanceaclark (commented):
I am overall fine with this approach, but I think we should just blanket change this for all consumers rather than opt-in. This is more of an underlying refactor and shouldn't break any consumers.

An aside, I found one situation where this is not great, but it has to deal with z-indexes rather than the portal location. I also had to force that dropdown open using DOM selectors/events so no one should run into this unless they are truly hacking something together:

Image

Can you just make sure to pressure test this a little more with other components that deal with z-indexes?

packages/big-design/src/components/Dropdown/Dropdown.tsx
```diff
@@ -140,6 +142,53 @@ export const Dropdown = memo(
       strategy: positionFixed ? 'fixed' : 'absolute',
     });
 
+    const [mounted, setMounted] = useState(false);
+    const [portalContainer, setPortalContainer] = useState<Element | null>(null);
+    const [zIndexValue, setZIndexValue] = useState<number | 'popover' | 'modal'>('popover');
```
chanceaclark, line 147: I should have clarified in my last comment that no real action was needed for the modal + dropdown being visible, as that use-case would be rare. The only use-case I can think of is if a developer pops a modal up without a user action but that should be bad practice.

With that being said, I don't think we should be doing this z-index calculation magic, just adds complexity that's not entirely need. I just wanted to litmus test with different common user interactions to make sure it behaves nicely.

MichalDobry: Thanks Clark, that makes sense. I’ll simplify the implementation

chanceaclark:
For the sake of not going back and forth again, I ended up pushing some cleanup to this PR. That means we need another reviewer from @bigcommerce/big-design-stewards to move forward.

golcinho (approved):
Good job, thank you @MichalDobry and @chanceaclark ! Great changes 👍

PR #1772, opened 2025-10-29
fix(component): move to a stable ref for Dropdowns

chanceaclark:
## What?

Updates the Dropdown component to use stable refs (`useRef`) instead of state (`useState`) for managing references to the toggle button and popper elements.

## Why?

The Dropdown component was using `useState` to store references to DOM elements (`referenceElement` and `popperElement`), which caused unnecessary re-renders whenever these references were set. This is an anti-pattern in React because:

1. **Performance**: Setting state triggers re-renders, but ref updates do not. DOM element references don't need to trigger re-renders since they're only used by the Popper.js library for positioning calculations.

2. **Stability**: Using `useRef` provides a stable reference that persists across renders without causing additional render cycles. This is the recommended React pattern for storing DOM references.

3. **Popper.js Integration**: The `usePopper` hook works correctly with both state-based refs and stable refs, but stable refs are more appropriate since we only need to access the DOM nodes, not react to their changes.

This change improves the component's performance by eliminating unnecessary re-renders while maintaining the same functionality.

Screenshot 2025-10-29 at 11 45 25

## Testing/Proof

The existing test suite continues to pass without modifications, confirming that the behavior remains unchanged. The Dropdown component's functionality is preserved:
- The popper positioning still works correctly
- The toggle button interaction remains functional  
- Open/close state management is unaffected
- Portal rendering continues to work as expected

This refactor only changes the internal implementation detail of how DOM references are stored, moving from a stateful approach to a more performant ref-based approach.

---

This pull request description was generated with the assistance of AI. Portions of the code and/or implementation ideas in this PR may also have been created or influenced by AI tools.

### DIFF
--- a/.changeset/shy-moons-relate.md
+++ b/.changeset/shy-moons-relate.md
@@ -1,0 +1,5 @@
+---
+'@bigcommerce/big-design': patch
+---
+
+Update the Dropdown ref from a state value to the stable ref.
--- a/packages/big-design/src/components/Dropdown/Dropdown.tsx
+++ b/packages/big-design/src/components/Dropdown/Dropdown.tsx
@@ -6,7 +6,7 @@ import React, {
   useCallback,
   useId,
   useMemo,
-  useState,
+  useRef,
 } from 'react';
 import { createPortal } from 'react-dom';
 import { usePopper } from 'react-popper';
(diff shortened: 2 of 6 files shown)


golcinho (approved):
Amazing🎃

PR #1779, opened 2025-11-05
fix(component): dropdown items click not firing inside modal

golcinho:
## What?/Why?

This PR fixes an issue where interactive components rendered via portals (e.g., Dropdown, Select) inside a Modal could not be clicked or interacted with.
The problem occurred because the modal's focus-trap considered these portaled elements to be outside the modal, blocking all pointer events and preventing handlers such as onItemClick from firing.

The fix introduces a focused change to the modal’s focus trap configuration to safely allow pointer events on menus and listboxes without breaking accessibility or modal isolation.

**Changes:**
- Updated the createFocusTrap configuration to include a custom `allowOutsideClick` predicate.
- This predicate now allows clicks on elements (or their descendants) with `role="menu"` or `role="listbox"`.
- Retains full focus trap functionality:
  - Focus is still restricted within the modal for keyboard navigation.
  - Backdrop clicks and Escape key continue to close the modal when enabled.
- No API or behavioral changes for other components.

**Before:**

https://github.com/user-attachments/assets/02c86a62-0dda-42ff-86fc-9871902250fe

**After:**

https://github.com/user-attachments/assets/32f00a8e-59ba-4dcf-8f1b-3de9c1dcce66

## Testing/Proof

- Local testing (see video)
- Additional unit tests

### DIFF
--- a/.changeset/heavy-llamas-join.md
+++ b/.changeset/heavy-llamas-join.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/big-design": patch
+---
+
+fix(component): dropdown items click not firing inside modal
--- a/packages/big-design/src/components/Dropdown/spec.tsx
+++ b/packages/big-design/src/components/Dropdown/spec.tsx
@@ -6,6 +6,7 @@ import React from 'react';
 import 'jest-styled-components';
 
 import { Button } from '../Button';
+import { Modal } from '../Modal';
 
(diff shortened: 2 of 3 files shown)


bc-maksym-konohorov (approved):

apostoliukG (approved):

chanceaclark (approved):
Awesome! Thanks @golcinho!

PR #1846, opened 2026-04-03
fix(component): [worksheet] prevent paste into disabled parent rows

apostoliukG:
## What?

Fixes the paste handler (`useCopyPaste`) in the Worksheet component to correctly prevent pasting values into disabled parent rows via Cmd+C / Cmd+V.

## Why?

The paste handler was comparing `cell.rowIndex + 1` against `disabledRows`, but `disabledRows` contains actual **row IDs** (arbitrary strings/numbers), not 1-based indices. This meant:

1. The disabled row check in `updateItems` effectively never matched unless row IDs happened to be sequential numbers starting at 1.
2. `setSelectedCells` was called **before** any disabled row check, so the pasted value would visually appear in the cell even on a disabled row.

### Changes:
- Add `!disabledRows.includes(rows[selectedCells[0].rowIndex]?.id)` guard **before** `setSelectedCells` and `updateItems` run
- Replace `cell.rowIndex + 1` with `rows[cell.rowIndex]?.id` in the `updateItems` filter so multi-row paste also correctly skips disabled rows
- Add a null check for `columns[selectedCells[0].columnIndex]` and handle missing `cellsToUpdate` entries safely

N/A — behavioral fix, no visual changes.

## Testing/Proof

- Verified TypeScript compiles cleanly (`tsc --noEmit`)
- Manual verification: copying a cell value and pasting into a disabled parent row should now be blocked (no visual update, no data change)
- Multi-row paste across a mix of enabled/disabled rows correctly skips disabled rows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/big-design/src/components/Worksheet/hooks/useCopyPaste/useCopyPaste.ts
+++ b/packages/big-design/src/components/Worksheet/hooks/useCopyPaste/useCopyPaste.ts
@@ -121,14 +121,20 @@ export const useCopyPasteHandler = () => {
         if (
           copiedCells[0] &&
           selectedCells[0] &&
-          isPasteAllowed(copiedCells[0], selectedCells[0], columns[selectedCells[0].columnIndex])
+          columns[selectedCells[0].columnIndex] &&
+          !disabledRows.includes(rows[selectedCells[0].rowIndex]?.id) &&
+          isPasteAllowed(copiedCells[0], selectedCells[0], columns[selectedCells[0].columnIndex]!)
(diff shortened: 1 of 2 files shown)


bc-maksym-konohorov (approved):

RomanKrasinskyi (approved):

golcinho (approved):

chanceaclark (commented):
Make sure to add a changeset in order to release this fix.

PR #1855, opened 2026-05-28
feat(component): update collapse component with a new API

bc-nick:
## What?

Update `Collapse` component with a new API

## Why?

Rewrites `Collapse` as a headless compound component. The root provides state via Context; consumers compose `Collapse.Trigger `and `Collapse.Panel` anywhere in the subtree.

https://github.com/user-attachments/assets/9df539c2-bcde-4fea-bc76-5d724cbb63b9

## Testing/Proof

Manual testing

### DIFF
--- a/packages/big-design/src/components/Collapse/Collapse.tsx
+++ b/packages/big-design/src/components/Collapse/Collapse.tsx
@@ -27,7 +27,7 @@ export const Collapse: CollapseComponent = ({
   const wasDisabled = useRef(disabled);
 
   const isControlled = isOpen !== undefined;
-  const [uncontrolledOpen, setUncontrolledOpen] = useState(initiallyOpen);
+  const [uncontrolledOpen, setUncontrolledOpen] = useState(disabled ? false : initiallyOpen);
   const isCollapseOpen = isControlled ? isOpen : uncontrolledOpen;
 
   const triggerId = useId();
@@ -38,15 +38,14 @@ export const Collapse: CollapseComponent = ({
 
     wasDisabled.current = disabled;
 
-    if (isControlled) {
-      return;
-    }
+    if (justDisabled && isCollapseOpen) {
+      if (!isControlled) {
+        setUncontrolledOpen(false);
+      }
 
-    if (justDisabled && uncontrolledOpen) {
-      setUncontrolledOpen(false);
       onCollapseChange?.(false);
     }
-  }, [disabled, isControlled, uncontrolledOpen, onCollapseChange]);
+  }, [disabled, isControlled, isCollapseOpen, onCollapseChange]);
 
   const toggle = useCallback(() => {
     if (disabled) {
--- a/packages/big-design/src/components/Collapse/Collapse.tsx
+++ b/packages/big-design/src/components/Collapse/Collapse.tsx
@@ -1,61 +1,68 @@
-import { ExpandMoreIcon } from '@bigcommerce/big-design-icons';
-import React, { useId, useState } from 'react';
+import React, { useCallback, useEffect, useId, useState } from 'react';
 
-import { Box } from '../Box';
-
-import { StyledButton } from './styled';
+import { CollapsePanel, CollapsePanelProps } from './CollapsePanel';
+import { CollapseTrigger, CollapseTriggerProps } from './CollapseTrigger';
+import { CollapseContext } from './useCollapseContext';
 
 export interface CollapseProps {
   children?: React.ReactNode;
-  title: string;
-  initiallyOpen?: boolean;
+  isCollapseOpen?: boolean;
+  defaultOpen?: boolean;
   onCollapseChange?(isOpen: boolean): void;
+  disabled?: boolean;
+  triggerProps?: Omit<CollapseTriggerProps, 'title'>;
+  panelProps?: Omit<CollapsePanelProps, 'children'>;
 }
 
-export const Collapse: React.FC<CollapseProps> = ({
+type CollapseComponent = React.FC<CollapseProps> & {
+  Trigger: typeof CollapseTrigger;
+  Panel: typeof CollapsePanel;
+};
+
+export const Collapse: CollapseComponent = ({
   children,
-  title,
+  isCollapseOpen,
+  defaultOpen = false,
   onCollapseChange,
-  initiallyOpen = false,
+  disabled,
 }) => {
-  const [isOpen, setIsOpen] = useState(initiallyOpen);
+  const isControlled = isCollapseOpen !== undefined;
+  const [uncontrolledOpen, setUncontrolledOpen] = useState(disabled ? false : defaultOpen);
+  const isOpen = isControlled ? isCollapseOpen : uncontrolledOpen;
+
   const triggerId = useId();
   const panelId = useId();
 
-  const handleTitleClick = () => {
-    const nextIsOpen = !isOpen;
+  useEffect(() => {
+    if (disabled && isOpen) {
+      if (!isControlled) {
+        setUncontrolledOpen(false);
+      }
 
-    setIsOpen(nextIsOpen);
-
-    if (typeof onCollapseChange === 'function') {
-      onCollapseChange(nextIsOpen);
+      onCollapseChange?.(false);
     }
-  };
+  }, [disabled, isOpen, isControlled, onCollapseChange]);
+
+  const toggle = useCallback(() => {
+    if (disabled) {
+      return;
+    }
+
+    const next = !isOpen;
+
+    if (!isControlled) {
+      setUncontrolledOpen(next);
+    }
+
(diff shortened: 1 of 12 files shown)


packages/big-design/src/components/Collapse/Collapse.tsx
```diff
@@ -1,61 +1,68 @@
...
+import { CollapsePanel, CollapsePanelProps } from './CollapsePanel';
+import { CollapseTrigger, CollapseTriggerProps } from './CollapseTrigger';
+import { CollapseContext } from './useCollapseContext';
 
 export interface CollapseProps {
   children?: React.ReactNode;
-  title: string;
-  initiallyOpen?: boolean;
+  isCollapseOpen?: boolean;
+  defaultOpen?: boolean;
```
golcinho: why we rename this prop? `initiallyOpen` looks better to me

packages/big-design/src/components/Collapse/Collapse.tsx
```diff
@@ -1,61 +1,68 @@
...
-import { StyledButton } from './styled';
+import { CollapsePanel, CollapsePanelProps } from './CollapsePanel';
+import { CollapseTrigger, CollapseTriggerProps } from './CollapseTrigger';
+import { CollapseContext } from './useCollapseContext';
 
 export interface CollapseProps {
   children?: React.ReactNode;
-  title: string;
-  initiallyOpen?: boolean;
+  isCollapseOpen?: boolean;
```
golcinho: the name of the prop can be simpler, just `isOpen`.

packages/big-design/src/components/Collapse/Collapse.tsx
```diff
@@ -1,61 +1,68 @@
...
+  defaultOpen?: boolean;
   onCollapseChange?(isOpen: boolean): void;
+  disabled?: boolean;
+  triggerProps?: Omit<CollapseTriggerProps, 'title'>;
+  panelProps?: Omit<CollapsePanelProps, 'children'>;
```
golcinho: unused props

packages/big-design/src/components/Collapse/Collapse.tsx
```diff
@@ -1,61 +1,68 @@
...
-    const nextIsOpen = !isOpen;
+  useEffect(() => {
+    if (disabled && isOpen) {
+      if (!isControlled) {
+        setUncontrolledOpen(false);
+      }
+
+      onCollapseChange?.(false);
+    }
+  }, [disabled, isOpen, isControlled, onCollapseChange]);
```
golcinho: in controlled mode this emits `onCollapseChange(false)` whenever `disabled && isOpen`, including on initial mount of `<Collapse isCollapseOpen disabled>`, and can re-fire each render when `onCollapseChange` is an inline function.

we only manage the uncontrolled case, and only on the disabled transition, for example:

```
const wasDisabled = useRef(disabled);

useEffect(() => {
  const justDisabled = disabled && !wasDisabled.current;
  wasDisabled.current = disabled;

  // Controlled consumers own their state — don't override it.
  if (isControlled) {
    return;
  } 

  if (justDisabled && uncontrolledOpen) {
    setUncontrolledOpen(false);
    onCollapseChange?.(false);
  }
}, [disabled, isControlled, uncontrolledOpen, onCollapseChange]);
```

packages/big-design/src/components/Collapse/Collapse.tsx
```diff
@@ -1,61 +1,68 @@
...
-  title,
+  isCollapseOpen,
+  defaultOpen = false,
   onCollapseChange,
-  initiallyOpen = false,
+  disabled,
 }) => {
-  const [isOpen, setIsOpen] = useState(initiallyOpen);
+  const isControlled = isCollapseOpen !== undefined;
+  const [uncontrolledOpen, setUncontrolledOpen] = useState(disabled ? false : defaultOpen);
```
golcinho: `defaultOpen` silently dropped when mounted `disabled`

```
  const [uncontrolledOpen, setUncontrolledOpen] = useState(defaultOpen);
  // the disabled auto-close effect (prev comment) already closes it while disabled,
  // so the defaultOpen intent is preserved once re-enabled.
```

packages/big-design/src/components/Collapse/CollapseTrigger/CollapseTrigger.tsx
```diff
@@ -0,0 +1,37 @@
+import { ExpandMoreIcon } from '@bigcommerce/big-design-icons';
+import React from 'react';
+
+import { MarginProps } from '../../../helpers';
+import { useCollapseContext } from '../useCollapseContext';
+
+import { StyledButton } from './styled';
+
+export interface CollapseTriggerProps extends MarginProps {
+  title: string | React.ReactNode;
```
golcinho: why we update the type `title: string` with `title: string | React.Node`?

when `title` is a `ReactNode`, the chevron icon gets `title={undefined}`, dropping the accessible name it previously always had.

packages/big-design/src/components/Collapse/spec.tsx
```diff
@@ -176,5 +187,193 @@ test('onCollapseChange is called', async () => {
...
+      <Collapse.Panel backgroundColor="secondary20" padding="medium">
+        <Text>Content</Text>
+      </Collapse.Panel>
+    </Collapse>,
+  );
+
+  const panel = screen.getByRole('region');
+
+  expect(panel).toHaveStyle('padding: 1rem');
+});
```
golcinho, line 294: test only checks padding

packages/big-design/src/components/Collapse/CollapseTrigger/CollapseTrigger.tsx
bc-nick: Went back to `title` as the `string` type only. Thanks

packages/big-design/src/components/Collapse/spec.tsx
bc-nick: added additional check

packages/big-design/src/components/Collapse/spec.tsx
```diff
@@ -176,5 +187,197 @@ test('onCollapseChange is called', async () => {
...
+      <Collapse.Panel>
+        <Text>Content</Text>
+      </Collapse.Panel>
+    </Collapse>,
+  );
+
+  expect(screen.getByRole('button')).toHaveStyle('margin-top: 1.25rem');
+});
+
+test('controlled mode: isCollapseOpen drives the open state', async () => {
```
golcinho:
```suggestion
test('controlled mode: isOpen drives the open state', async () => {
```

packages/docs/PropTables/CollapsePropTable.tsx
```diff
@@ -4,23 +4,73 @@ import { Prop, PropTable, PropTableWrapper } from '../components';
...
     types: 'boolean',
-    description: 'Defines if panel with content is visible by default.',
+    defaultValue: 'false',
+    description: 'Initial open state when uncontrolled.',
+  },
+  {
+    name: 'isOpen',
+    types: 'boolean',
+    description:
+      'Controlled open state. When provided, the component is controlled and ignores isOpen.',
```
golcinho:
```suggestion
      'Controlled open state. When provided, the component is controlled and ignores initiallyOpen.',
```

packages/big-design/src/components/Collapse/spec.tsx
golcinho: @bc-nick the test is still named `'panel applies backgroundColor and padding'` but only checks `padding`

packages/big-design/src/components/Collapse/Collapse.tsx
```diff
@@ -1,61 +1,75 @@
...
+  isOpen,
   initiallyOpen = false,
+  onCollapseChange,
+  disabled,
 }) => {
-  const [isOpen, setIsOpen] = useState(initiallyOpen);
+  const wasDisabled = useRef(disabled);
+
+  const isControlled = isOpen !== undefined;
+  const [uncontrolledOpen, setUncontrolledOpen] = useState(initiallyOpen);
```
golcinho: Sorry, my bad. Please restore previous

```suggestion
  const [uncontrolledOpen, setUncontrolledOpen] = useState(disabled ? false : initiallyOpen);
```

and please re-add the regression guard

```
  test('panel stays hidden when disabled and initiallyOpen', () => {
    render(
      <Collapse disabled initiallyOpen>
        <Collapse.Trigger title="title" />
        <Collapse.Panel>
          <Text>Content</Text>
        </Collapse.Panel>
      </Collapse>,
    );

    expect(screen.getByRole('region', { hidden: true })).not.toBeVisible();
  });
```

golcinho (approved):

PR #1857, opened 2026-06-08
feat(component): Update Checkbox component based on new requirements

bc-nick:
## What?

Update Checkbox component with a new API

## Why?

Add `collapseOptions`, `img`, and `disableWhenUnchecked` props to `Checkbox`. `collapseOptions` renders a collapsible `Collapse.Trigger`/`Collapse.Panel` beneath the checkbox row, `img` displays an image beside the label, and `disableWhenUnchecked` disables the collapse while the checkbox is unchecked.

https://github.com/user-attachments/assets/249b262c-af54-4447-a296-5cf46868f656

## Testing/Proof

Manual testing

### DIFF
--- a/packages/big-design/src/components/Checkbox/Checkbox.tsx
+++ b/packages/big-design/src/components/Checkbox/Checkbox.tsx
@@ -11,11 +11,15 @@ import React, {
 
 import { typedMemo, warning } from '../../utils';
 import { Badge, BadgeProps } from '../Badge';
+import { Box } from '../Box';
+import { Collapse, CollapsePanelProps, CollapseProps, CollapseTriggerProps } from '../Collapse';
 import { FormControlDescription, FormControlDescriptionLinkProps } from '../Form';
 
 import { CheckboxLabel } from './Label';
 import {
   CheckboxContainer,
+  CheckboxContentContainer,
+  CheckboxImgContainer,
   CheckboxLabelContainer,
   HiddenCheckbox,
   StyledCheckbox,
@@ -27,6 +31,9 @@ interface Props {
   label: React.ReactNode;
   description?: CheckboxDescription | string;
   badge?: BadgeProps;
+  collapseOptions?: CheckboxCollapse;
+  disableWhenUnchecked?: boolean;
+  img?: CheckboxImg;
 }
 
 interface CheckboxDescription {
@@ -34,11 +41,22 @@ interface CheckboxDescription {
   link?: FormControlDescriptionLinkProps;
 }
 
+export interface CheckboxImg {
+  src: string;
+  alt?: string;
+}
+
 interface PrivateProps {
   forwardedRef: Ref<HTMLInputElement>;
 }
 
 export type CheckboxProps = Props & ComponentPropsWithoutRef<'input'>;
+
+export interface CheckboxCollapse {
+  collapse?: CollapseProps;
+  trigger: CollapseTriggerProps;
+  panel?: CollapsePanelProps;
+}
 
 const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
   checked,
@@ -50,10 +68,15 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
   label,
   forwardedRef,
   style,
+  collapseOptions,
+  img,
+  disableWhenUnchecked,
   badge,
   onClick,
   ...props
 }) => {
+  const isVerticalCenter = Boolean(collapseOptions || img);
+
   const uniqueCheckboxId = useId();
   const labelId = useId();
   const id = props.id ? props.id : uniqueCheckboxId;
@@ -100,8 +123,13 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
     return <FormControlDescription link={link}>{text}</FormControlDescription>;
   }, [description]);
 
-  return (
-    <CheckboxContainer className={className} onClick={onClick} style={style}>
+  const checkboxRow = (
+    <CheckboxContainer
+      className={className}
+      isVerticalCenter={isVerticalCenter}
+      onClick={onClick}
+      style={style}
+    >
       <HiddenCheckbox
         checked={checked}
         disabled={disabled}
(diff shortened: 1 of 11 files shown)


packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -138,12 +166,37 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
...
+  }
+
+  return (
+    <Collapse
+      {...collapseOptions.collapse}
+      disabled={disableWhenUnchecked ? !checked : collapseOptions.collapse?.disabled}
+    >
+      <Box>
+        {checkboxRow}
+        <Collapse.Panel backgroundColor="secondary20" padding="medium" {...collapseOptions.panel} />
```
golcinho, line 196: `panel` is optional in `CheckboxCollapse` but we always render the panel. when only `trigger` is passed this renders an empty `role="region"` with `secondary20` bg + `medium` padding - an empty grey box that expands, and a landmark screen readers announce.

also these `secondary20`/`medium` literals feel like they belong on `Collapse.Panel` as a default, not hardcoded per consumer here.

```suggestion
        {collapseOptions.panel && (
          <Collapse.Panel backgroundColor="secondary20" padding="medium" {...collapseOptions.panel} />
        )}
```

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -138,12 +166,37 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
...
   );
+
+  if (!collapseOptions) {
+    return checkboxRow;
+  }
+
+  return (
+    <Collapse
+      {...collapseOptions.collapse}
+      disabled={disableWhenUnchecked ? !checked : collapseOptions.collapse?.disabled}
```
golcinho, line 192: `disabled` is spread after `{...collapseOptions.collapse}`, so when `disableWhenUnchecked` is set and the box is `checked` this resolves to `false` and silently drops a consumer's `collapse={{ disabled: true }}`. let's compose instead of override:

```suggestion
      disabled={collapseOptions.collapse?.disabled || (disableWhenUnchecked && !checked)}
```

separately: with a controlled `collapse={{ isOpen: true }}` + unchecked, the panel still shows since `CollapsePanel` keys off `isOpen` not `disabled`, so `disableWhenUnchecked` won't keep it closed.

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -138,12 +166,37 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
...
+        isVerticalCenter={isVerticalCenter}
+      >
+        <CheckboxLabelContainer>
+          {img && <CheckboxImgContainer alt={img.alt ?? ''} src={img.src} />}
+          <Box>
+            {renderedLabel}
+            {renderedDescription}
+          </Box>
+        </CheckboxLabelContainer>
+        {collapseOptions && <Collapse.Trigger marginVertical="none" {...collapseOptions.trigger} />}
```
golcinho, line 180: `onClick` is on `CheckboxContainer` and the trigger renders inside it. `Collapse.Trigger`'s `toggle` doesn't `stopPropagation`, so clicking the trigger bubbles up and fires the consumer's row `onClick` on every expand/collapse.

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -27,19 +31,33 @@ interface Props {
...
 interface PrivateProps {
   forwardedRef: Ref<HTMLInputElement>;
 }
 
 export type CheckboxProps = Props & ComponentPropsWithoutRef<'input'>;
 
+export interface CheckboxCollapse {
+  collapse?: CollapseProps;
+  trigger: CollapseTriggerProps;
+  panel?: CollapsePanelProps;
```
golcinho, line 58: `panel` optional but we always render it (see the panel comment below) — let's make the type and behavior agree: either required, or rendered conditionally.

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -138,12 +166,37 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
       >
         {!checked && isIndeterminate ? <RemoveIcon /> : <CheckIcon />}
       </StyledCheckbox>
-      <CheckboxLabelContainer hasContent={Boolean(label) || Boolean(description)}>
-        {renderedLabel}
-        {renderedDescription}
-      </CheckboxLabelContainer>
+      <CheckboxContentContainer
```
golcinho, line 169: these wrappers render for every checkbox, not just the img/collapse ones — that's the table/tableNext/worksheet snapshot churn. the inner `Box` has no props either. can we keep the extra nesting only for the img/collapse case?

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -138,12 +166,37 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
...
+        hasContent={Boolean(label || description || img || collapseOptions)}
```
golcinho, line 170: `label` is required so `hasContent` is effectively always true (only false on the empty-label path we already `warning()` about). could just be a constant margin.

packages/big-design/src/components/Checkbox/styled.tsx
```diff
@@ -10,17 +10,35 @@ interface StyledCheckboxProps {
   disabled?: boolean;
 }
 
-export interface StyledLabelProps {
-  hidden?: boolean;
-  disabled?: boolean;
-}
+export const CheckboxImgContainer = styled.img`
```
golcinho, line 13: this duplicates `FileUploader`'s `StyledImage` (same 40x40 `styled.img`). also no `object-fit`, so a non-square `src` gets squashed into 40x40. reuse a shared image + add `object-fit: cover`?

golcinho:
stepping back from the line comments - the `collapseOptions` shape feels like the root of most of them. we *just* rewrote `Collapse` into a headless compound component (`Collapse.Trigger`/`Collapse.Panel`), and here we immediately re-wrap it as a nested props bag (`{ collapse, trigger, panel }`). that flattening is what produces the `disabled` override, the always-rendered empty `panel`, and the trigger-click bubbling - they're all symptoms of `Checkbox` swallowing the whole Collapse API instead of composing it.

it also changes what `Checkbox` *is*: depending on `collapseOptions` it either returns a checkbox row or wraps itself in a `Collapse` + panel. a form primitive owning a disclosure widget feels off.

and `disableWhenUnchecked` reads as consumer logic (`disabled={!checked}`) living in a generic primitive - coupling checked-state to collapse-disabled belongs at the call site.

wonder if this should be a composed pattern rather than props on the core `Checkbox`:

```tsx
<Collapse disabled={!checked}>
  <Checkbox label="…" img={…} />
  <Collapse.Trigger title="…" />
  <Collapse.Panel>…</Collapse.Panel>
</Collapse>
```

keeps `Checkbox` a checkbox, drops the spread-override bugs, and lets consumers put arbitrary content between trigger and panel. `img` seems fine to keep on `Checkbox` imo. thoughts?

bc-nick:
Closed this one since another one was created as the most actual: https://github.com/bigcommerce/big-design/pull/1858

PR #1858, opened 2026-06-15
feat(component): adds a new CollapsibleCheckbox pattern to big-design-patterns and extends the base Checkbox with image support to power it

bc-nick:
## What?

Adds a new CollapsibleCheckbox pattern to big-design-patterns and extends the base Checkbox with image support to power it.

## Why?

  `big-design` — Checkbox

  - Added an optional img prop ({ src: string; alt?: string }) that renders a 40×40 thumbnail between the checkbox and its label.
  - Added an isVerticalCenter prop on CheckboxContainer; the row is vertically centered automatically when an image is present, and can be toggled externally otherwise.
  - New CheckboxImgContainer styled  uses object-fit: cover (no distortion for non-square images) and flex-shrink: 0 (holds its size next to long labels).

  `big-design-patterns` — new CollapsibleCheckbox

  - Pairs a Checkbox with a Collapse trigger + panel: the checkbox and trigger sit on one centered, full-width row (space-between), with the panel below.
  - Controlled component — checked is required and paired with onChange. The panel is disabled while unchecked by default, overridable via collapseProps.disabled.
  - Extends CheckboxProps (so label, description, img, badge, etc. pass through) and exposes triggerTitle, children, collapseProps, triggerProps, panelProps.

  `docs`

  - New CollapsibleCheckbox page with Basic, With image, and Initially open examples, plus a props table (and a "CollapsibleCheckbox" entry in the Patterns nav).
  - Documented the new Checkbox img / isVerticalCenter props in the Checkbox prop table.

This is a component library so we love visually looking at changes! If this applies to your pull request, show us your hard work in action.

## Testing/Proof

  - big-design and big-design-patterns typecheck clean; CollapsibleCheckbox tests pass (5/5).
  - docs typechecks clean.
  - manual testing

https://github.com/user-attachments/assets/177cec61-ad8d-49a5-8c11-25dc87efd544

### DIFF
--- a/packages/big-design-patterns/src/components/CollapsibleCheckbox/CollapsibleCheckbox.tsx
+++ b/packages/big-design-patterns/src/components/CollapsibleCheckbox/CollapsibleCheckbox.tsx
@@ -6,13 +6,15 @@ import {
   CollapseProps,
   CollapseTriggerProps,
 } from '@bigcommerce/big-design';
-import React, { ReactNode } from 'react';
+import React, { ChangeEventHandler, ReactNode } from 'react';
 
 import { StyledCheckboxRow } from './styled';
 
 export interface CollapsibleCheckboxProps extends CheckboxProps {
   /** Controls the checked state. The component is controlled — pair it with `onChange`. */
   checked: boolean;
+  /** Change handler for the checkbox. Required because the component is controlled. */
+  onChange: ChangeEventHandler<HTMLInputElement>;
   /** Title shown on the collapse trigger button. */
   triggerTitle: string;
   /** Content revealed inside the collapse panel. */
--- a/packages/big-design-patterns/src/components/CollapsibleCheckbox/CollapsibleCheckbox.tsx
+++ b/packages/big-design-patterns/src/components/CollapsibleCheckbox/CollapsibleCheckbox.tsx
@@ -1,0 +1,49 @@
+import {
+  Checkbox,
+  CheckboxProps,
+  Collapse,
+  CollapsePanelProps,
+  CollapseProps,
+  CollapseTriggerProps,
+} from '@bigcommerce/big-design';
+import React, { ReactNode } from 'react';
+
+import { StyledCheckboxRow } from './styled';
+
+export interface CollapsibleCheckboxProps extends CheckboxProps {
+  /** Controls the checked state. The component is controlled — pair it with `onChange`. */
+  checked: boolean;
+  /** Title shown on the collapse trigger button. */
+  triggerTitle: string;
+  /** Content revealed inside the collapse panel. */
+  children: ReactNode;
+  /**
+   * Props forwarded to the underlying `Collapse`. By default the panel is disabled while the
+   * checkbox is unchecked; pass `disabled` here to override that behavior.
+   */
+  collapseProps?: Omit<CollapseProps, 'children'>;
+  /** Props forwarded to the underlying `Collapse.Trigger`. */
+  triggerProps?: Omit<CollapseTriggerProps, 'title'>;
+  /** Props forwarded to the underlying `Collapse.Panel`. */
+  panelProps?: Omit<CollapsePanelProps, 'children'>;
+}
+
+export const CollapsibleCheckbox = ({
+  checked,
+  triggerTitle,
+  children,
+  collapseProps,
+  triggerProps,
+  panelProps,
+  ...checkboxProps
+}: CollapsibleCheckboxProps) => (
+  <Collapse disabled={!checked} {...collapseProps}>
+    <StyledCheckboxRow alignItems="center" justifyContent="space-between">
+      <Checkbox {...checkboxProps} checked={checked} />
+      <Collapse.Trigger marginVertical="none" {...triggerProps} title={triggerTitle} />
+    </StyledCheckboxRow>
+    <Collapse.Panel backgroundColor="secondary20" padding="medium" {...panelProps}>
+      {children}
+    </Collapse.Panel>
+  </Collapse>
+);
--- a/packages/big-design-patterns/src/components/CollapsibleCheckbox/spec.tsx
+++ b/packages/big-design-patterns/src/components/CollapsibleCheckbox/spec.tsx
@@ -1,0 +1,96 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+
+import { CollapsibleCheckbox } from './CollapsibleCheckbox';
+
+test('renders the checkbox, trigger and (hidden) panel', () => {
+  render(
+    <CollapsibleCheckbox
+      checked={false}
+      label="Enable feature"
+      onChange={jest.fn()}
+      triggerTitle="Configure"
+    >
+      Panel content
+    </CollapsibleCheckbox>,
+  );
+
+  expect(screen.getByRole('checkbox')).toBeInTheDocument();
+  expect(screen.getByRole('button')).toHaveTextContent('Configure');
+  expect(screen.getByText('Panel content')).not.toBeVisible();
+});
+
+test('disables the trigger by default while unchecked', () => {
+  render(
+    <CollapsibleCheckbox
+      checked={false}
+      label="Enable feature"
+      onChange={jest.fn()}
+      triggerTitle="Configure"
+    >
+      Panel content
+    </CollapsibleCheckbox>,
+  );
+
+  expect(screen.getByRole('button')).toBeDisabled();
+});
+
+test('collapseProps.disabled overrides the default gating', () => {
+  render(
+    <CollapsibleCheckbox
+      checked={true}
+      collapseProps={{ disabled: true }}
+      label="Enable feature"
+      onChange={jest.fn()}
+      triggerTitle="Configure"
+    >
+      Panel content
+    </CollapsibleCheckbox>,
+  );
+
+  expect(screen.getByRole('button')).toBeDisabled();
+});
+
+test('trigger is enabled and the panel can open while checked', () => {
+  render(
+    <CollapsibleCheckbox
+      checked={true}
+      label="Enable feature"
+      onChange={jest.fn()}
+      triggerTitle="Configure"
+    >
+      Panel content
+    </CollapsibleCheckbox>,
+  );
+
+  const trigger = screen.getByRole('button');
+
+  expect(trigger).toBeEnabled();
+
+  fireEvent.click(trigger);
+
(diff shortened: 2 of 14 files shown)


packages/docs/PropTables/CheckboxPropTable.tsx
```diff
@@ -75,6 +75,26 @@ const checkboxProps: Prop[] = [
...
+    types: '{ src: string; alt?: string }',
+    description: (
+      <>
+        Renders a thumbnail between the <Code primary>Checkbox</Code> and its label. When provided,
+        the row is vertically centered.
+      </>
+    ),
+  },
+  {
+    name: 'isVerticalCenter',
```
golcinho: docs and the changeset list `isVerticalCenter` as a `Checkbox` prop, but `Props` only adds `img` — `Checkbox.tsx` hardcodes `isVerticalCenter={Boolean(img)}` on the container and never reads it from props. so `<Checkbox isVerticalCenter />` is a type error and does nothing. either expose it for real (`isVerticalCenter?: boolean` on `Props`, defaulting to `Boolean(img)`) or drop it from the prop table + changeset.

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -138,6 +151,7 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
       >
         {!checked && isIndeterminate ? <RemoveIcon /> : <CheckIcon />}
       </StyledCheckbox>
+      {img ? <CheckboxImgContainer alt={img.alt ?? ''} src={img.src} /> : null}
```
golcinho, line 153: `alt` defaults to `''`, i.e. decorative. that's the right default when the label already names the row (the docs example label is `Enable payment provider`), but `img.alt` is optional so a meaningful thumbnail silently gets no accessible name — `WCAG 1.1.1`. fine as long as "decorative unless the caller sets `alt`" is the intended contract; worth saying so in the `img` prop description.

packages/big-design-patterns/src/components/CollapsibleCheckbox/spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+  );
+
+  const checkbox = screen.getByRole('checkbox');
+
+  expect(checkbox).toBeChecked();
+
+  fireEvent.click(checkbox);
+
+  expect(onChange).toHaveBeenCalledTimes(1);
+});
```
golcinho, line 98: the core behavior isn't covered: check → open the panel → uncheck should collapse it (`Collapse` fires its `justDisabled` effect and closes). the spec asserts the disabled gating and that it opens while `checked`, but not that unchecking an open panel closes it. add that regression guard — it's the whole point of the component.

packages/big-design/src/components/Checkbox/styled.tsx
```diff
@@ -19,8 +19,16 @@ export const CheckboxLabelContainer = styled.div<{ hasContent?: boolean }>`
   margin-left: ${({ hasContent, theme }) => (hasContent ? theme.spacing.xSmall : 0)};
 `;
 
-export const CheckboxContainer = styled.div`
-  align-items: flex-start;
+export const CheckboxImgContainer = styled.img`
```
golcinho, line 22: 🍹 this duplicates `FileUploader`'s `StyledImage` (same 40x40 `styled.img`). yours is the better version — it adds `object-fit: cover` + `flex-shrink: 0` — so consider promoting one shared thumbnail rather than keeping two.

golcinho (commented):
solid pattern overall — it reuses the new `Collapse` compound API cleanly and the gating (`disabled={!checked}`, overridable via `collapseProps`) reads well.

main concerns before merge:
- `isVerticalCenter` is documented (prop table + changeset) as a `Checkbox` prop, but it isn't one — it's hardcoded to `Boolean(img)` and never read from props. wire it up or drop it from the docs/changeset.
- the headline behavior — unchecking an open `CollapsibleCheckbox` should collapse the panel — isn't tested. the spec covers the disabled gating but not that transition.
- one real `a11y` point on the `img` `alt` default to confirm, plus a couple of 🍹 nits (reuse + requiring `onChange`).

packages/big-design-patterns/src/components/CollapsibleCheckbox/CollapsibleCheckbox.tsx
```diff
@@ -0,0 +1,49 @@
...
+  Collapse,
+  CollapsePanelProps,
+  CollapseProps,
+  CollapseTriggerProps,
+} from '@bigcommerce/big-design';
+import React, { ReactNode } from 'react';
+
+import { StyledCheckboxRow } from './styled';
+
+export interface CollapsibleCheckboxProps extends CheckboxProps {
```
golcinho, line 13: 🍹 `checked` is required but `onChange` is still optional (inherited as optional from the input attrs). a controlled checkbox with no `onChange` is read-only and warns in React. since you're committing to controlled here, consider requiring `onChange` too.

packages/big-design/src/components/Checkbox/styled.tsx
bc-nick: This duplication is low cost. `CheckboxImgContainer` should be isolate for  `Checkbox`. No need to make it as shared component yet.

golcinho (commented):
overall solid. the `Collapse` integration forwards through `collapseProps`/`triggerProps`/`panelProps` instead of exposing a parallel API that duplicates `CollapseProps` — good. `img` a11y is handled right too: decorative empty `alt` by default, `WCAG 1.1.1` cited. trigger is a sibling of the checkbox (not nested in `CheckboxContainer`), so no `onClick` propagation bug.

one thing worth fixing before merge:
- the `collapseProps.disabled overrides the default gating` test doesn't test the documented override. with `checked={true}` the default is already `disabled={!checked} === false`, so it only proves force-*disabling*; enabling the panel while *unchecked* (the case the spread order actually guards) is untested.

plus one 🍹 cleanup inline, take or leave.

packages/big-design-patterns/src/components/CollapsibleCheckbox/spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+      triggerTitle="Configure"
+    >
+      Panel content
+    </CollapsibleCheckbox>,
+  );
+
+  expect(screen.getByRole('button')).toBeDisabled();
+});
+
+test('collapseProps.disabled overrides the default gating', () => {
```
golcinho: test name says `overrides the default gating`, but with `checked={true}` the default is already `disabled={!checked} === false` — so this only proves `collapseProps={{ disabled: true }}` can force-*disable* an otherwise-enabled panel. the documented override is the other direction: enabling the panel *while unchecked* (`checked={false}` + `collapseProps={{ disabled: false }}`). that's the case the spread order `<Collapse disabled={!checked} {...collapseProps}>` actually has to guard, and it's untested.

packages/big-design/src/components/Checkbox/styled.tsx
```diff
@@ -19,8 +19,16 @@ export const CheckboxLabelContainer = styled.div<{ hasContent?: boolean }>`
...
+  flex-shrink: 0;
+  height: ${remCalc(40)};
+  object-fit: cover;
+  width: ${remCalc(40)};
+  margin: 0 ${({ theme }) => theme.spacing.xxSmall};
+`;
+
+export const CheckboxContainer = styled.div<{ isVerticalCenter?: boolean }>`
+  align-items: ${({ isVerticalCenter }) => (isVerticalCenter ? 'center' : 'flex-start')};
```
golcinho: 🍹 `isVerticalCenter` is only ever passed `Boolean(img)` (`Checkbox.tsx:118`), never anything else — the parameterized styled prop is unused flexibility. could just center when `img` is present and drop the prop.

chanceaclark:
I've chatted a bit with @andrewreifman about the design of this. He's fine with adding image support to `Checkbox` component but we probably want to avoid adding it as a pattern to `@bigcommerce/big-design-patterns` and instead just localize the component in the repo. I don't know if we'd use this pattern in the future.

From my side, I want to push back a little on this component. It feels weird from a user experience perspective. From what I see, the main goal of using a checkbox is to show that a row is enabled or not at a high level. Either way, when I user wants to enable a provider, they will need to dive into the configuration in order to set it up anyways, so why not let the user dive into the option if they are wanting to enable it first. We can show the status of the provider via a `Badge` instead to show the status of a provider so a user doesn't have to dive into the config. For example, I could see us using the `AccordionPanel` (_with BigDesign modifications to support this_) along with conditional children if the option is enabled (_not 100% sold on the conditional children_):

Screenshot 2026-06-17 at 16 28 55
Screenshot 2026-06-17 at 16 29 00

bc-nick:
> I've chatted a bit with @andrewreifman about the design of this. He's fine with adding image support to `Checkbox` component but we probably want to avoid adding it as a pattern to `@bigcommerce/big-design-patterns` and instead just localize the component in the repo. I don't know if we'd use this pattern in the future.
> 
> From my side, I want to push back a little on this component. It feels weird from a user experience perspective. From what I see, the main goal of using a checkbox is to show that a row is enabled or not at a high level. Either way, when I user wants to enable a provider, they will need to dive into the configuration in order to set it up anyways, so why not let the user dive into the option if they are wanting to enable it first. We can show the status of the provider via a `Badge` instead to show the status of a provider so a user doesn't have to dive into the config. For example, I could see us using the `AccordionPanel` (_with BigDesign modifications to support this_) along with conditional children if the option is enabled (_not 100% sold on the conditional children_):
> 
> Screenshot 2026-06-17 at 16 28 55 Screenshot 2026-06-17 at 16 29 00

@chanceaclark hi! Thanks for your attention on this matter.  I've been looking into `AccordionPanel` from the very beginning. `AccordionPanel` is not an option in our case.
Having only one item (checkbox which only enable/disable feature) in the configuration is not good for the user experience. But it is not only one reason. We make enabling features harder in this way. We need to make it simple and flexible. This component have to include a Checkbox to ensure consistency in the user interface, as at list. I tried to design the component with the possibility of future use in mind.

If there are any doubts about this, it's better to isolate it in payment-settings-ui. When you mentioned localizing a component in the repository, did you mean adding it to the place where it's used? For example, i needed to use this component in `payments-settings-ui`. If that's the case, I would go that way.

bc-nick:
This PR is closed since component implementation will be moved/isolated in the payment-settings-ui.

PR #1862, opened 2026-07-01
feat(component): Add optional virtualization to Tree component

Meztimuro:
## What?

Adds optional `virtualized `prop to `StatefullTree` component
Adds optional optional virtualization to `Tree` component

`StatefullTree` behaviour without `virtualized` prop is unchanged

## Why?
Component performance noticably drops on large trees (over 10k+ categories)

https://github.com/user-attachments/assets/23113476-a865-47d5-b90d-5852ba2510a7

## Testing/Proof

Updated tests


packages/big-design/src/components/StatefulTree/StatefulTree.tsx
```diff
@@ -15,17 +16,22 @@ export interface StatefulTreeProps<T>
...
-  defaultExpanded,
-  defaultSelected,
-  disabledNodes = [],
-  iconless,
-  onNodeClick,
-  onExpandedChange,
-  onSelectionChange,
-  selectable: type,
-}: StatefulTreeProps<T>): React.ReactElement<StatefulTreeProps<T>> => {
+export type StatefulTreeProps<T> = StatefulTreeBaseProps<T> & TreeVirtualizationProps;
```
chanceaclark: this actually breaks `StatefulTreeProps` for anyone extending it downstream, it went from an `interface` to a `type` intersected with a union and TS won't let you `extends` that. it's public too, not internal. changeset has it as `minor` but this needs to be flagged as breaking or fixed. `TableNext` already solved this exact problem with `pagination?: TablePaginationProps` nested under one optional prop instead of flattening it in, I'd do the same thing here instead of intersecting `TreeVirtualizationProps` directly

packages/big-design/src/components/Tree/Tree.tsx
```diff
@@ -23,39 +32,142 @@ export const Tree = <T,>({
...
+  focusedNodeRef.current = focusable.focusedNode;
+
+  const hasValidMaxHeight =
+    typeof maxHeight === 'number' && Number.isFinite(maxHeight) && maxHeight > 0;
+  const virtualized = virtualizedProp === true && hasValidMaxHeight;
+  const onNodeRefChange = useCallback(
+    (nodeId: TreeNodeId, node: HTMLLIElement | null, wasFocused = false) => {
+      if (!node) {
+        if (wasFocused) {
+          pendingFocusNodeRef.current = focusedNodeRef.current;
```
chanceaclark, line 58: found a weird one, scroll a focused row out of view with the mouse (not arrow keys) and this never gets cleared except by a matching remount. scroll it back into view later and it steals focus back even after you've clicked somewhere else. also means Tab from outside the tree just skips the tree entirely while that row's off-window since it's the only one with tabIndex=0. probably just needs to clear on any focus change, not only on remount

packages/big-design/src/components/Tree/Tree.tsx
```diff
@@ -23,39 +32,142 @@ export const Tree = <T,>({
...
+    [expandable.expandedNodes],
+  );
+  const selectedNodesSet = useMemo(
+    () => new Set(selectable?.selectedNodes ?? []),
+    [selectable?.selectedNodes],
+  );
+  const disabledNodesSet = useMemo(() => new Set(disabledNodes ?? []), [disabledNodes]);
+
+  const selectedChildrenCounts = useMemo(
+    () => (selectable?.type ? getSelectedChildrenCounts(nodes, selectedNodesSet) : EMPTY_COUNTS),
```
chanceaclark, line 95: random thought, this (and `useNodeMap`) still walks the whole tree every time regardless of `virtualized`, so for a huge tree you're saving the DOM cost but not the compute cost on every select/expand. might be worth gating these too, or at least calling it out as a follow up

packages/big-design/src/components/Tree/types.ts
```diff
@@ -55,6 +55,23 @@ export interface TreeProps<T> {
   onNodeClick?: TreeOnNodeClick;
 }
 
+export type TreeVirtualizationProps =
```
chanceaclark: not blocking, but curious if lazy loading came up at all here, this still needs the full `nodes` tree in memory upfront. `TableNext` already has `getChildren`/`getLoadMoreAction` for lazy children, might be worth a follow up issue pointing at that so we don't reinvent it differently later

packages/big-design/src/components/Tree/TreeNode/styled.tsx
```diff
@@ -4,8 +4,17 @@ import styled, { css } from 'styled-components';
 import { Flex, FlexItem } from '../../Flex';
 import { StyleableText } from '../../Typography/private';
 
-export const StyledLi = styled.li`
+export const StyledLi = styled.li<{ level?: number }>`
```
chanceaclark: should this be `$level`? right now it's not leaking to the DOM only because styled-components happens to filter unknown attrs, the other new props here ($maxHeight, $virtualized) are already transient

packages/big-design/src/components/Tree/types.ts
```diff
@@ -55,6 +55,23 @@ export interface TreeProps<T> {
...
+    }
+  | {
+      virtualized?: false;
+      maxHeight?: never;
+    };
+
+export type TreeProps<T> = TreeBaseProps<T> & TreeVirtualizationProps;
+
 export interface TreeContextState<T> {
   disabledNodes?: TreeNodeId[];
```
chanceaclark: looks like nothing reads this off context anymore now that `disabledNodesSet` exists, probably safe to drop

packages/big-design/src/components/Tree/hooks/index.ts
```diff
@@ -1,3 +1,5 @@
+export * from './useFlatVisibleNodes';
 export * from './useNodeMap';
 // Purposefully left out useSelectedNodeCount since it's an internal only hook.
```
chanceaclark: this comment mentions `useSelectedNodeCount` which doesn't exist anymore

chanceaclark (commented):
Before you push up any changes, if you utilize Claude, can run `/code-review`, `/simplify` and `/security-review` to prematurely review your own code?

packages/big-design/src/components/StatefulTree/StatefulTree.tsx
Meztimuro: Fixed
Added nested `virtualization?: TreeVirtualizationProps` prop

packages/big-design/src/components/Tree/Tree.tsx
Meztimuro: Fixed both issues
Thank you for finding it

packages/big-design/src/components/Tree/Tree.tsx
Meztimuro: I thought about how to improve it, but my current ideas seem like they will significantly grow this already large PR.

As of now, virtualization significantly improves performance for a very common use case (category selection tree for users with a large number of categories).

So I'd address this in a follow-up/later.

packages/big-design/src/components/Tree/types.ts
Meztimuro: Tbh, main idea was to fix a very specific issue. Issue being that **very** large category trees noticeably stuttered.

And since in our cases we always have full tree, adding optional virtualization seemed like a good solution for fixing issue while avoiding breaking/changing anything for existing usages of StatefulTree.

packages/big-design/src/components/Tree/TreeNode/styled.tsx
Meztimuro: Fixed, thank you

packages/big-design/src/components/Tree/types.ts
Meztimuro: Removed, thank you for noticing

packages/big-design/src/components/Tree/hooks/index.ts
Meztimuro: Removed, thank you for noticing

packages/big-design/src/components/Tree/hooks/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './useFlatVisibleNodes';
 export * from './useNodeMap';
-// Purposefully left out useSelectedNodeCount since it's an internal only hook.
 export * from './useTreeKeyEvents';
+export * from './useTreeVirtualizer';
```
chanceaclark: Do we actually want `useFlatVisibleNodes` and `useTreeVirtualizer` to be public API? This barrel gets re-exported from the package root, and the comment that used to live here purposefully kept internal-only hooks out. StatefulTree can import `useFlatVisibleNodes` by path, and nothing external needs `useTreeVirtualizer`. Once these ship we're stuck supporting their signatures. Am I missing a consumer that needs them?

packages/big-design/src/components/Tree/Tree.tsx
```diff
@@ -27,35 +36,205 @@ export const Tree = <T,>({
...
+  useEffect(() => {
+    if (hasVirtualization && !hasValidMaxHeight) {
+      warning(
+        'Tree: `virtualization.maxHeight` must be a positive, finite number to bound the scroll area.',
+      );
+    }
+  }, [hasVirtualization, hasValidMaxHeight]);
+
+  const flatNodes = useFlatVisibleNodes({
+    nodes: virtualized ? nodes : [],
```
chanceaclark: 🍹 nit: this passes a fresh `[]` on every render in non-virtualized mode, so the memo inside recomputes each time. It's flattening an empty array so it costs basically nothing, but hoisting an `EMPTY_NODES` constant next to `EMPTY_COUNTS` would let it actually memoize.

packages/big-design/src/components/Tree/Tree.tsx
```diff
@@ -27,35 +36,205 @@ export const Tree = <T,>({
...
+    [selectable?.selectedNodes],
+  );
+  const disabledNodesSet = useMemo(() => new Set(disabledNodes ?? []), [disabledNodes]);
+
+  const selectedChildrenCounts = useMemo(
+    () => (selectable?.type ? getSelectedChildrenCounts(nodes, selectedNodesSet) : EMPTY_COUNTS),
+    [nodes, selectable?.type, selectedNodesSet],
+  );
+
+  const initialTreeContext: TreeContextState<T> = useMemo(
```
chanceaclark, line 99: 🍹 For context, this memo doesn't buy much when rendered from StatefulTree: `expandable`/`focusable`/`selectable` are fresh object literals every render over there, so the context value changes anyway. The real per-node savings come from the memoized Sets/counts. Fine to keep as-is, but if the goal was a stable context those objects would need memoizing on the StatefulTree side too.

packages/docs/pages/statefulTree.tsx
```diff
@@ -158,6 +158,45 @@ const StatefulTreePage = () => {
...
+                  <Text>
+                    For large trees (thousands of nodes), pass <Code primary>virtualization</Code>{' '}
+                    with a <Code primary>maxHeight</Code> to render only the nodes within the
+                    viewport.
+                  </Text>
+                  <CodePreview>
+                    {/* jsx-to-string:start */}
+                    {function Example() {
+                      // 500 categories × 10 subcategories = 5,500 nodes.
+                      const nodes = Array.from({ length: 500 }, (_, index) => ({
```
chanceaclark: 🍹 The 5,500 node array gets rebuilt on every render, which defeats StatefulTree's memo. Low stakes in docs, but since this is the reference example for the feature, mind wrapping it in a `useMemo`?

packages/big-design/src/components/Tree/hooks/index.ts
Meztimuro: Thank you. Removed both hooks from the barrel and switched to direct imports.

packages/big-design/src/components/Tree/Tree.tsx
Meztimuro: Fixed. Thank you

packages/big-design/src/components/Tree/Tree.tsx
Meztimuro: Fixed. Thank you

packages/docs/pages/statefulTree.tsx
Meztimuro: Fixed. Thank you

chanceaclark (approved):

PR #1863, opened 2026-07-02
feat(component): img prop updates

bc-nick:
## What?

Allow the Checkbox `img` prop to accept native `img` attributes

## Why?

It is better approach. I did the similar thing for `InfoCard`.

This is a component library so we love visually looking at changes! If this applies to your pull request, show us your hard work in action.

## Testing/Proof

Screenshot 2026-07-02 at 13 18 39

### DIFF
--- a/packages/big-design/src/components/Checkbox/Checkbox.tsx
+++ b/packages/big-design/src/components/Checkbox/Checkbox.tsx
@@ -28,7 +28,7 @@ interface Props {
   label: React.ReactNode;
   description?: CheckboxDescription | string;
   badge?: BadgeProps;
-  img?: ComponentPropsWithoutRef<'img'>;
+  img?: Omit<ComponentPropsWithoutRef<'img'>, 'style' | 'className'>;
 }
 
 interface CheckboxDescription {
--- a/packages/big-design/src/components/Checkbox/Checkbox.tsx
+++ b/packages/big-design/src/components/Checkbox/Checkbox.tsx
@@ -28,21 +28,12 @@ interface Props {
   label: React.ReactNode;
   description?: CheckboxDescription | string;
   badge?: BadgeProps;
-  img?: CheckboxImg;
+  img?: ComponentPropsWithoutRef<'img'>;
 }
 
 interface CheckboxDescription {
   text: string;
   link?: FormControlDescriptionLinkProps;
-}
-
-export interface CheckboxImg {
-  src: string;
-  /**
-   * Accessible name for the thumbnail. Defaults to an empty string (decorative) when omitted —
-   * set it whenever the image carries meaning the label doesn't already convey.
-   */
-  alt?: string;
 }
 
 interface PrivateProps {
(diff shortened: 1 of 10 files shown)


golcinho (dismissed):

packages/big-design/src/components/Checkbox/styled.tsx
```diff
@@ -20,11 +20,7 @@ export const CheckboxLabelContainer = styled.div<{ hasContent?: boolean }>`
 `;
 
 export const CheckboxImgContainer = styled.img`
-  flex-shrink: 0;
-  height: ${remCalc(40)};
-  object-fit: cover;
```
chanceaclark, line 25: Are you sure you don't need `flex-shrink` and `object-fit` here?

packages/docs/PropTables/CheckboxPropTable.tsx
```diff
@@ -77,12 +77,14 @@ const checkboxProps: Prop[] = [
   },
   {
     name: 'img',
-    types: '{ src: string; alt?: string }',
+    types: "ComponentPropsWithoutRef<'img'>",
```
chanceaclark: Can you follow the pattern we have for documenting this. I believe a recent PR did this.

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -28,23 +28,14 @@ interface Props {
   label: React.ReactNode;
   description?: CheckboxDescription | string;
   badge?: BadgeProps;
-  img?: CheckboxImg;
+  img?: ComponentPropsWithoutRef<'img'>;
```
chanceaclark, line 31: Would you mind omitting `style` and `className` props from being passed into the img?

packages/big-design/src/components/Checkbox/styled.tsx
bc-nick: We do not need this here

packages/big-design/src/components/Checkbox/Checkbox.tsx
bc-nick: Lets leave it as it is

packages/docs/PropTables/CheckboxPropTable.tsx
bc-nick: Done

packages/big-design/src/components/Checkbox/styled.tsx
bc-nick: @chanceaclark your are right. better to leave it as it is but `contain` not `cover` value. updated where needed

packages/big-design/src/components/Checkbox/Checkbox.tsx
bc-nick: @chanceaclark omitted these props

packages/big-design/src/components/Checkbox/Checkbox.tsx
```diff
@@ -150,7 +141,7 @@ const RawCheckbox: React.FC<CheckboxProps & PrivateProps> = ({
       >
         {!checked && isIndeterminate ? <RemoveIcon /> : <CheckIcon />}
       </StyledCheckbox>
-      {img ? <CheckboxImgContainer alt={img.alt ?? ''} src={img.src} /> : null}
+      {img ? <CheckboxImgContainer height={40} width={40} {...img} alt={img.alt ?? ''} /> : null}
```
chanceaclark: You also need to omit the style and className props here too.

bc-nick: @chanceaclark omitted at the type level

chanceaclark: @bc-nick that doesn't actually omit the props from being passed. Typescript will yell at you but you can still `@ts-expect-error` on it and still pass in `style` or `className` and it'll work.

chanceaclark (approved):

PR #1870, opened 2026-07-27
ref: TRAC-1302 Replace v5-only styled-components type names in published types

chanceaclark:
Jira: [TRAC-1302](https://bigcommercecloud.atlassian.net/browse/TRAC-1302)

## What?

Removes `FlattenSimpleInterpolation` and `StyledComponent` from 4 call sites in our published `.d.ts` files. `FlattenSimpleInterpolation` (11 files) is replaced with a derived, version-agnostic alias `CSSRules = ReturnType<typeof css>`, defined in `big-design-theme`'s `helpers.ts` and consumed via `@bigcommerce/big-design-theme` everywhere else. The `styled<StyledComponent<...>>(...)` call sites in `Alert`, `Message`, `InlineMessage`, and `Fieldset/Legend` drop the explicit generic in favor of inferred typing, using the standard `.attrs()`/tagged-template generic merge instead.

`Form/Label`, `Radio/Label`, and `Checkbox/Label` still use the `StyledComponent`-based generic and are intentionally left out of this PR — see Why.

## Why?

Both type names only exist in styled-components v5's types (`@types/styled-components`) and vanish once we move to v6 (which ships its own types). Leaving them in the public type surface would break every v5 consumer's build the moment we cut over, so they need to come out now, ahead of the v6 upgrade, using aliases that resolve against either version's types.

Removing the explicit generic on the Radio/Checkbox/Form label components surfaces a latent, unsound type: their public props (`StyledLabelProps`/`FormControlLabelProps`) extend the native `label` HTML attributes, while the underlying styled component is built on `Text`/`H4` (i.e. a `p`/`h4` element), which only ever exposes `p`/`h4` attributes. The old generic cast silently hid the fact that these never actually matched once merged honestly (event handlers like `onCopy` are typed per-element and aren't interchangeable between `label` and `p`/`h4`). Fixing that properly (without narrowing the public prop shape or resorting to a cast) means rebuilding those 3 components on a native `styled.label` base rather than wrapping `Text`/`H4`, which is a real structural change beyond a type-name swap — it's being done as a separate PR (#1872) so it gets its own focused review and snapshot verification.

No visual changes — type-only refactor with no rendering differences.

## Testing/Proof

`pnpm run lint`, `pnpm run typecheck`, and `pnpm run test` all pass (70 suites / 1115 tests / 182 snapshots, no snapshot updates needed). Confirmed no new type assertions or eslint-disables introduced. Added a `patch` changeset for `@bigcommerce/big-design` and `@bigcommerce/big-design-theme`.

[TRAC-1302]: https://bigcommercecloud.atlassian.net/browse/TRAC-1302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/big-design-theme/src/helpers/helpers.ts
+++ b/packages/big-design-theme/src/helpers/helpers.ts
@@ -1,14 +1,19 @@
 import { em, hideVisually, math, rem, transparentize } from 'polished';
-import { css, FlattenSimpleInterpolation } from 'styled-components';
+import { css } from 'styled-components';
 
 import { themeOptions } from '../options';
+
+// Derived from `css` rather than referencing styled-components' own type name directly, so
+// it resolves against either v5 or v6 types and doesn't leak a version-specific name into
+// our published `.d.ts` files.
+export type CSSRules = ReturnType<typeof css>;
 
 export interface Helpers {
   addValues(first: string, second: string): string;
   createRGBA(color: string, alpha: number): string;
   remCalc(value: string | number): string;
   emCalc(value: string | number): string;
-  listReset: FlattenSimpleInterpolation;
+  listReset: CSSRules;
   hideVisually: typeof hideVisually;
 }
 
--- a/.changeset/replace-v5-only-styled-components-types.md
+++ b/.changeset/replace-v5-only-styled-components-types.md
@@ -1,0 +1,6 @@
+---
+"@bigcommerce/big-design": patch
+"@bigcommerce/big-design-theme": patch
+---
+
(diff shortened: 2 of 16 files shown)


packages/big-design-theme/src/helpers/helpers.ts
```diff
@@ -1,14 +1,19 @@
 import { em, hideVisually, math, rem, transparentize } from 'polished';
-import { css, FlattenSimpleInterpolation } from 'styled-components';
+import { css } from 'styled-components';
 
 import { themeOptions } from '../options';
 
+// Derived from `css` rather than referencing styled-components' own type name directly, so
+// it resolves against either v5 or v6 types and doesn't leak a version-specific name into
+// our published `.d.ts` files.
+export type CSSRules = ReturnType<typeof css>;
```
parthshahp, line 9: 🙏

parthshahp (approved):

PR #1872, opened 2026-07-28
refactor: TRAC-1302 rebuild label components on native styled.label

chanceaclark:
Jira: [TRAC-1302](https://bigcommercecloud.atlassian.net/browse/TRAC-1302)

## What?

Follow-up to #1870. Rebuilds `RadioLabel`, `CheckboxLabel`, and `FormControlLabel`'s underlying styled component on a native `styled.label` base, instead of wrapping the `Text`/`H4` typography primitives (`styled.p`/`styled.h4`) with an `as: 'label'` attrs override. This removes the last `styled<StyledComponent<...>>(...)` generic (a v5-only styled-components type name) from our published `.d.ts` files, without narrowing the public prop types — `StyledLabelProps`/`FormControlLabelProps` still extend the full native `label` element attributes, unchanged.

## Why?

#1870 covered the straightforward call sites, but Radio/Checkbox/Form Label needed more thought: their public props extend the native `label` HTML attributes, while the underlying styled component was built on `Text`/`H4` (i.e. a `p`/`h4` element). `.attrs({ as: 'label' })` only ever changes the *runtime* tag — styled-components never changes the *static* element type from it — so the old `styled<StyledComponent<...>>(...)` generic was the only thing making TypeScript believe otherwise (it faked the static element as `'label'`). Removing it surfaces a real type conflict: DOM event handlers generically parameterized by `HTMLLabelElement` aren't assignable where `HTMLParagraphElement`/`HTMLHeadingElement` is expected, since those have a deprecated `align` attribute `label` lacks.

Rather than narrow the public prop type to work around that (a real, if minor, API shape change), this rebuilds all three on a native `styled.label`, reproducing the exact CSS the wrapped `Text`/`H4` component provided (color/font-size/font-weight/line-height/margin) plus each component's own rules (cursor, hidden, margin overrides) directly. The static element is now genuinely `'label'`, so the full `ComponentPropsWithoutRef<'label'>` type-checks honestly with no narrowing, no assertions, and no `eslint-disable`s.

No visual changes. Verified in the docs site that Radio, Checkbox, and Form labels render identically (same computed styles, confirmed via snapshot diffs below).

Screenshot 2026-07-28 at 09 34 56

## Testing/Proof

`pnpm run lint` and `pnpm run typecheck` pass clean. `pnpm run test` passes (1115 tests); 23 snapshots needed updating (`Checkbox`, `Form`, `Input`, `Radio`, `Table`, `TableNext`, `Textarea`, `Worksheet`) — reviewed every diff and confirmed the only change is generated class-name numbering/merging (styled-components previously emitted a separate class for the extended definition; the native base now emits one), with every actual CSS declaration (color, font-size, font-weight, line-height, margin, cursor) byte-identical to before. Confirmed no remaining `StyledComponent`/`FlattenSimpleInterpolation` references in `Radio/Label`, `Checkbox/Label`, or `Form/Label`, and no new type assertions or eslint-disables introduced. Added a `patch` changeset for `@bigcommerce/big-design`.

[TRAC-1302]: https://bigcommercecloud.atlassian.net/browse/TRAC-1302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/big-design/src/components/Radio/Label/styled.tsx
+++ b/packages/big-design/src/components/Radio/Label/styled.tsx
@@ -1,18 +1,26 @@
+import { theme as defaultTheme } from '@bigcommerce/big-design-theme';
 import { ComponentPropsWithoutRef } from 'react';
-import styled, { css, DefaultTheme, StyledComponent } from 'styled-components';
+import styled, { css } from 'styled-components';
 
-import { TextProps } from '../../Typography';
-import { StyleableText } from '../../Typography/private';
+import { withMargins } from '../../../helpers';
 
 export interface StyledLabelProps extends ComponentPropsWithoutRef<'label'> {
   disabled?: boolean;
 }
 
-export const StyledLabel = styled<
-  StyledComponent<'label' | 'p', DefaultTheme, Partial<TextProps>> & StyledLabelProps
->(StyleableText).attrs({
-  as: 'label',
-})<StyledLabelProps>`
+export const StyledLabel = styled.label<StyledLabelProps>`
+  color: ${({ theme }) => theme.colors.secondary70};
+  font-size: ${({ theme }) => theme.typography.fontSize.medium};
+  font-weight: ${({ theme }) => theme.typography.fontWeight.regular};
+  line-height: ${({ theme }) => theme.lineHeight.medium};
+  margin: 0 0 ${({ theme }) => theme.spacing.medium};
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+
+  ${withMargins()};
+
   cursor: pointer;
 
   ${({ disabled }) =>
@@ -21,3 +29,5 @@ export const StyledLabel = styled<
       cursor: not-allowed;
     `}
 `;
+
+StyledLabel.defaultProps = { theme: defaultTheme };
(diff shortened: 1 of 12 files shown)


packages/big-design/src/components/Radio/Label/styled.tsx
```diff
@@ -1,18 +1,26 @@
...
 export interface StyledLabelProps extends ComponentPropsWithoutRef<'label'> {
   disabled?: boolean;
 }
 
-export const StyledLabel = styled<
-  StyledComponent<'label' | 'p', DefaultTheme, Partial<TextProps>> & StyledLabelProps
->(StyleableText).attrs({
-  as: 'label',
-})<StyledLabelProps>`
+export const StyledLabel = styled.label<StyledLabelProps>`
```
parthshahp, line 11: Will removing the 'p' option break things?

parthshahp (approved):
Looks good, just one small comment for understanding.

packages/big-design/src/components/Radio/Label/styled.tsx
chanceaclark: We were only using `styled(StylableText)` to get the text types, but we were "casting" the types and element to a `label` element. This shouldn't break things as we inlined the text styles and the types are left unchanged.

PR #1879, opened 2026-07-30
refactor: TRAC-1340 migrate Box and shared helpers to transient styled props

chanceaclark:
Jira: [TRAC-1340](<https://bigcommercecloud.atlassian.net/browse/TRAC-1340>)

Stage 0 (foundation + first consumer) of [LTRAC-1396](https://linear.app/commerce/issue/LTRAC-1396/migrate-bigdesign-styled-components-props-to-transient-dollar-prefixed). Later stages migrate the remaining \~70 files in small per-component PRs.

## What?

* Convert the shared `withMargins` / `withPaddings` / `withDisplay` helpers to read transient (`$`-prefixed) props, and add internal `toTransient{Margin,Padding,Display}Props()` mapping utilities (not re-exported from `helpers/index.ts` — imported directly from the helper subpaths).
* Migrate the first consumer, `Box`: a new internal-only `StyledBoxProps` with `$`-prefixed system props. Public `BoxProps` and CSS output are unchanged; `Box.tsx` now destructures every non-DOM prop out of the blind `{...props}` spread and routes them through the transient utilities + explicit `$name` JSX props. This is the template later blind-spread components copy.
* Add non-leakage assertions (`not.toHaveAttribute('display')`, etc.) to `Box/spec.tsx`.

## Why?

styled-components 6 drops v5's automatic filtering of unknown props on tag-target styled components, so system/style props (`backgroundColor`, `display`, `margin`, ...) would leak onto real DOM nodes and fire React's unknown-prop warning on every render once on v6. Transient props are the idiomatic fix — styled-components never forwards `$`-prefixed props to the DOM, at no per-render cost.

**Two deviations from the ticket, worth a careful look:**

1. **Temporary dual-read shim instead of "no shim".** The ticket argues no shim is needed because changing the helper generic "breaks tsc the instant the helper's generic shape changes," forcing an all-at-once conversion. That premise is false: the transient props are optional, so tsc stays green — but every unconverted consumer then silently drops its spacing/display CSS *at runtime* (54 snapshot failures when only Box was converted). The helpers therefore read `$prop ?? prop` (marked `MIGRATION SHIM (LTRAC-1396)`) so consumers migrate one PR at a time; the fallback is removed in a final cleanup PR once all are transient.
2. `display` **leaked on v5 too.** `display` is a valid SVG presentation attribute, so v5 forwarded it to the DOM. Fixing Box's blind spread removes `display="[object Object]"` / `display="none"` from components that route display through Box/Flex (Stepper, OffsetPagination, StatelessPagination, PillTabs, Table, TableNext). The updated snapshots are **removal-only** (all `-` lines, zero CSS or `+` changes) — verified.

No visual change — this is an internal prop-plumbing refactor with identical CSS output. The only DOM difference is the removal of the stray leaked `display` attribute described above.

## Testing/Proof

`pnpm --filter @bigcommerce/big-design run lint && run typecheck && run test` all green: lint clean, typecheck clean, **1146 tests pass / 182 snapshots**. `setupTests.ts`'s `failOnConsole()` plus the snapshot serializer's full-DOM output together act as a leak detector — any non-`$` prop reaching a DOM node fails the suite. `build:dt` succeeds and `dist/index.d.ts` contains no `Transient*` / `toTransient*` symbols, confirming the public API is unchanged.

Refs [LTRAC-1340](https://linear.app/commerce/issue/LTRAC-1340/support-image-optimization-for-makeswift-and-static-images)

[TRAC-1340]: https://bigcommercecloud.atlassian.net/browse/TRAC-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/big-design/src/components/OffsetPagination/__snapshots__/spec.tsx.snap
+++ b/packages/big-design/src/components/OffsetPagination/__snapshots__/spec.tsx.snap
@@ -347,7 +347,6 @@ exports[`render offset pagination component 1`] = `
         data-popper-escaped="true"
         data-popper-placement="bottom-start"
         data-popper-reference-hidden="true"
-        display="none"
         style="position: fixed; left: 0px; top: 0px; right: auto; bottom: auto; transform: translate(0px, 4px);"
       >
         <div
@@ -782,7 +781,6 @@ exports[`render offset pagination component with invalid page info 1`] = `
         data-popper-escaped="true"
         data-popper-placement="bottom-start"
         data-popper-reference-hidden="true"
-        display="none"
         style="position: fixed; left: 0px; top: 0px; right: auto; bottom: auto; transform: translate(0px, 4px);"
       >
         <div
@@ -1217,7 +1215,6 @@ exports[`render offset pagination component with invalid range info 1`] = `
         data-popper-escaped="true"
         data-popper-placement="bottom-start"
         data-popper-reference-hidden="true"
-        display="none"
         style="position: fixed; left: 0px; top: 0px; right: auto; bottom: auto; transform: translate(0px, 4px);"
       >
         <div
@@ -1653,7 +1650,6 @@ exports[`render offset pagination component with no items 1`] = `
         data-popper-escaped="true"
         data-popper-placement="bottom-start"
         data-popper-reference-hidden="true"
-        display="none"
         style="position: fixed; left: 0px; top: 0px; right: auto; bottom: auto; transform: translate(0px, 4px);"
       >
         <div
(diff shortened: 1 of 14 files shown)


packages/big-design/src/components/OffsetPagination/__snapshots__/spec.tsx.snap
```diff
@@ -347,7 +347,6 @@ exports[`render offset pagination component 1`] = `
         data-popper-escaped="true"
         data-popper-placement="bottom-start"
         data-popper-reference-hidden="true"
-        display="none"
```
chanceaclark, line 350: This is a valid SVG attribute, so our custom `display` prop was leaked into the DOM. This shouldn't affect or break any of these components but it will be removed moving forward.

rtalvarez (approved):

PR #1896, opened 2026-08-04
feat(component): migrate Dropdown, Popover, Select, MultiSelect to @floating-ui/react

chanceaclark:
## What?

Swaps `react-popper`'s `usePopper` for `@floating-ui/react`'s `useFloating` in the four internal-only consumers: Dropdown, Popover, Select, and MultiSelect. Popper modifiers map to floating-ui middleware for behavior parity: `offset [0, 4]` → `offset(4)`, `preventOverflow` → `shift()`, default flip → `flip()`, Popover's custom `sameWidth` modifier → `size()`, and popper's scroll/resize `eventListeners` → `autoUpdate` gated on `isOpen` so listeners still only attach while a menu is open.

Tooltip still uses `react-popper` (it exposes popper types publicly and is migrated separately as a breaking change), so the dependency stays installed for now.

## Why?

`react-popper` is unmaintained (no release since 2021) and its React peer caps at 18, blocking React 19 support. `@floating-ui/react` is the successor project from the same maintainer.

No public API change: the `placement` props stay typed with popper's `Placement`. That type includes `auto`/`auto-start`/`auto-end`, which floating-ui only supports via the `autoPlacement` middleware (and Popover *defaults* to `auto`), so a small internal `getFloatingPlacement` util maps those values.

Snapshot churn explained: floating-ui applies the offset transform synchronously in jsdom (popper never got that far in tests), and it consumes a `useId` internally which shifts subsequent generated ids. No positional values changed.

Positioning parity was measured on the docs site with headless Chromium via `getBoundingClientRect`:

| Scenario | Result |
| -- | -- |
| Dropdown open | 4px gap below trigger, `bottom-start` aligned (`leftDelta: 0`) |
| Dropdown, window scrolled 120px while open | menu tracks anchor, gap stays 4px |
| Dropdown near viewport bottom | flips above with 4px gap |
| Select open + viewport resize while open | position holds |
| MultiSelect with no room below | flips above input |
| Popover (`placement="auto"` default) | picks roomiest side, 4px distance |

## Testing/Proof

* `pnpm run lint`, `pnpm run typecheck`, `pnpm run test` all green (1152 tests; 17 snapshot updates reviewed, all id/transform churn as described above).
* Manual parity pass against `pnpm run start` docs in headless Chromium: opened each of the 4 components, measured gaps/alignment, scrolled and resized while open, and forced flip near viewport edges (numbers in the table above). Esc and outside-click close behavior on Popover re-verified. Zero browser console errors.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/.changeset/floating-ui-internal-positioning.md
+++ b/.changeset/floating-ui-internal-positioning.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/big-design": minor
+---
+
+Migrate Dropdown, Popover, Select, and MultiSelect positioning from `react-popper` to `@floating-ui/react`. Placement, offsets, flipping, and overflow behavior are unchanged, and the `placement` props keep accepting the same values (including `auto*` placements, now handled via floating-ui's `autoPlacement` middleware). Tooltip still uses `react-popper` and is migrated separately.
--- a/packages/big-design/package.json
+++ b/packages/big-design/package.json
@@ -39,6 +39,7 @@
     "@babel/runtime": "^7.26.10",
     "@bigcommerce/big-design-icons": "workspace:^",
     "@bigcommerce/big-design-theme": "workspace:^",
+    "@floating-ui/react": "^0.27.20",
     "@popperjs/core": "^2.11.6",
     "date-fns": "4.1.0",
     "downshift": "9.0.10",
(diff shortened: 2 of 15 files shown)


chanceaclark:
Screenshot 2026-08-04 at 13 03 23

rtalvarez (approved):

parthshahp (approved):

PR #1908, opened 2026-08-18
chore: TRAC-1511 remove commitlint from repo and circleci

chanceaclark:
## What?

Removes Commitlint from the repository and CircleCI: drops the `@commitlint/cli` and `@commitlint/config-conventional` devDependencies, deletes `commitlint.config.js`, and removes the `commitlint` CircleCI job and its `build_and_test` workflow entry.

## Why?

Commitlint is no longer useful and was the last CI gate depending on commit message shape (no local husky hook was ever installed, so it only ever ran in CircleCI). Removing it lets the pipeline run without that check, per LTRAC-1605.

N/A — CI/tooling change only, no UI impact.

## Testing/Proof

- Confirmed no remaining `commitlint` references anywhere in the repo (`grep -rIn "commitlint"` returns nothing outside this diff).
- `pnpm run typecheck` and `pnpm run lint` both pass across all packages.
- Regenerated `pnpm-lock.yaml` via `pnpm install --lockfile-only` to drop the now-unused transitive dependency tree.
- Added a changeset covering the affected published/private packages, consistent with how other root-level CI/tooling changes in this repo are versioned.


parthshahp (approved):
🙏

PR #1909, opened 2026-08-18
build(deps): TRAC-1508 Upgrade pragmatic-drag-and-drop to 3.0 (+ hitbox 2.1)

chanceaclark:
## What?

Bumps `@atlaskit/pragmatic-drag-and-drop` 2.0.1 → 3.0.0 and `@atlaskit/pragmatic-drag-and-drop-hitbox` 2.0.0 → 2.1.0 together, and migrates off deprecated import paths in `Table`/`TableNext` row DnD (4 files).

## Why?

Hitbox 2.1.0 requires core 3.0.0 (they're released as a coupled set), so they must bump together. While in there, the `element/adapter`, `combine`, `element/preserve-offset-on-source`, and `element/set-custom-native-drag-preview` import paths we use are marked `@deprecated` in 3.0.0 in favor of `adapter/element-adapter` and `utils/*`. Verified each replacement path against the published package's own `@deprecated` JSDoc redirects before swapping. No functional or API changes; `-live-region` is untouched since it isn't part of the coupled release.

Part of the pre-release dependency audit for the BigDesign React 19 release ([LTRAC-1600](https://linear.app/commerce/issue/LTRAC-1600)).

N/A — dependency/import-path only change, no visual or behavioral difference.

## Testing/Proof

- `pnpm run lint` and `pnpm run typecheck` pass with the new import paths.
- Full package test suite passes: 70 suites / 1155 tests, including all `Table`/`TableNext` DnD specs.
- Confirmed via npm/package inspection that `3.0.0`/`2.1.0` are the latest published versions and that the four deprecated paths still resolve with no functional break.

### DIFF
--- a/.changeset/upgrade-pragmatic-dnd-3.md
+++ b/.changeset/upgrade-pragmatic-dnd-3.md
@@ -1,0 +1,5 @@
+---
+'@bigcommerce/big-design': patch
+---
+
+chore(deps): upgrade `@atlaskit/pragmatic-drag-and-drop` 2.0.1 → 3.0.0 and `-hitbox` 2.0.0 → 2.1.0 (coupled release), used in `Table`/`TableNext` row DnD. Also migrates off the now-`@deprecated` `element/adapter`, `combine`, `element/preserve-offset-on-source`, and `element/set-custom-native-drag-preview` import paths in favor of `adapter/element-adapter` and `utils/*`. No functional or API changes.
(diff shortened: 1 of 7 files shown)


chanceaclark:
https://github.com/user-attachments/assets/9d0ceb41-ff14-472a-93ed-f67fe502cdf1

mfaris9 (approved):

PR #1910, opened 2026-08-18
build(deps): TRAC-1509 upgrade react-datepicker 7.6 to 9.1

chanceaclark:
## What?

Bumps `react-datepicker` 7.6.0 → 9.1.0, used in `Datepicker`. Also replaces the pre-existing popper.js-v2-shaped custom `popperModifiers` object with `@floating-ui/react`'s built-in `offset(-10)` middleware.

## Why?

Two majors on paper, but the type defs diffed clean between 7.6.0 and 9.1.0 — every prop this repo uses (`customInput`, `dateFormat`, `locale`, `min`/`maxDate`, `onBlur`/`onChange`/`onFocus`, `placeholderText`, `popperModifiers`, `popperPlacement`, `showPopperArrow`, `renderCustomHeader`, `required`, `selected`, `registerLocale`) is unchanged; additions are purely additive. The popper→`@floating-ui/react` migration already happened before 7.6.0. The new optional peer dep `date-fns-tz` (only needed for the new `timeZone` prop, which we don't use) installs cleanly with no warnings.

Regression testing surfaced one real behavior change not visible in the type diff: the library reworked the calendar's ARIA markup. Day cells changed `role` from `option` to `gridcell` (now wrapped in `table`/`rowgroup`/`row`, matching the WAI-ARIA date picker pattern) — the spec.tsx query for `role: 'option'` was updated to `gridcell` to match. The day-name header row (Sun/Mon/Tue/...) now also renders a visually-hidden full name alongside the visible abbreviation for screen readers; since we don't import `react-datepicker`'s base stylesheet, that visually-hidden span had no CSS to actually hide it, so both the full name and abbreviation rendered on top of each other. Fixed by adding the missing `.react-datepicker__sr-only` visually-hidden rule (copied from react-datepicker's own stylesheet) to `Datepicker`'s styles.

While in the file, also cleaned up the pre-existing `popperModifiers` usage flagged in the ticket as tech debt. It was a popper.js-v2-shaped `{ name, fn }` object that manually subtracted 10px from `y` — under `@floating-ui/react`'s middleware pipeline (which react-datepicker migrated to before 7.6.0) this isn't a no-op: it cancels out react-datepicker's internal `offset(10)` middleware, which is what keeps the calendar flush against the input today. Replaced it with the equivalent, more idiomatic `offset(-10)` from `@floating-ui/react` (already a direct dependency, used elsewhere in Dropdown/Popover/Select). Verified pixel-identical positioning before and after (~1px gap in both cases).

Not included inline; manually verified in Chrome — see Testing/Proof below.

## Testing/Proof

- `pnpm -F @bigcommerce/big-design run typecheck` — clean
- `pnpm -F @bigcommerce/big-design run test` — 70/70 suites, 1155/1155 tests pass (including the updated `Datepicker` spec)
- `pnpm -F @bigcommerce/big-design run lint` — clean
- Manually exercised the `Datepicker` docs page (basic + error-states examples) in Chrome: opened the calendar, confirmed the day-name header renders without overlapping text, selected a date, and measured the popup-to-input gap via JS (`popperRect.top - inputRect.bottom ≈ 1px`, matching pre-upgrade behavior).

Refs TRAC-1509

### DIFF
--- a/.changeset/upgrade-react-datepicker-9.md
+++ b/.changeset/upgrade-react-datepicker-9.md
@@ -1,0 +1,9 @@
+---
+'@bigcommerce/big-design': patch
+---
+
+chore(deps): upgrade `react-datepicker` 7.6.0 → 9.1.0, used in `Datepicker`. No functional or public API changes.
+
+The underlying library reworked the calendar's ARIA markup: day cells changed `role` from `option` to `gridcell` (now wrapped in `table`/`rowgroup`/`row`, matching the WAI-ARIA date picker pattern), and the day-name header row (Sun/Mon/Tue/...) now renders a visually-hidden full name alongside the visible abbreviation for screen readers. Since we don't import `react-datepicker`'s base stylesheet, that visually-hidden span had no CSS to hide it and rendered overlapping text in the header row; fixed by adding the missing `.react-datepicker__sr-only` visually-hidden rule to `Datepicker`'s styles. Test suite updated to match the new `gridcell` role.
+
+Also replaces the pre-existing popper.js-v2-shaped custom `popperModifiers` object (a `{ name, fn }` pair that manually subtracted 10px from `y`) with `@floating-ui/react`'s built-in `offset(-10)` middleware, canceling out `react-datepicker`'s internal `offset(10)` to keep the calendar flush against the input, same as before. No visual change.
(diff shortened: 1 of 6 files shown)


chanceaclark:
Screenshot 2026-08-18 at 11 51 07

rtalvarez (approved):

PR #1911, opened 2026-08-18
build(deps): TRAC-1512 upgrade Babel toolchain 7 to 8

chanceaclark:
Jira: [TRAC-1512](https://bigcommercecloud.atlassian.net/browse/TRAC-1512)

## What?

Bumps `@babel/core`, `-cli`, `-preset-env`, `-preset-react`, `-preset-typescript`, `-plugin-transform-runtime`, `-runtime` (big-design, -icons, -theme, -patterns) and `-standalone` (docs) from 7.x to 8.x, plus everything that upgrade actually required beyond the version bumps themselves.

## Why?

The version bump alone doesn't build. Three things needed real fixes:

- **`babel-plugin-styled-components` doesn't support Babel 8 yet**, even at its latest release — it hard-pins `@babel/core: ^7.0.0` as a peer and pulls `@babel/plugin-syntax-jsx`/`@babel/helper-annotate-as-pure`/`@babel/helper-module-imports` at `^7.x`, which fail Babel 8's `assertVersion` check at build time. Added a root `pnpm.overrides` forcing those three transitive deps to `^8.x` — the plugin's own code is fine once its deps are current.
- **Babel 8's parser now requires the `<T,>` trailing-comma disambiguator for single-type-param generic arrow functions in `.ts` files too**, not just `.tsx` as before (function declarations like `function foo<T>()` are unaffected). Fixed the 18 affected call sites in `big-design`. Each needed a `// prettier-ignore` too, since Prettier still treats the comma as redundant in `.ts` files and strips it back out otherwise (the common `<T extends unknown>` workaround is blocked by `@typescript-eslint/no-unnecessary-type-constraint`).
- **`CodePreview.tsx`'s two `@babel/standalone` calls** used the now-removed `allExtensions`/`isTSX` options. Replaced with `ignoreExtensions: true` (+ an explicit `syntax-jsx` plugin for the type-stripping-only pass), and pinned `runtime: 'classic'` on the live-eval pass — `preset-react`'s default JSX runtime flipped classic→automatic in v8, but `react-live` evaluates transpiled code directly with no module resolution, so it can't handle the automatic runtime's `import ... from "react/jsx-runtime"`.

Also dropped the now-removed `useESModules` option from `plugin-transform-runtime` (behavior is automatic via `package.json#exports` in v8), and kept `@types/babel__standalone` — `@babel/standalone@8.x` ships no bundled types, so it's still required for `docs` to typecheck.

N/A, no visual changes. The classic→automatic JSX runtime flip changes build output bytes (`react/jsx-runtime` imports instead of `React.createElement` calls) but renders identically.

## Testing/Proof

- `pnpm run build`, `typecheck`, `lint`, and `test` all pass clean across every package (1155 tests, 70 suites).
- Manually verified the live docs preview in a browser (`pnpm run start`): the Button page's `CodePreview` renders and live-edits correctly (edited the label text and saw the preview update with no console errors), and StatefulTable/StatefulTree — which exercise the reducer and hooks that needed the trailing-comma fix — render, sort, expand, and select correctly.
- Confirmed via a decompiled `@babel/standalone` build and isolated `@babel/core` repros that each option change (`ignoreExtensions`, `syntax-jsx`, `runtime: 'classic'`) behaves as described before applying it.

Changeset included: patch bumps for `@bigcommerce/big-design`, `-icons`, `-patterns`, `-theme`, and `docs`.

[TRAC-1512]: https://bigcommercecloud.atlassian.net/browse/TRAC-1512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/upgrade-babel-toolchain-8.md
parthshahp, line 1: Nitpick, but can we make the changeset a bit more comprehensible 😅 ? Things like: 
```
despite what the originating ticket assumed
```
probably don't really belong here

parthshahp (approved):

.changeset/upgrade-babel-toolchain-8.md
chanceaclark: For sure!

PR #1915, opened 2026-08-19
build(deps): TRAC-1516 upgrade Prettier 2 to 3 in docs CodePreview

chanceaclark:
## What?

Upgrades `prettier` from `2.8.8` to `3.9.6` in `packages/docs`, used by `CodePreview.tsx` for live code-sample formatting. Drops `@types/prettier`, which npm already flags as deprecated now that Prettier 3 ships its own types.

## Why?

Prettier 3 removes `prettier/parser-babel` in favor of separately loading `prettier/plugins/babel` (parser) and `prettier/plugins/estree` (printer), and `format()` is now async — Prettier 2's `format()` was synchronous.

`CodePreview.tsx` called `getInitialCode()` (which internally calls `format()`) directly in the render body to seed `useState`, and again in a `useEffect` on `children`/`language` change. Neither call site can `await` anymore, so restructured: `getInitialCode` is now `async`, and both `initialCode` (used by the reset button) and `code` are plain `useState` values populated by a single effect that awaits `getInitialCode` and sets both, guarded against setting state after the effect's inputs have changed/unmounted.

`react-live` (the live-preview engine) has no prettier dependency of its own — it uses `sucrase` for its own JSX transform — so the blast radius stays contained to this one file.

N/A — behavior-preserving migration, no visual changes.

## Testing/Proof

- `pnpm -F @bigcommerce/docs run build` — succeeds, all doc pages prerender.
- `pnpm run lint` and `pnpm run typecheck` — pass across the monorepo, no regressions.
- Ran the docs dev server and manually exercised `/button` in Chrome: confirmed the live editor renders correctly-formatted TSX on load, re-formats correctly when switching to the JS tab (babel transform + prettier format both re-run via the effect), live-edits the preview as expected, and the reset button restores the properly formatted initial code. No console errors related to the prettier change.


chanceaclark:
Screenshot 2026-08-19 at 14 37 57

parthshahp (approved):

PR #1927, opened 2026-08-20
fix: TRAC-1547 revert @babel/runtime to 7.x in shipped dependencies

chanceaclark:
Jira: [TRAC-1547](https://bigcommercecloud.atlassian.net/browse/TRAC-1547)

## What?

Reverts `@babel/runtime` (and the matching Babel 7.x devDependency toolchain used to compile these 4 packages) from `^8.0.0` back to `^7.26.10` in `big-design`, `big-design-icons`, `big-design-patterns`, and `big-design-theme`. Restores the `useESModules` option in the shared babel config that v8 made automatic. Leaves `packages/docs`' `@babel/standalone` 8.x bump and the `<T,>` trailing-comma syntax fixes from the original Babel 8 upgrade alone — neither is affected by this revert.

## Why?

LTRAC-1606 (TRAC-1512, already merged to main) bumped `@babel/runtime` to `^8.0.0` in all 4 packages' regular `dependencies` — the runtime helper package that every consumer's own dependency resolution has to satisfy, not just our own build tooling. Babel 8 removed the `./regenerator` export subpath. Any consumer that Babel-transpiles its own code for pre-native-generator targets (doesn't use SWC) breaks on 100% of routes once resolution hoists a shared `@babel/runtime` satisfying our `^8.0.0` range.

Found via the LTRAC-1370 release-gate packed-tarball smoke test: `sample-app-nodejs` (a real Next.js app with SWC disabled via `.babelrc`, exactly the profile this breaks) 500'd on every route with `Package subpath './regenerator' is not defined by "exports"`. LTRAC-1606's own scope (dev-toolchain compatibility) never tested against a real Babel-based consumer, which is exactly what this smoke test caught.

N/A — dependency version revert, no visual changes.

## Testing/Proof

- `pnpm run lint` / `pnpm run typecheck` / `pnpm run test` (1155 tests) / `pnpm run build` all pass clean on this branch.
- Rebuilt all 4 tarballs from this branch (combined with the LTRAC-1640 ESM-type-import fix, since `sample-app-nodejs` needs both), did a fresh `npm install` in the real `sample-app-nodejs` checkout, and started `next dev`: root `/` and `/products` (which renders `StatefulTable`/`Table`/`Dropdown`/`Select`) both now return `HTTP 200` with zero errors in the server log, where every route previously 500'd.

Must merge before the held Version Packages PR (#1905), alongside #1925 (LTRAC-1640), so both fixes fold into the pending `5.0.0`/`3.0.0`/`7.0.0`/`3.0.0` release.

[TRAC-1547]: https://bigcommercecloud.atlassian.net/browse/TRAC-1547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jorgemoya (approved):
Will we be able to adopt babel runtime 8 at a later point?

chanceaclark:
> Will we be able to adopt babel runtime 8 at a later point?

@jorgemoya tl'dr Claude opened this PR because `sample-app-nodejs` doesn't support the `@babel/runtime` version. It's more that the sample app hasn't been updated in a while and not really a huge issue. Claude thinks it is, but it really isn't. I'm tempted to close this PR out and just have it upgrade the sample-app repo to support 8.x. Thoughts?

jorgemoya:
@chanceaclark Probably a better idea

PR #1928, opened 2026-08-20
fix: TRAC-1548 fix circular require crash from inline type-only imports

chanceaclark:
Jira: [TRAC-1548](<https://bigcommercecloud.atlassian.net/browse/TRAC-1548>)

## What?

Follow-up to [LTRAC-1546](https://linear.app/commerce/issue/LTRAC-1546/unblock-makeswift-is-not-connected-cases-with-www-redirects) (bigcommerce/big-design#1925, already merged). Changes `@typescript-eslint/consistent-type-imports`'s `fixStyle` from `'inline-type-imports'` to `'separate-type-imports'` (the rule's default) and reapplies the resulting import corrections across `big-design`, `big-design-icons`, and `packages/docs` — the packages actually affected by the style difference.

## Why?

[LTRAC-1546](https://linear.app/commerce/issue/LTRAC-1546/unblock-makeswift-is-not-connected-cases-with-www-redirects) fixed the "type-only import compiles into a real runtime import" bug, but the `fixStyle` it used doesn't fully fix it. Verified with a minimal isolated repro through the real `@babel/core`: an inline `import { type X } from 'y'` still compiles to a bare `require('y')` for the module specifier — Babel can't prove the module has no side effects, so it keeps a side-effect-only import around. Only the whole-statement `import type { X } from 'y'` form is fully elided with zero runtime footprint.

That distinction is not cosmetic: `utils/messagingHelpers.ts` and `treeHelpers.ts` import prop types from `../components/Button`, `../components/Link`, and `../components/Tree`. Under the inline style, the compiled output still requires those modules, creating a real circular require chain back through `utils/index.ts` before its `typedMemo` export (declared near the end of the barrel) is assigned — crashing with `TypeError: (0, _utils.typedMemo) is not a function` the moment anything imports `@bigcommerce/big-design`.

This is reproducible with a plain `node -e "require('@bigcommerce/big-design')"` — no bundler, no Next.js, nothing involved beyond Node's own CommonJS loader. Confirmed via side-by-side tarball tests that it's a genuine regression from the Babel 8 toolchain upgrade ([LTRAC-1512](https://linear.app/commerce/issue/LTRAC-1512/goalie-rotation-22)): the currently-published `4.0.0` (built with Babel 7) doesn't crash; a fresh tarball built from `changeset-release/main` before this fix does. Also confirmed this fix is independent of and unblocks the separate `@babel/runtime` regenerator-subpath question ([LTRAC-1547](https://linear.app/commerce/issue/LTRAC-1547/cornerstone-versions-after-6190-dont-display-rule-based-out-of-stock)): a tarball with only this fix applied, `@babel/runtime` left at `^8.0.0`, loads cleanly via plain `require`, and a full Next.js app on SWC now serves every route at 200 where every route previously crashed.

N/A — import-statement syntax only, no visual changes.

## Testing/Proof

* `pnpm run lint` / `pnpm run typecheck` / `pnpm run test` (1155 tests) / `pnpm run build` all pass clean on this branch (based on current `main`, includes the [LTRAC-1545](https://linear.app/commerce/issue/LTRAC-1545/fix-makeswift-preview-loading-for-golfboxcomau) test fixes already merged there).
* Verified the compiled `dist/cjs/utils/messagingHelpers.js` and `treeHelpers.js` no longer contain the bare `require(...)` calls to `../components/Button`, `../components/Link`, `../components/Tree`.
* Repro/fix verified with a plain `node -e "require('@bigcommerce/big-design')"` against a freshly packed tarball, both before (crashes) and after (succeeds) this fix.

This is a correctness regression currently live on `main` (merged via bigcommerce/big-design#1925 before this was found) — should merge promptly, and definitely before [#1905](<https://github.com/bigcommerce/big-design/issues/1905>).

[TRAC-1548]: https://bigcommercecloud.atlassian.net/browse/TRAC-1548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/fix-circular-dependency-typedmemo.md
+++ b/.changeset/fix-circular-dependency-typedmemo.md
@@ -1,0 +1,7 @@
+---
+"@bigcommerce/big-design": patch
+"@bigcommerce/big-design-icons": patch
+"@bigcommerce/docs": patch
+---
+
+Fix a regression from the type-only-import fix (see the "ESM build shipping runtime imports for type-only names" changeset): using inline `type` modifiers (`import { type X }`) instead of the whole-statement form (`import type { X }`) still left a bare `require(...)` of the module behind, since Babel can't prove a plain import specifier has no side effects. In `utils/messagingHelpers.ts` and `utils/treeHelpers.ts`, that bare require created a real circular `require` chain back through `utils/index.ts` before its `typedMemo` export was assigned, crashing with `(0, _utils.typedMemo) is not a function` on `require('@bigcommerce/big-design')` — reproducible with no bundler involved, and present even under CommonJS/Node.js. `@typescript-eslint/consistent-type-imports`'s `fixStyle` is now `separate-type-imports`, which fully elides the runtime footprint whenever every binding from a specifier is type-only.
--- a/packages/big-design-icons/src/flags/base/index.tsx
+++ b/packages/big-design-icons/src/flags/base/index.tsx
@@ -3,11 +3,11 @@ import {
   type Spacing,
   type ThemeInterface,
 } from '@bigcommerce/big-design-theme';
-import {
-  type ComponentPropsWithoutRef,
-  type ForwardRefExoticComponent,
-  type PropsWithoutRef,
-  type RefAttributes,
+import type {
+  ComponentPropsWithoutRef,
+  ForwardRefExoticComponent,
+  PropsWithoutRef,
+  RefAttributes,
 } from 'react';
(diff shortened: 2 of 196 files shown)


jorgemoya (approved):

chanceaclark:
Closes #1927

